PR #137, opened 2022-04-27
XXX test theory around #88

rzezeski:
So I only have a moment to write this up before heading off to court but I have a theory: perhaps #88 has to do with the amount of codegen units, which explodes the `ld` argument list. IIRC rust/cargo recently reintroduced incremental compilation, which itself had an implicit side effect of bumping `codegen-units` from `16` to `256`. This makes them explicit, taking the units back down to `16`, except in the case of `release` profile where I took it down to `1` (b/c supposedly it _could_ produce more efficient code).

### DIFF
--- a/opte/Cargo.toml
+++ b/opte/Cargo.toml
@@ -68,10 +68,10 @@ pcap-parser = { version = "0.11.1", features = ["serialize"] }
 codegen-units = 1
 
 [profile.dev]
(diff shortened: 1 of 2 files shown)


rzezeski:
This is dead for now as I don't think my theory was correct and this issue has not been a problem for us since then.

PR #141, opened 2022-05-05
Add commands to add/remove firewall rules to OPTE ioctl crate

bnaecker:


rzezeski:
IIRC there was a large discussion around firewall rules a while back and the conclusion was that, at least for the time being, we want sled agent to be able to always set firewall rules wholesale, in an atomic manner, for a given OPTE port. In that case we should just bite the bullet and implement that API now instead of making you do extra work in sled agent. If that was indeed the plan I have a wad drafted up:

https://github.com/oxidecomputer/opte/commit/9b82df5521e0be884eb51d0af7af5e7b77c1a729

bnaecker:
Actually, one option might be to add that commit to this PR. I agree the sled-agent will definitely be atomically replacing a rule set wholesale, but given that the API exists, it might make sense to keep it around. I don't feel strongly though, whatever works.

bnaecker:
@rzezeski This seems superseded by https://github.com/oxidecomputer/opte/pull/145, right?

PR #146, opened 2022-05-09
packaging: add xde.conf to p5p package

rcgoodfellow:
Without `/kernel/drv/xde.conf` the `xde` driver will fail to attach. This commit adds `xde.conf` to the p5p package for `xde`.


rzezeski:
Seems good. FWIW I don't see other manifests setting the `mode` except for executables. Not sure it matters at all but thought I'd mention it.

rcgoodfellow:
I think the doc I was following for the original package where I set the file perms was something like this one

- https://docs.oracle.com/cd/E26502_01/html/E21383/pkgcreate.html

Happy to remove the `mode` option if this is a better way to do things. For this PR, was just trying to be consistent with what's already there.

rzezeski:
Totally, I should have been more clear: this is good to go by me. I just thought I'd mention it because these things always seem so finicky and upon checking several manifests I noticed they didn't bother to set mode.

rzezeski (approved):

PR #147, opened 2022-05-10
Move `ListPortsReq` into the ioctl API

bnaecker:


rzezeski (approved):
Thanks Ben. Push at your discretion.

Issue #204, opened 2022-07-28
Implement paused state


rzezeski:
Implement `PortState::Paused`. This is the first step towards support for live-migration. This issue is specifically limited to only the `Port`'s notion of the paused state; the greater question of quiescence and interaction with mac is a driver (xde) concern will be a separate issue. This deals only with making sure that no traffic can flow across a `Port` when in the paused state. This is needed as a basis for implementing a save mechanism, which will also be a separate issue.

In general, when in the paused state, only introspective actions should be allowed on the port. No traffic should flow, rules become immutable, all dynamic state around flows becomes immutable, etc. Any action that is not allowed should return an error to the caller indicating that the port is in a paused state.

rzezeski:
Addressed in 6620202b4576be0c5957ce9e7c3c355c351ed5e5.

PR #206, opened 2022-07-29
PR: Implement paused state #204

rzezeski:
Coming soon to a theater near you.


opte/src/api/mac.rs
```diff
@@ -94,4 +94,8 @@ impl MacAddr {
     pub fn bytes(&self) -> [u8; 6] {
         self.inner
     }
+
+    pub const fn from_const(bytes: [u8; 6]) -> Self {
```
bnaecker, line 98: Is this to have a `const` version of `impl From<[u8; 6]>`?

opte/src/engine/layer.rs
```diff
@@ -1228,6 +1239,16 @@ impl<'a> RuleTable {
         Err(Error::RuleNotFound { id })
     }
 
+    pub fn rule_id(&self, query_rule: &Rule<Finalized>) -> Option<RuleId> {
+        for (rule_id, rule) in self.rules.iter() {
```
bnaecker, line 1243: I usually reach for [find](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find) for something like this, but it's not much clearer given how simple the predicate is.

opte/src/engine/port.rs
```diff
@@ -258,15 +258,16 @@ impl PortBuilder {
 /// PortBuilder::create() -> Ready
 /// PortBuilder::restore() -> Restored
 ///
-/// Ready -- mc_start() --> Running
+/// Ready -- Port::start() --> Running
```
bnaecker, line 261: I don't quite understand what these functions refer to. Are this a mapping from method name to the state that it puts the port in? I'm guessing it's really:

```
current_state -- transition_func --> new_state
```

But a comment explaining would be really helpful!

opte/src/engine/port.rs
```diff
@@ -389,7 +386,7 @@ pub struct Port {
 macro_rules! check_state {
     ( $sg:expr, [ $( $state:expr ),* ] ) => {
         if $( *$sg != $state )&&* {
-            Err(OpteError::BadState)
+            Err(OpteError::BadState(format!("{}", *$sg)))
```
bnaecker, line 389: Might not be a big deal, but I think we could leave the `OpteError::BadState` variant as carrying the actual state, rather than the stringification. That would make it easier to operate on the failure programmatically, though I don't know that's important yet.

opte/src/engine/port.rs
```diff
@@ -561,12 +576,17 @@ impl Port {
...
+
         for l in &*self.layers {
             if l.name() == name {
                 return Ok(l.dump());
             }
         }
 
+        // Perhaps this is silly, but I want to make sure the state
+        // lock is held for the execution of this method.
+        drop(state);
```
bnaecker, line 589: I think this is a solid way to do that. The other way I've seen is to introduce an inner scope:

```rust
    {
        let state = self.state.lock();
        // Do stuff with state
    }
    // Lock is dropped here.
    Err(OpteError::LayerNotFound(name.to_string()))
```

opte/src/engine/port.rs
```diff
@@ -693,6 +719,9 @@ impl Port {
             });
         }
 
+        // Perhaps this is silly, but I want to make sure the state
+        // lock is held for the execution of this method.
+        drop(state);
```
bnaecker, line 724: Actually, I'm looking at the lock in more detail now. It seems like it protects just the `PortState`, but this clearly implies it protects more than that, is that right? If so, we may want to put all the data that needs synchronization behind the lock. If that's a lot of fields, on common pattern is to put everything that needs to be locked into its own "inner" struct, and then storing that in the mutex.

opte/src/engine/rule.rs
```diff
@@ -1216,6 +1244,36 @@ pub struct RulePredicates {
...
+        if self.hdr_preds.len() != other.hdr_preds.len() {
+            return false;
+        }
+
+        if self.data_preds.len() != other.data_preds.len() {
+            return false;
+        }
+
+        for hp in &self.hdr_preds {
+            if !other.hdr_preds.contains(hp) {
```
bnaecker, line 1260: This is probably a _tiny_ issue now, but we may want to store the predicates in a way that avoid this O(N**2) comparison here. If we can sort the predicates, you could use [`binary_search`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search) to get O(N log N) or just compare hashes to get O(2 * N) == O(N).

opte/src/oxide_vpc/api.rs
```diff
@@ -217,6 +217,19 @@ pub struct AddRouterEntryIpv4Req {
     pub target: RouterTarget,
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct DelRouterEntryIpv4Req {
```
bnaecker, line 221: Nice definitely needed!

xde/src/xde.rs
```diff
@@ -413,7 +413,11 @@ const ONE_SECOND: Interval = Interval::from_duration(Duration::new(1, 0));
 
 #[no_mangle]
 fn expire_periodic(port: &mut Arc<Port>) {
-    port.expire_flows(Moment::now());
+    // XXX The call fails if the port is paused; in which case we
```
bnaecker, line 416: Is it an issue that this would also ignore the error when the port is _not_ paused? I'm not sure what we can do about it, other than log a message to the syslog or something, but it might be good to emit it somewhere at this point.

opte/src/engine/int_test.rs
```diff
@@ -99,6 +102,234 @@ fn next_block(offset: &[u8]) -> (&[u8], LegacyPcapBlock) {
...
+            }
+        }
+
+        assert_port!($port_and_vps);
+    };
+}
+
+// Like incr!, but do not assert_port! at the end. This is useful when
+// other state needs to be updated before asserting.
+macro_rules! incr_na {
```
bnaecker, line 229: We could write `incr!` in terms of this macro, I think, if you wanted to trim a few lines.

opte/src/engine/int_test.rs
```diff
@@ -163,18 +395,41 @@ fn oxide_net_builder(name: &str, cfg: &VpcCfg) -> PortBuilder {
...
+        .create(UFT_LIMIT.unwrap(), TCP_LIMIT.unwrap());
+    assert_eq!(port.state(), PortState::Ready);
+    assert_eq!(port.epoch(), 1);
+    check_no_flows(&port);
+    let vps = VpcPortState::new();
+    let mut pav = PortAndVps { port, vps };
+    update!(
+        pav,
+        [
+            "set:arp.rules_in=1,arp.rules_out=2",
```
bnaecker, line 414: So I admit I'm a bit confused about these fields. It seems like we're counting the rules or flows for the port, is that right? Making this separate seems like we have two sources of truth, though. I'd have thought the counts could be either updated automatically as rules or flows are added or removed; or they'd be computed on the fly whenever we need them.

I guess I mean, why do we want the counts in the `VpcPortState` field, distinct from whatever's added to the `Port` itself?

bnaecker: Ok, I _completely_ missed the fact that this is a test-only file. I'm betting that answers my question, and sorry about that!

bnaecker (commented):
This is a lot of work, thanks! I don't have any substantive comments, since I'm still just learning how everything fits together.

One general question I have is, what happens to inbound traffic, from the peer's perspective, when a port is paused? That is, if I have an open TCP connection to the guest when it's paused, does my connection appear to hang? Do the segments sit in the host OS buffers until they're either dropped, the guest reads them, or my own TCP stack times out the connection?

opte/src/api/mac.rs
rzezeski: Yea, it allowed me to simplify test code (see the `int_test::GW_MAC_ADDR`).

opte/src/engine/layer.rs
rzezeski: Yea, in this case it also needs to be followed with an `Option::map()`. In my Erlang days I was all about HOFs, but I think the last 6 years of C has made me imperative again. In this case it's small and reads fine, so I'm going to change it.

Also, the first time I read this too quickly and thought you were referring to the function name. So now that function name is also `find_rule`...which honestly is probably more clear.

opte/src/engine/port.rs
rzezeski: Yea, you got it right. I'm adding text to help clarify.

opte/src/engine/port.rs
rzezeski: Yea, I thought about that. The problem is it then requires `PortState` to be part of the API. Which, maybe it should be, but I didn't want to go down that refactoring path with this already large PR.

opte/src/engine/port.rs
rzezeski: Yea, so this kind of happened for a bunch of different reasons. Part of it has to do with my influence from illumos/mac/C and how we typically structure protection of resources. We usually have multiple locks to allow a semi-fine-grained approach in order to minimize contention. Another typical pattern is to prevent the admin path from contending with the data path. In the case of mac this is done via the "mac perimeter" which makes sure all admin commands execute sequentially but does not block any data path traffic. However, occasionally an admin command needs to coordinate-with or freeze the datapath, and mac does that with a system of per-CPU locks and reference counts. I'm not quite sure how you do something like that in Rust, as you aren't really partitioning the ownership but rather the locks that say "yes, you can access this". Though, you could use the flow id/5-tuple to truly partition ownership, and then I think this maps better with Rust.

All that said, you are totally right to call this out, as it's basically me trying to map my C/illumos like idioms into Rust. I agree that in Rust the idea is for the mutex to represent ownership of a thing at runtime, allowing one to publish something to multiple consumers without fear of corruption or races. To me it was important that I kept some form of "performance" by using the design I currently have, but now I'm realizing 1) it assumes the consumer of the generic OPTE is mac (as in it assumes that mc_start/mc_stop is how a port is started/stopped) and 2) you could theoretically have a race where the port is paused and then one more packet is processed. That could be detrimental to the save routine (which isn't written yet but needs to be) as it iterates the layer/UFT state, as it could interleave its reading of state with the packet's mutation of state.

This is making me think that the first thing to do is make things obviously correct but using a BFM (big fucking mutex), and thus forcing admin and data to contend with each other. However, when this becomes a hinderance on perf we could transition to something more like what mac does and partition ownership on some mechanism. For example, I already want to split out the UFT into TCP, UDP, and Other protocols. So that could be a natural place to split ownership of state. Then, when an admin command needs to coordinate with the datapath, it could use a similar mechanism to mac to quiesce the data.

Although, if we make it a requirement that opte is always executed in the context of a mac provider (like xde), then perhaps we take advantage of that fact? We would make sure the mac perim is used for all admin commands (by making sure the DLD entrypoint grabs the perim...assuming that's possible), and then on the datapath just assume the port's state cannot change while a packet is being processed (because this is exactly what mac promises to a provider). You still need the various fine-grained locks currently in opte as you can have concurrent data threads, but you could get rid of the port state lock as it can only change when data is quiesced and there is no traffic flowing through the port.

I need to think through that more and maybe talk with Robert, but certainly this has raised the need to write up an issue. For the purposes of this PR I want to leave it as-is, as it works fine for our purposes at the moment.

opte/src/engine/port.rs
rzezeski: I don't think that does what I'm after. The point is I'm not doing anything with `state`, but what the lock to be help. I wasn't sure if the compile could detect this and just optimize it away or something (also I want to silence the warning). I'll just leave it as-is for now.

opte/src/engine/rule.rs
rzezeski: Yea, honestly I'm not too worried about this atm with everything else. It also makes me thing the first thing we might want is to have an upper bound on number of predicates as that should probably be a thing no matter what algo we use.

xde/src/xde.rs
rzezeski: This call can only fail when the port is paused, that's what my comment is trying to say. I think the better solution is to just remove the periodic when the port is in this state, but that's work that needs to happen in xde and isn't part of this PR.

opte/src/engine/int_test.rs
rzezeski: Good call.

opte/src/engine/int_test.rs
rzezeski: Yea, sorry, I didn't really write up any comments in the file about what this is call for. The point of VpcPortState and its macros is to drastically cut down the redundancy in the tests around the checking of various port state like number of rules, flows, and the current epoch. I'll add a comment.

rzezeski:
> One general question I have is, what happens to inbound traffic, from the peer's perspective, when a port is paused? That is, if I have an open TCP connection to the guest when it's paused, does my connection appear to hang? Do the segments sit in the host OS buffers until they're either dropped, the guest reads them, or my own TCP stack times out the connection?

This PR is only concerned with pause from the perspective of the Port. From it's perspective it just returns an error and it's up to the caller on what to do.

In the case of xde we could just choose to drop the packet on the floor. However, I think instead we'll probably make use of mac's quiescing mechanism in order to pause traffic. In that case I believe it should stop the client from sending Tx frames (though in the case of a VPC guest and RFD 71 perhaps the guest is already stopped and we have no Tx to worry about), and on the Rx side I believe mac will buffer N bytes of data in the SRS (Soft Ring Set). That part of pause is kind of the next step of things to look at: how do we pause from the perspective of xde? How is the pause command sent down to xde? etc.

opte/src/engine/port.rs
bnaecker: Thanks for the details, that's very helpful context.

It's true that Rust's `Mutex` owns the data it locks, which is usually presented as a good thing since one can't inadvertently access a field that is protected by the lock. You just can't get that field without acquiring the lock. I don't think that precludes fine-grained locking, since we can group fields that all require the same synchronization (e.g., admin vs data-path) into separate `Mutex`es. I don't know enough about MAC to understand how that interacts.

opte/src/engine/port.rs
bnaecker: Yeah, that's fair, I'm not sure what happens to a lock guard that you never access. Maybe an argument for locking the data itself, to avoid worrying about that. FWIW, you can silence the "unused variable" warnings by prefixing the name with `_`, e.g., `let _state = self.state.lock();`.

Leaving it is fine, I agree it works.

opte/src/engine/rule.rs
bnaecker: Yeah, I figured it'd be lost in the noise of other stuff. Thanks!

bnaecker (approved):

Issue #169, opened 2022-06-30
package should allow xde.conf to be edited


jclulow:
The `opte` package delivers the `xde.conf` driver configuration file.  When you upgrade that file, it gets replaced with the new contents.  This is somewhat surprising to folks, I think, who are setting tuneables in there.

It should probably be marked with a `preserve` attribute with value `renamenew`.  As per **pkg(7)**:

```
                   o   If the file delivered by the upgraded version of the
                       package exists in the image and has different content
                       from the file delivered by the currently installed
                       version of the package, the upgrade is done according
                       to the value of the preserve attribute:
...
                             o   If the file delivered by the upgraded version
                                 of the package has a preserve value of
                                 renamenew, the new file is installed with the
                                 extension .new and the existing file is not
                                 modified.
...
                   o   If the file delivered by the upgraded version of the
                       package exists in the image, has the same content as
                       the file delivered by the currently installed version
                       of the package, and has a preserve value of either
                       renameold or renamenew, the existing file is replaced
                       by the file delivered by the upgraded version of the
                       package, including replacing permissions and timestamp
                       (if present).
```

i.e., if you haven't edited it, we'll deliver the new one; otherwise we'll deliver the new one alongside and mention it to you for review.

The attribute `preserve=renamenew` would go here in the template:

https://github.com/oxidecomputer/opte/blob/14410f9420184702fa4b0ee134a0aa37216398eb/pkg/opte.template.p5m#L16

bnaecker:
Closed by #213

PR #213, opened 2022-08-09
Add pkg metadata to preserve an existing xde.conf file

bnaecker:


bnaecker:
Should resolve #169

rzezeski (commented):
LGTM, thanks for doing this.

Issue #212, opened 2022-08-08
Move integration tests to oxide_vpc


rzezeski:
Move the integration tests to `oxide_vpc` where they belong. Maybe rename to `integration_tests` or `vpc_tests` as well. Need #206 to land first.

rzezeski:
Addressed by 33df8a4180053a4cd6981da876cb87ea11a18276.

PR #217, opened 2022-08-10
Move Oxide VPC impl to its own crate

rzezeski:
~~Still in progress.~~ Started working on #212 and realized I might as well move VPC into its own crate (which has been the plan for a while). Will write up some other issues around this soon.


rzezeski:
This looks bigger than it is. The main thrust of this was to move the integration test next to the VPC code (#212), because it tests the VPC code. While there I figured I should finally just split the VPC implementation into its own crate (#218), as I've wanted to do for a while now. Those two things were pretty mechanical. While working on #218 I had a compiler error around implementing the `Resource` trait, and I realized that the implementation for resources could be simpler (#219). That's the novel work, but it was a fairly small change.

rzezeski:
Oh, before I forget: I need to make an update to omicron to use the oxide-vpc crate and test this locally before merging.

oxide-vpc/src/engine/overlay.rs
```diff
@@ -420,7 +414,7 @@ impl Virt2Phys {
 }
 
 impl Resource for Virt2Phys {}
-impl Resource for Arc<Virt2Phys> {}
+// impl Resource for std::sync::Arc<Virt2Phys> {}
```
bnaecker, line 417: Dead code?

xde/src/xde.rs
```diff
@@ -1860,6 +1859,7 @@ unsafe extern "C" fn xde_mc_propinfo(
 fn new_port(
     name: String,
     cfg: &VpcCfg,
+    v2p: Arc<overlay::Virt2Phys>,
```
bnaecker, line 1862: This seems like the heart of the change around the `Resource` trait, IIUC. Previously, this data was passed to the `Port::process` method as the last argument. Now, the V2P mapping is still maintained by the driver's soft-state, and a pointer to that is passed to each port at the time it's created. Is that a fair, if rough, summary?

Does the `overlay::Virt2Phys` object do its own internal synchronization? It seems like it has to, for it to be shared in an `Arc<_>` instead of an `Arc<Mutex<_>>`.

bnaecker (approved):
LGTM, lots of cleanup / separation, and the `Resource` simplification is a nice win too. Thanks.

oxide-vpc/src/engine/overlay.rs
rzezeski: Yep, good catch.

xde/src/xde.rs
rzezeski: Yes. Previously we placed this in the `Resources` map which was just anymap underneath; and we threaded that thru the entire callstack down to the action. The V2P mapping is still owned by xde, and each port's encap action gets a pointer/Arc to this mapping.

And yes, `Virt2Phys` performs its own synchronization since it's in an Arc.

rzezeski:
For posterity: the only change I made in the most recent force-push was to remove the line of dead code (I so look forward to using Gerrit as CR).

Issue #220, opened 2022-08-11
Delete comment in v2p about mutex


rzezeski:
There is an old XXX comment in the `Vrit2Phys` implementation that laments about the use of mutex. I want to remove this comment but I thought it would be good to writeup an issue so others can understand my decision making.

```
    // XXX Wrapping these in a mutex is definitely a terrible idea (in
    // terms of lock contention as this resource is actually shared by
    // all ports); but for purposes of dev this is fine for now.
    // However, before v1 we'll want something like a shadow copy and
    // atomic pointer swap or perhaps some CoW/persistent data
    // strcuture that can be efficiently and safely updated
```

While past Ryan's shadow-copy idea is certainly interesting, I think he was a little cracked out on the coffee that day. Yes, this table is shared by all ports, and thus any guest sending traffic will at some point have to access (read) it, but the access to this table is limited.

* The V2P is read by the action only when a new flow is starting up. After the flow is established the mapping is cached in the LFT/UFT.
* Only ports in the same VPC access the same V2P table.
* The only time the V2P table is modified is via a `SetVirt2Phys` command. And that only happens when a guest is added, removed, or migrated.

Given this access pattern we are probably okay with a mutex. Next you might wonder why we don't just make it a rwlock, and that's a perfectly fine idea. But when I have that thought I always come back to the [rwlock(9f)](https://illumos.org/man/9F/rwlock) man page, which heeds this warning:

```
Readers/writer locks are slightly more expensive than mutex locks, and
the advantage of multiple read access may not occur if the lock will only
be held for a short time.
```

These reads are held for a **very** short time, just enough to create a struct value. This is one of those cases where I would personally wait to prove lock contention is an issue and then play with rwlock.

However, even if this did turn into a hot lock, there's another escape hatch. Right now, anytime there is a flow between GuestA and GuestB, the encap action has to lookup the mapping for each new flow as it is created. But you could also create another lookup cache, that is owned purely by the `EncapAction`, which stores results from v2p lookups, and requires no locking. That would also require a little more machinery to track an epoch, to know when the global v2p has marched forward and to invalidate the local cache. Of course, then in the worst case you have to do 2 map lookups instead of one, but I digress. The point I'm trying to make is that there's little sense getting wrapped around the axle about this until it's proven to be a problem. And I can say that in the normal case the path around the UFT is going to be the hot path (as that's the entire idea behind the VFP architecture).

rzezeski:
[...]

PR #222, opened 2022-08-11
PR: Delete comment in v2p about mutex (#220)

rzezeski:


bnaecker (approved):
Thanks for the detailed writeup, both here and in the related issue!

Issue #223, opened 2022-08-12
Produce a better error message on stray writes


rzezeski:
Implement the improvements I discussed in [this comment](https://github.com/oxidecomputer/opte/issues/63#issuecomment-1067262966). Comment reproduced below.

Based on the line numbers in the stack you should be running `d861d26636e1b38db40c2b74bf6a61bc01e468d1` (not the commit you linked). That said, still looks like a legit bug. Also, the method could be more helpful by reporting the issue differently:

* For `start > wptr`, report it as `OutOfRange`.
  * Modify `OutOfRange` to include the `start` value, the `limit` value (where `limit is `wptr`), and a `diff` value with how many bytes past.
  *  Remove the `ModAppend` and `Prepend` types as they are not used and I have plans to replace all this logic anyways.
* For `end > wptr`, continue reporting it as `NotEnoughBytes`, but correctly calculate `available` as `wptr.offset_from(start)`.

rzezeski:
Addressed in 9f235d91e3d2c611596dece3033778016a5ca1e4.

PR #224, opened 2022-08-12
PR: Produce a better error message on stray writes (#223)

rzezeski:


opte/src/engine/packet.rs
```diff
@@ -2550,7 +2519,40 @@ mod test {
...
+    #[test]
+    fn out_of_range() {
+        let mp1 = allocb(54);
+        let pkt = unsafe { Packet::<Uninitialized>::wrap(mp1) };
+        assert_eq!(pkt.num_segs(), 1);
+        assert_eq!(pkt.avail(), 54);
+
+        let body = vec![];
+        let mut tcp = TcpHdr::new(3839, 80);
+        tcp.set_seq(4224936861);
```
bnaecker, line 2537: Out of curiosity, how do you pick these sequence numbers? Why not start at 0 or 1?

opte/src/engine/packet.rs
```diff
@@ -1919,8 +1891,17 @@ pub enum ReadErr {
...
-    OutOfRange,
+    NotEnoughBytes {
+        start: *const u8,
+        end: *const u8,
+        available: usize,
+        needed: usize,
+    },
+    OutOfRange {
+        start: *const u8,
+        limit: *const u8,
```
bnaecker, line 1902: IIUC, the field name `limit` comes from the `db_lim` field of the `struct datab`, which is described as the "last byte (+1) of buffer". My reading is that this is the first byte _past the end_ of the buffer, so that the size of the buffer is `db_lim - db_base`. If that's all correct, I think this name could use a quick comment, which could be just a reference to the `db_lim` field and/or `man datab.9s`.

bnaecker (commented):
All looks good, but I think we could use a quick comment on the `limit` field of the error type. Other than that, LGTM!

opte/src/engine/packet.rs
rzezeski: Just pulled it from random tcp dump laying around. The ISN (initial sequence number) is not meant to start at 0/1, or at least I believe it is highly unlikely.

bnaecker: Interesting, thank you. Some sources seem to say the ISN _cannot_ be zero, but [the spec](https://www.ietf.org/archive/id/draft-ietf-tcpm-rfc793bis-25.html#section-3.4-21) just says that it must not be predictable, and should be a pseudorandom value based on the timestamp and connection 4-tuple (plus nonce). TIL!

opte/src/engine/packet.rs
rzezeski: This field isn't referring to `db_lim`, it's the end of the buffer that the write has available to it. In the case of a `Modify` that is `wptr`, as it can only write over existing bytes. I pushed a change that hopefully makes these field names more clear.

bnaecker (approved):
Much clearer, thanks.

Issue #161, opened 2022-06-23
Want way to check for API_VERSION update


rzezeski:
```
/// NOTE: Unfortunately this doesn't automatically catch changes to
/// the API and upate itself. We must be vigilant to increment this
/// number when modifying the API.
```

As my comment above `API_VERSION` warns: it is easy to forget to increment the number when changing the API. I recently ran into an instance create issue in Omicron when upgrading `xde` from the current included commit of `b9980158540d15d44cfc5d17fc0a5d1848c5e1ae` to the latest commit of `12450552bdea4b3f97d309bfed5ad65dc2e2a775`. The failure is the following, found in the sled-agent log:

```
{"msg":"request completed","v":0,"name":"SledAgent","level":30,"time":"2022-06-23T19:32:58.149139933Z","hostname":"kalm","pid":2419,"uri":"/instances/d9356749-e1af-4856-94e4-fd1b827fe5dc","method":"PUT","req_id":"f56b57b4-db98-4975-8585-fac2dc0374c5","remote_addr":"[fd00:1122:3344:101::3]:56544","local_addr":"[fd00:1122:3344:101::1]:12345","component":"dropshot (SledAgent)","error_message_external":"Internal Server Error","error_message_internal":"Error managing instances: Instance error: Failure interacting with the OPTE ioctl(2) interface: command CreateXde failed: DeserCmdReq(\"Hit the end of buffer, expected more data\")","response_code":"500"} 
```

This points to the `CreateXde` ioctl. Sure enough, if we look at the 7 commits that have happened in the meantime, we find our culprit:

```
Author: Ryan Zezeski <ryan@oxide.computer>
Date:   Fri May 20 12:55:33 2022 -0600

    SNAT should predicate on router target (#153)
```

This change specifically, which updates the `CreateXdeReq` structure.

https://github.com/oxidecomputer/opte/commit/204784212bd21d88c29e64683bb53b17237c7604#diff-69375b575b303e0752185db891c91e1476979415c39d98a9db7f438e7dfbd676

As I failed to update the `API_VERSION` value alongside this change it means that the older client will not catch the change upfront and thus will send a serialized buffer that is too short, causing the kernel's handler to fail with an `OpteError::DeserCmdReq` during the `copy_in_req()` call.

Three things should happen:

1. Update the `API_VERSION` immediately in its own commit to hopefully spare others this pain.
2. Update sled-agent to work with the new API.
[...]

PR #230, opened 2022-08-16
PR: Want way to check for API_VERSION update (#161)

rzezeski:
Let's see if this works.


rzezeski:
This looks bigger than it is. Mainly just moved the `opte::api` module to the `opte-api` crate. Along with that I had to move some of the methods that are implemented for the API types from `opte::engine` to the `opte-api` crate. There's also some import churn because I've been removing the blank lines between import statements just to clean things up. Finally, there is the check-api-version script, which is the real point of this PR. I'm leaving it in this failed state so you can verify the CI failure for the opte-api job.

opte-api/src/ip.rs
```diff
@@ -157,6 +198,7 @@ pub enum Protocol {
     ICMP = 0x1,
     IGMP = 0x2,
     TCP = 0x6,
+    Chaos = 0x10,
```
bnaecker, line 201: This sounds fun! What's this part about, actually?

opte/Cargo.toml
```diff
@@ -11,7 +11,7 @@ engine = ["api"]
 # This feature indicates that OPTE is being built with std. This is
 # mostly useful to consumers of the API, providing convenient methods
 # for working with the API types in a std context.
-std = []
+std = ["opte-api/std"]
```
bnaecker, line 14: Is there any consumer of the `opte-api` crate that does _not_ use the `std` feature? There are some types that are duplicated with those in `std`, like the IP addresses. I think that came initially from defining those types directly in the `opte` core crate or some other no-std environment, but we might reduce maintenance by using those `std` types directly if we can get away with it.

opte-api/src/ip.rs
rzezeski: This was just to make a change to opte-api to test that the check script catches it in CI. It will be removed before putting back this change.

opte-api/src/ip.rs
```diff
@@ -213,13 +271,69 @@ pub struct Ipv4Addr {
...
+    pub fn safe_mask(self, prefix_len: Ipv4PrefixLen) -> Self {
+        self.mask(prefix_len.0).unwrap()
+    }
+
+    /// Produce a `u32` which itself is stored in memory in network
+    /// order. This is needed for passing this type up to DTrace so
+    /// its inet_ntoa() subroutine works.
+    pub fn to_be(self) -> u32 {
+        // First we create a native-endian u32 from the network-order
+        // bytes, then we convert that to an in-memroy network-order
```
bnaecker, line 311: Nit: spelling `in-memroy` -> `in-memory`

opte/Cargo.toml
rzezeski: Yes, the API must be shared with clients as well as the engine, as these types are how clients send requests to the engine.

.github/buildomat/jobs/opte-api.sh
```diff
@@ -0,0 +1,48 @@
...
+cd opte-api
+
+header "check API_VERSION"
+./check-api-version.sh
+
+header "check style"
+ptime -m cargo fmt -- --check
+
+header "analyze std"
+ptime -m cargo check
```
bnaecker, line 30: Any particular reason we're checking here, if we're going to build in the next step? I usually think of `cargo check` as a quicker approximation of `cargo build`.

bnaecker (commented):
The main course, the check on the API version, looks great. I've a few questions and nits otherwise. Thanks!

opte/Cargo.toml
bnaecker: Oh, I just misread this as a default feature. I understand the API types are shared between consumers and the engine, I think I just convinced myself that the engine is built _with_ this feature, which it may not be.

opte-api/src/ip.rs
bnaecker: Sweet, thanks!

.github/buildomat/jobs/opte-api.sh
rzezeski: You know that's a good question. I must of somehow got it in my head that `check` was doing additional stuff. But reading the cargo docs makes it clear it's the same as `build` without the final codegen. So we should remove these.

opte/Cargo.toml
rzezeski: You are actually onto something here. Now that the API types are their own crate the only real reason to build opte with `std` is for testing. However, there was also the historical idea that perhaps opte would also run in contexts other than as part of a mac provider like `xde`. Going back to the conversations we had around concurrency control, we might want to put a stake in the ground and say that opte is only ever meant to run in that type of context. Or perhaps we want to keep it flexible and just state that whatever is driving the opte engine needs to enforce concurrency control. But this is all opening a can of worms I'm trying to keep closed for the moment, or at least well outside the scope of this change. That said, now that the API types are in their own crate it might make sense to at least not have `std + api` as defaults, but I'll leave that to its own issue/PR if we decide that feels better.

opte-api/src/ip.rs
rzezeski: Done.

.github/buildomat/jobs/opte-api.sh
rzezeski: I'm going to modify the build scripts in a separate PR just to keep from adding to this one.

rzezeski:
Thanks for the review. I fixed the spelling error and removed the `Chaos` protocol change so now I expect everything should pass.

opte/Cargo.toml
bnaecker: That all makes sense, and putting this off to another PR after more discussion sounds solid. It also answers the original question, with "Yes, we still need types that work in both std and no-std environments".

rzezeski:
Looks like some intermittent GH failures:

```
69 | 2022-08-17T17:14:17.896Z | fatal: unable to access 'https://github.com/oxidecomputer/opte/': Empty reply from server
-- | -- | --
```

[69](https://buildomat.eng.oxide.computer/wg/0/details/01GAPC7Z10WYG53247EDQNP24V/cfq8T5ViWr39oVnMPgRApXcUq1U0QdHIBBdrLrt5vBstaBWm/01GAPC9M0SRBFCHZ2KB8Y08PCZ#S69)	2022-08-17T17:14:17.896Z	fatal: unable to access 'https://github.com/oxidecomputer/opte/': Empty reply from server

Issue #232, opened 2022-08-17
Remove superfluous work in buildomat jobs


rzezeski:
Currently the various buildomat jobs both run `cargo check` and `cargo build`, but the former is the same as the later just without the additional step of codegen -- we are duplicating work.

Furthermore, we are running both `--debug` and `--release` builds for each job, and I'm not sure that makes any sense. I think that only makes sense for jobs that are delivering an artifact, like `opteadm`, `xde`, etc. In fact, for jobs that deliver no artifacts there is no point in running `build` as there is no reason for codegen, and we can just use `check`.

rzezeski:
Done in d315a4a3582642d81a8da42c04b15c00a8439ef5.

PR #233, opened 2022-08-17
PR: Remove superfluous work in buildomat jobs (#232)

rzezeski:


rzezeski:
LOL, the `check-api-version.sh` script is a little too strong for its own good. Need to fix that so it only fails if there is a change to a source file without a version bump.

bnaecker (approved):

PR #238, opened 2022-08-23
Add minimal support for IPv6 extension headers

bnaecker:
- Use smoltcp to parse IPv6 extension headers. This removes the
  `Ipv6HdrRaw` and uses `smoltcp`'s "packet" types in its place.
- Store raw byte array (uninterpreted) corresponding to a packet's
  extension headers in the `Ipv6Hdr` type. We only need the length, so
  we know where the ULP starts, and the raw bytes, so we can re-emit
  them in the processed packet.
- Adds tests verifying the handling of the four currently supported
  extension headers

### DIFF
--- a/opte-api/src/ip.rs
+++ b/opte-api/src/ip.rs
@@ -237,6 +237,39 @@ impl TryFrom<u8> for Protocol {
     }
 }
 
+impl TryFrom<smoltcp::wire::IpProtocol> for Protocol {
+    type Error = String;
+
+    fn try_from(
+        proto: smoltcp::wire::IpProtocol,
+    ) -> core::result::Result<Self, Self::Error> {
+        use smoltcp::wire::IpProtocol::*;
+        match proto {
+            Icmp => Ok(Protocol::ICMP),
+            Igmp => Ok(Protocol::IGMP),
+            Tcp => Ok(Protocol::TCP),
+            Udp => Ok(Protocol::UDP),
+            Icmpv6 => Ok(Protocol::ICMPv6),
+            Unknown(x) if x == 0xFF => Ok(Protocol::Reserved),
+            _ => Err(format!("unhandled IP protocol: 0x{:X}", u8::from(proto))),
+        }
+    }
+}
+
+impl From<Protocol> for smoltcp::wire::IpProtocol {
+    fn from(proto: Protocol) -> smoltcp::wire::IpProtocol {
+        use smoltcp::wire::IpProtocol::*;
+        match proto {
+            Protocol::ICMP => Icmp,
+            Protocol::IGMP => Igmp,
+            Protocol::TCP => Tcp,
+            Protocol::UDP => Udp,
+            Protocol::ICMPv6 => Icmpv6,
+            Protocol::Reserved => Unknown(0xFF),
+        }
+    }
+}
+
 /// An IPv4 or IPv6 address.
 #[derive(
     Clone, Copy, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize,
--- a/opte/src/engine/ip6.rs
+++ b/opte/src/engine/ip6.rs
@@ -7,15 +7,17 @@
 use super::checksum::Checksum;
 use super::headers::{
     Header, HeaderAction, IpMeta, IpMetaOpt, ModActionArg, PushActionArg,
-    RawHeader,
 };
 use super::ip4::Protocol;
-use super::packet::{PacketRead, ReadErr, WriteError};
+use super::packet::{PacketRead, ReadErr};
 use core::convert::TryFrom;
-use core::mem::size_of;
 pub use opte_api::{Ipv6Addr, Ipv6Cidr};
 use serde::{Deserialize, Serialize};
-use zerocopy::{AsBytes, FromBytes, LayoutVerified, Unaligned};
+use smoltcp::wire::IpProtocol;
+use smoltcp::wire::Ipv6FragmentHeader;
+use smoltcp::wire::Ipv6HopByHopHeader;
+use smoltcp::wire::Ipv6Packet;
+use smoltcp::wire::Ipv6RoutingHeader;
 
 cfg_if! {
     if #[cfg(all(not(feature = "std"), not(test)))] {
@@ -27,8 +29,9 @@ cfg_if! {
 
 pub const IPV6_HDR_VSN_MASK: u8 = 0xF0;
 pub const IPV6_HDR_VSN_SHIFT: u8 = 4;
-pub const IPV6_HDR_SZ: usize = size_of::<Ipv6HdrRaw>();
+pub const IPV6_HDR_SZ: usize = smoltcp::wire::IPV6_HEADER_LEN;
 pub const IPV6_VERSION: u8 = 6;
+pub const DDM_HEADER_ID: u8 = 0xFE;
 
 #[derive(
     Clone, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize,
@@ -43,27 +46,7 @@ impl PushActionArg for Ipv6Meta {}
 
 impl From<&Ipv6Hdr> for Ipv6Meta {
     fn from(ip6: &Ipv6Hdr) -> Self {
-        Ipv6Meta {
-            src: ip6.src,
-            dst: ip6.dst,
-            // XXX Parse extension headers
-            //
-            // * Create ExtensionHdrs enum and various Extension
-            // Header types
-            //
-            // * Create a NextHdr enum which can be either Protocol or ExtHdr
-            //
-            // * Write parse_ipv6_ext() function to parse various ext
-            // headers. Call that in parse_ipv6().
-            //
-            // * Change how this meta structure is built. That is,
-            // replace this function with something that builds the
-            // IPv6 Metadata from base header + extensions.
-            //
-            // * Remember that you don't need to do all this right
-            // now, you have a demo to make.
-            proto: ip6.next_hdr,
-        }
+        Ipv6Meta { src: ip6.src, dst: ip6.dst, proto: ip6.proto }
     }
 }
 
@@ -88,15 +71,17 @@ impl Ipv6Meta {
 #[derive(Clone, Debug)]
 pub struct Ipv6Hdr {
     vsn_class_flow: [u8; 4],
+    // Length of payload, including extension headers
     payload_len: u16,
-    // The next_hdr is the first Next Header value. The proto is the
-    // actual upper layer protocol.
-    next_hdr: Protocol,
+    // Protocol of the next header, which may be an extension or the upper-layer
+    // protocol.
+    next_hdr: IpProtocol,
+    // The upper-layer protocol
     proto: Protocol,
     hop_limit: u8,
     src: Ipv6Addr,
     dst: Ipv6Addr,
-    // XXX Add vec, array, anymap of extension headers
+    extension_headers: Vec<u8>,
 }
 
 #[macro_export]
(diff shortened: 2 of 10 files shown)


opte/src/engine/ip6.rs
```diff
@@ -209,49 +210,116 @@ impl Header for Ipv6Hdr {
...
+                    extension_headers.extend_from_slice(remainder);
+                }
+                x => {
+                    return Err(Ipv6HdrError::UnexpectedNextHeader {
+                        next_header: x.into(),
+                    });
+                }
+            }
+        }
+        // Safety: The protocol is the last value of next header, and since
```
rzezeski, line 277: For the comments around `unwrap()` I've been using an `Unwrap:` tag (that's not totally true, there's tons of unwrap that has no annotation). Now I'm wondering if instead it should be `Panic:`, with the idea being that we document why we do not expect a panic -- which typically would involve the fact that we have information that we can't give to the compiler. Thoughts?

opte/src/engine/ip6.rs
```diff
@@ -262,113 +330,242 @@ impl From<&Ipv6Meta> for Ipv6Hdr {
...
+    // - Extension header length
+    // - Full header length
+    #[test]
+    fn test_extension_header_lengths_ok() {
+        for n_extensions in 0..SUPPORTED_EXTENSIONS.len() {
+            for extensions in
+                SUPPORTED_EXTENSIONS.into_iter().permutations(n_extensions)
+            {
+                let (buf, pos) = generate_test_packet(extensions.as_slice());
+                let mut simple_packet = SimplePacket::from(buf);
```
rzezeski, line 542: Why do we need `SimplePacket`? Can't we just use something like `PacketReader::new(Packet<Initialized>::copy(buf), ())`?

opte-api/src/ip.rs
```diff
@@ -237,6 +237,39 @@ impl TryFrom<u8> for Protocol {
     }
 }
 
+impl TryFrom<smoltcp::wire::IpProtocol> for Protocol {
```
rzezeski, line 240: So unfortunately this would require that clients (like sled-agent) pull in smoltcp, which isn't great. We really want this to remain an implementation detail of opte itself.

opte/src/engine/ip6.rs
```diff
@@ -46,23 +49,8 @@ impl From<&Ipv6Hdr> for Ipv6Meta {
...
-            //
-            // * Change how this meta structure is built. That is,
-            // replace this function with something that builds the
-            // IPv6 Metadata from base header + extensions.
-            //
-            // * Remember that you don't need to do all this right
-            // now, you have a demo to make.
-            proto: ip6.next_hdr,
+            proto: ip6.proto,
+            // TODO-completeness: Handle IPv6 extension headers in metadata
```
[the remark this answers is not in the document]
bnaecker: Sure, that's fine. It was an attempt to keep the relevant portion of the existing TODO, but seems OK to remove it regardless.

opte/src/engine/ip6.rs
```diff
@@ -209,49 +210,116 @@ impl Header for Ipv6Hdr {
...
+                    let buf = rdr.slice(rdr.seg_left())?;
+                    let header = Ipv6RoutingHeader::new_checked(buf)?;
+                    let n_bytes = 8 * (usize::from(header.header_len()) + 1);
+                    extension_headers.extend_from_slice(&buf[..n_bytes]);
+                    next_header = header.next_header();
+                    rdr.seek_back(buf.len() - n_bytes)?;
+                }
+                IpProtocol::Ipv6Frag => {
+                    // This header's length is fixed
+                    let buf = rdr.slice(8)?;
```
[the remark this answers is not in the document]
bnaecker: Sure can!

opte/src/engine/ip6.rs
bnaecker: It sure likes like it. See, the thing is, you have to _know that's there_. 😬 Thanks for pointing it out, I'll switch everything over.

opte/src/engine/ip6.rs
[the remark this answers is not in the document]
bnaecker: Actually, no:

This:

```
dbg!(std::mem::size_of::<smoltcp::wire::Ipv6FragmentRepr>());
```

yields:

```
[src/engine/ip6.rs:253] std::mem::size_of::<smoltcp::wire::Ipv6FragmentRepr>() = 12
```

Someone isn't using `repr(packed)`...

opte/src/engine/ip6.rs
bnaecker: I think `Panic:` is a good choice. I've seen `Safety` in a lot of places, but it's usually explain the explicit use of an `unsafe` block. I'll switch it.

opte-api/src/ip.rs
bnaecker: Makes sense. I'll remove it, and put some conversions into the `ip6` module directly. We can figure out how to make them more accessible if we need to later.

bnaecker: Actually, I just realized that I didn't have to update `Cargo.toml`, because it's already in there. Is that intentional? I'll just move these conversions now, we can strip that out later if a follow-up PR.

opte/src/engine/ip6.rs
```diff
@@ -126,47 +116,56 @@ macro_rules! assert_ip6 {
...
-    /// Return the length of the payload portion of the packet.
+    /// Return the length of the payload portion of the packet, including any
+    /// extension headers.
     ///
     /// NOTE: This currently does not entertain Jumbograms.
     ///
     /// XXX We should probably check for the Jumbogram extension
     /// header and drop any packets with it.
     pub fn pay_len(&self) -> usize {
-        self.payload_len as usize - self.ext_len()
```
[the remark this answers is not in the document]
bnaecker: Thanks for pointing that distinction. I agree we should probably stick to having `pay_len` return the payload length as defined by the spec. That's just the least likely to cause confusion.

I'll change calls from `pay_len()` to `ulp_len()`, and look at the `HeaderGroup` method as well. Thanks.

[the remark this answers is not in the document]
bnaecker: One question, how would one write a test to catch this? I'm assuming it requires that the `pay_len` and `ulp_len` methods return different values, i.e., that there _are_ extension headers? I wasn't sure how to put that in place, but it seems like a good idea given this.

opte-api/src/ip.rs
rzezeski: Sigh, yea because I already did this myself. And I think I had to do that to satisfy the orphan rule, e.g. I couldn't implement `From<smoltcp::wire::Ipv4Address> for Ipv4Addr` in opte because the types are defined in opte-api now.

opte/src/engine/ip6.rs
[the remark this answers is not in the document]
rzezeski: Integration tests around IPv6 going through full port process would have caught this. E.g. see the `integration_test.rs` in oxide-vpc.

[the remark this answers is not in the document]
rzezeski: Well, actually, don't need all that. It should be enough to use `Packet::parse()` on an IPv6 packet with extension headers.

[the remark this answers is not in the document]
bnaecker: Ok that's good to know. I'm going to add a test into `packet.rs` that parses an IPv6 packet with some extension headers.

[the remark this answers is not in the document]
bnaecker: Done in https://github.com/oxidecomputer/opte/pull/238/commits/50c482edefc0540b256a5d63ee34cc4cde229986

opte/src/engine/packet.rs
```diff
@@ -2622,4 +2620,48 @@ mod test {
...
+                );
+                let eth = EtherHdr::new(EtherType::Ipv6, SRC_MAC, DST_MAC);
+
+                let mut packet_buf = Vec::with_capacity(1024);
+                packet_buf.extend_from_slice(&eth.as_bytes());
+                packet_buf.extend_from_slice(&ip6.as_bytes());
+                packet_buf.extend_from_slice(&tcp.as_bytes());
+
+                let generated_packet = Packet::copy(&packet_buf);
+                generated_packet.parse().unwrap();
```
rzezeski, line 2663: You'll want to assert header offsets here. That's why you didn't catch the bug in `parse_hg_ipv6()` when it was still using fixed size. For example, I was able to induce errors by changing that method back to the the previous version and adding the following assertions:

```Rust
                let parsed = generated_packet.parse().unwrap();

                let offsets = &parsed.state.hdr_offsets;
                assert_eq!(offsets.inner.ip.as_ref().unwrap().seg_idx, 0);
                assert_eq!(
                    offsets.inner.ip.as_ref().unwrap().seg_pos,
                    ETHER_HDR_SZ
                );
                assert_eq!(
                    offsets.inner.ip.as_ref().unwrap().pkt_pos,
                    ETHER_HDR_SZ
                );
                assert_eq!(offsets.inner.ulp.as_ref().unwrap().seg_idx, 0);
                assert_eq!(
                    offsets.inner.ulp.as_ref().unwrap().seg_pos,
                    ETHER_HDR_SZ + ipv6_header_size
                );
                assert_eq!(
                    offsets.inner.ulp.as_ref().unwrap().pkt_pos,
                    ETHER_HDR_SZ + ipv6_header_size
                );
```

That's just my example, feel free to code that up differently if you think there is a cleaner way. Also probably should add check for the ether header offset just to be safe.

opte/src/engine/packet.rs
```diff
@@ -2622,4 +2620,48 @@ mod test {
...
+    // Verify that we correctly parse an IPv6 packet with extension headers
+    #[test]
+    fn parse_ipv6_extension_headers_ok() {
+        use crate::engine::ip6::test::generate_test_packet;
+        use crate::engine::ip6::test::SUPPORTED_EXTENSIONS;
+        use itertools::Itertools;
+        use smoltcp::wire::IpProtocol;
+        for n_extensions in 0..SUPPORTED_EXTENSIONS.len() {
+            for extensions in
+                SUPPORTED_EXTENSIONS.into_iter().permutations(n_extensions)
```
rzezeski, line 2633: This is certainly a nice way to automate testing various permutations, my only question is if the IPv6 spec allows arbitrary combinations? Like, is there any reason we should fail to parse a packet given a certain combination/ordering of extension headers?

opte/src/engine/packet.rs
bnaecker: Thanks for taking a look. Makes sense, I think I just didn't see the other tests that make similar assertions, and left it out of this one. I'll add those in now.

opte/src/engine/packet.rs
bnaecker: Good question. On my first reading of the [relevant section](https://www.rfc-editor.org/rfc/rfc2460#section-4.1) of the RFC, I focused on this:

> When more than one extension header is used in the same packet, it is
   recommended that those headers appear in the following order:

I took "recommended" to imply that this is just a recommendation, not a requirement for a conforming implementation. On closer reading, I think you're right, that there is a required order. Specifically, Hop-By-Hop headers are required to appear just after the base header (if they're present at all), and only once:

> IPv6 nodes must accept and attempt to process extension headers in
   any order and occurring any number of times in the same packet,
   except for the Hop-by-Hop Options header which is restricted to
   appear immediately after an IPv6 header only.

In that case, what's the right behavior should we receive a packet that doesn't conform? I imagine it's to drop it, but I saw elsewhere in the IPv6 spec that sometimes sending a special ICMP packet back to the sender is required. That's mentioned as a valid response to a Next Header that the receiver doesn't support, so I'm not sure it applies here.

rzezeski: Okay, so I read the RFC a bit and now I'm thinking we actually shouldn't care about order, and we should pass on packet with the extension headers as-is. If they are bogus, then either some router along the path or the destination itself will drop the packet and perhaps send an "ICMP Parameter Problem" message back in reply. OPTE is not really in the business of enforcing strictness of the headers. Yes, it does do some validation to make sure the headers are sane, but verifying order of extension headers feels like something else, at least for the guest packets.

bnaecker: We've agreed offline that we definitely don't want OPTE to be in the business of verifying EH order right now, and probably never. As @rzezeski notes, if a client of OPTE sends us a packet that has out-of-order headers, they're kinda on their own. Either they're IP stack is broken, or they're doing something funky / custom and it's up to them to make sure they'll get whatever replies they need, i.e., that the broken packets will make it to their destination and back.

rzezeski (approved):

PR #240, opened 2022-08-24
Fix fencepost error in IPv6 masking

bnaecker:


bnaecker:
This will fail the API version check, but I'm going to wait until it's approved to update that. I don't want to conflict with #238

rzezeski (approved):
Oh ranges...thanks for catching that!

bnaecker:
Completed in https://github.com/oxidecomputer/opte/commit/ef2f3bdf19edb9fa7c785c7753b67015d06d873f

Issue #160, opened 2022-06-14
soundness bug in `anymap`


davepacheco:
On omicron, dependabot [reported](https://github.com/oxidecomputer/omicron/security/dependabot/1) CVE-2021-38187, which looks like a soundness bug in "anymap".  There's no released fix yet, though it's been a while.  There's some mention of it [here](https://github.com/chris-morgan/anymap/issues/37).

I think it's coming into omicron from opte, here: 
https://github.com/oxidecomputer/opte/blob/12450552bdea4b3f97d309bfed5ad65dc2e2a775/opte/Cargo.toml#L47-L49

I'm not sure what the best plan is here.

rzezeski:
@bnaecker Thanks for bringing this to my attention. I'll look at this tomorrow and see what can be done.

rzezeski:
We solved half of the problem in #219. The other half is getting the layer metadata off of anymap. My feeling is it would be fine to just use a `String -> String` map. The main intent of the layer metadata is to allow one layer to pass on information to a downstream layer at runtime. For example, the VPC router layer passes the router target down to the overlay (encap) layer. Given that OPTE was my first major Rust project I was trying to make everything as strongly typed as possible, and thus I brought in anymap for this kind of safe, runtime casting. But in this case it's probably overkill, and anymap itself has had no maintenance for something like a year, with no sight of the original author nor anyone bothering to fork it.

rzezeski:
Addressed in a24dbb9bd627d882f8502fd92cca90e87ed61e7f.

PR #251, opened 2022-09-11
soundness bug in anymap (#160)

rzezeski:
- Replace the use of anymap with a map of String -> String.

- Rename the dubious `Meta` type to `ActionMeta` to better reflect its purpose.

- Now that a meatdata predicate is based on String we can derive `Predicate` equality instead of jumping through hoops.

- Removed some dead code around the EXT-IP hack exposed by this change.


bnaecker (approved):

rzezeski:
This was merged but GH didn't automatically catch it. Perhaps because I submitted this PR on top of another PR? Whatever.

Issue #214, opened 2022-08-09
One final pass in opte-drv


rzezeski:
I think there was one or two last things I wanted to pull out of opte-drv but cannot remember. Give it one last look and `rm -rf` it.

rzezeski:
Addressed in 19bf8d45589cd591f8763cff6d4e7c6676163be4.

PR #253, opened 2022-09-16
One final pass in opte-drv (#214)

rzezeski:
- Did a final sweep in opte-drv to see if there was anything else to migrate over to xde.

- Moved over the _Unwind_Resume hack. It seems xde has been working fine without it, but I also notice the Rust issue that originally prompted me to create this hack is still marked as open. So I thought it would be best to migrate the hack over to xde just as a safeguard against the problem coming back until that Rust issue is resolved.

  https://github.com/rust-lang/rust/issues/47493


xde/src/lib.rs
```diff
@@ -4,19 +4,25 @@
...
+// At some point we could hand write the stuff that is actually
+// needed, or come up with a better solution like using CTF data to
+// generate Rust types for only the stuff we need.
 #![allow(improper_ctypes)] // for long double -> u128
 #![allow(non_camel_case_types)] // for bindgen code in ip.rs
 #![allow(non_snake_case)] // for bindgen code in ip.rs
-#![deny(unused_must_use)]
 #![feature(alloc_error_handler)]
 #![feature(rustc_private)]
+#![deny(unused_must_use)]
```
bnaecker, line 25: Excellent :)

bnaecker (approved):
Nice, LGTM!

Issue #252, opened 2022-09-14
Regression in flow id handling causes it to be immutable


rzezeski:
In commit 70785d915875a3b3a3b453a90d8a7735c027da0d I regressed the intra-layer handling of the `InnerFlowId` value, making it immutable during layer processing. E.g., when a packet comes inbound the NAT layer rewrites it, but the firewall layer still sees the previous `InnerFlowId` value, before the NAT rewrite. I uncovered this issue while writing a new integration test for SNAT+ICMP as part of my work on #231. Furthermore, this regression was hidden in manual testing because of a second aspect of commit 70785d915875a3b3a3b453a90d8a7735c027da0d, which is that it pulled the default firewall rules out of the Oxide VPC implementation. Instead, it was decided that the implementation should provide a blank slate and it is up to Omicron to populate the default/initial rules. Because of this the firewall layer currently holds no rules, and when a layer holds no rules it has no action matches. When it has no action matches the packet simply passes through.

This is once again a lesson for all you kids out there to write integration tests when you can. I got lazy, and the gods have punished me for it. As part of #231 I'm also going to fix this regression, as well as add other NAT tests like SNAT+TCP and 1:1 NAT + ICMP/TCP.

rzezeski:
Fixed in 54f64c4145499c06ca67f504e49496ce6c95881b.

Issue #231, opened 2022-08-17
ICMP echo does not work over SNAT


rzezeski:
The `SNat4` action produces a panic when a guest tries to send outbound traffic destined for the Internet Gateway.

```
> ::status 
debugging crash dump /var/crash/kalm/vmcore.38 (64-bit) from kalm
operating system: 5.11 helios-1.0.21218 (i86pc)
build version: remotes/ci-build/xde-0-gafd7837f27

image uuid: 6f34b505-1a19-6370-b438-e580a85fe48e
panic message: panicked at 'called `Option::unwrap()` on a `None` value', /work/oxidecomputer/opte/opte/src/engine/headers.rs:538:48
dump content: kernel pages only

> $G
C++ symbol demangling enabled
> $C
fffffe007bce6a80 vpanic()
fffffe007bce6b00 rust_begin_unwind+0x8a()
fffffe007bce6b40 core::panicking::panic_fmt::hfcd69d41fa9fd79d+0x2d()
fffffe007bce6b90 core::panicking::panic::h710e604d40f28e68+0x44()
fffffe007bce6c20 opte::engine::rule::HT::run::h1b5372cbdc538114+0x47e()
fffffe007bce6ef0 opte::engine::layer::Layer::process_out_rules::h7320cffce6e1340f+0x9ad()
fffffe007bce7130 opte::engine::layer::Layer::process::h5ba95c0ee56bbf59+0x288()
fffffe007bce7480 opte::engine::port::Port::process::h0ab95889e3395613+0x2131()
fffffe007bce7c40 xde_mc_tx+0x24f()
fffffe007bce7c70 mac_ring_tx+0x43(fffffe596ae57000, 0, fffffe589a8dfd40)
fffffe007bce7cc0 mac_provider_tx+0x85(fffffe596ae57000, 0, fffffe589a8dfd40, fffffe58ee7a72c0)
fffffe007bce7d80 mac_tx+0x295(fffffe58ee7a72c0, fffffe589a8dfd40, 0, 1, 0)
fffffe007bce7e40 viona_tx+0x48c(fffffe59818c6c80, fffffe59818c6dc0)
fffffe007bce7ea0 viona_worker_tx+0x73(fffffe59818c6dc0, fffffe59818c6c80)
fffffe007bce7ef0 viona_worker+0x11b(fffffe59818c6dc0)
fffffe007bce7f00 thread_start+0xb()

> fffffe589a8dfd40::list mblk_t b_cont|::mblk
[...]

PR #257, opened 2022-09-21
PR: ICMP echo does not work over SNAT (#231)

rzezeski:
In order for ICMP Echo to work over SNAT, we must map the Echo Identifier into the SNAT port space. There are two ways to do this:

1. Map ICMP message bodies into a pseudo-ULP header and allow them to be acted upon just like other header data.

2. Add support for body transforms (like VFP's "action context").

OPTE used to implement option (1), but there were some issues in the past that caused me to remove that behavior. It might be good to reintroduce that behavior at some point, but instead I opted to take option (2). We might eventually need body transforms for other reasons, so I thought it would be good to at least see what it might look like. This implementation of body transforms is simple: it only allows transforming of bytes, it doesn't allow a change in body length (removing or adding bytes). So while this works for the purposes of implementing ICMP/SNAT rewrite, it may need some further development for future, more advanced body transformations.

This work only supports outbound ICMP Echo (and associated inbound reply). There is more ICMP support we need to figure out:

- A DU reply to an ICMP Echo over SNAT. This should be easy enough to deal with in the body transform.

- A DU reply to a TCP/UDP flow over SNAT or NAT. A harder problem to solves is dealing with DU replies (from some external network) to TCP/UDP traffic. The inbound flow id will be for ICMP and therefore will match nothing. We could allow all inbound ICMP DU traffic, but the user may explicitly block that traffic in the firewall, and we need to honor that. Rather, we need to be smart enough to know there is a stateful firewall flow and that we need to allow inbound ICMP DU for active flows. It could be that we reach into the ICMP DU body and use that data to build its flow id, as in some sense **that is** its real flow id -- it's a response to **that** flow.

Other Work
----------

- Regression in flow id handling causes it to be immutable (#252)

- Added ability to snapshot (clone) kstats. This is useful for verifying stat values in integration tests.

### DIFF
--- a/dtrace/opte-layer-process.d
+++ b/dtrace/opte-layer-process.d
@@ -9,10 +9,11 @@
 #include "common.h"
 #include "protos.d"
 
-#define	HDR_FMT		"%-16s %-16s %-3s %-48s %s\n"
+#define	HDR_FMT		"%-16s %-16s %-3s %-48s $-48s %s\n"
 
 BEGIN {
-	printf(HDR_FMT, "PORT", "LAYER", "DIR", "FLOW", "RES");
+	printf(HDR_FMT, "PORT", "LAYER", "DIR", "FLOW BEFORE", "FLOW AFTER",
+	    "RES");
 	num = 0;
 }
 
@@ -20,11 +21,13 @@ layer-process-return {
 	this->dir = stringof(arg0);
 	this->port = stringof(arg1);
 	this->layer = stringof(arg2);
-	this->flow = (flow_id_sdt_arg_t *)arg3;
-	this->res = stringof(arg4);
+	this->flow_before = (flow_id_sdt_arg_t *)arg3;
+	this->flow_after = (flow_id_sdt_arg_t *)arg4;
+	this->res = stringof(arg5);
 
 	if (num >= 10) {
-		printf(HDR_FMT, "PORT", "LAYER", "DIR", "FLOW", "RES");
+		printf(HDR_FMT, "PORT", "LAYER", "DIR", "FLOW BEFORE",
+		    "FLOW AFTER", "RES");
 		num = 0;
 	}
 
@@ -36,13 +39,17 @@ layer-process-return {
 }
 
 layer-process-return /this->af == AF_INET/ {
-	FLOW_FMT(this->s, this->flow);
-	printf(HDR_FMT, this->port, this->layer, this->dir, this->s, this->res);
+	FLOW_FMT(this->s_before, this->flow_before);
+	FLOW_FMT(this->s_after, this->flow_after);
+	printf(HDR_FMT, this->port, this->layer, this->dir, this->s_before,
+	    this->s_after, this->res);
 	num++;
 }
 
 layer-process-return /this->af == AF_INET6/ {
-	FLOW_FMT6(this->s, this->flow);
-	printf(HDR_FMT, this->port, this->layer, this->dir, this->s, this->res);
+	FLOW_FMT6(this->s_before, this->flow_before);
+	FLOW_FMT6(this->s_after, this->flow_after);
+	printf(HDR_FMT, this->port, this->layer, this->dir, this->s_before,
+	    this->s_after, this->res);
 	num++;
 }
--- a/dtrace/opte-port-process.d
+++ b/dtrace/opte-port-process.d
@@ -6,31 +6,29 @@
 #include "common.h"
 #include "protos.d"
 
-#define	HDR_FMT		"%-12s %-3s %-8s %-43s %-5s %s\n"
-#define	LINE_FMT	"%-12s %-3s %-8u %-43s %-5u %s\n"
+#define	HDR_FMT		"%-12s %-3s %-8s %-43s %-43s %-5s %s\n"
+#define	LINE_FMT	"%-12s %-3s %-8u %-43s %-43s %-5u %s\n"
 
 BEGIN {
-	/*
-	 * Use an associative array to stringify the protocol number.
-	 */
-
-	printf(HDR_FMT, "NAME", "DIR", "EPOCH", "FLOW", "LEN", "RESULT");
+	printf(HDR_FMT, "NAME", "DIR", "EPOCH", "FLOW BEFORE", "FLOW AFTER",
+	    "LEN", "RESULT");
 	num = 0;
 }
 
 port-process-return {
 	this->dir = stringof(arg0);
 	this->name = stringof(arg1);
-	this->flow = (flow_id_sdt_arg_t *)arg2;
-	this->epoch = arg3;
-	this->mp = (mblk_t *)arg4;
+	this->flow_before = (flow_id_sdt_arg_t *)arg2;
+	this->flow_after = (flow_id_sdt_arg_t *)arg3;
+	this->epoch = arg4;
+	this->mp = (mblk_t *)arg5;
 	/* If the result is a hairpin packet, then hp_mp is non-NULL. */
-	this->hp_mp = (mblk_t *)arg5;
-	this->res = stringof(arg6);
+	this->hp_mp = (mblk_t *)arg6;
+	this->res = stringof(arg7);
 
 	if (num >= 10) {
-		printf(HDR_FMT, "NAME", "DIR", "EPOCH", "FLOW", "LEN",
-		    "RESULT");
+		printf(HDR_FMT, "NAME", "DIR", "EPOCH", "FLOW BEFORE",
+		    "FLOW AFTER", "LEN", "RESULT");
 		num = 0;
 	}
 
@@ -42,16 +40,18 @@ port-process-return {
 }
 
 port-process-return /this->af == AF_INET/ {
-	FLOW_FMT(this->s, this->flow);
-	printf(LINE_FMT, this->name, this->dir, this->epoch, this->s,
-	    msgsize(this->mp), this->res);
+	FLOW_FMT(this->s_before, this->flow_before);
+	FLOW_FMT(this->s_after, this->flow_after);
+	printf(LINE_FMT, this->name, this->dir, this->epoch, this->s_before,
+	    this->s_after, msgsize(this->mp), this->res);
 	num++;
 }
 
 port-process-return /this->af == AF_INET6/ {
-	FLOW_FMT6(this->s, this->flow);
-	printf(LINE_FMT,  this->name, this->dir, this->epoch, this->s,
-	    msgsize(this->mp), this->res);
+	FLOW_FMT6(this->s_before, this->flow_before);
+	FLOW_FMT6(this->s_after, this->flow_after);
+	printf(LINE_FMT,  this->name, this->dir, this->epoch, this->s_before,
+	    this->s_after, msgsize(this->mp), this->res);
 	num++;
 }
 
--- a/dtrace/usdt-opte-layer-process.d
+++ b/dtrace/usdt-opte-layer-process.d
@@ -3,26 +3,29 @@
  *
  * dtrace -ZCqs ./usdt-opte-layer-process.d
  */
-#define	HDR_FMT "%-16s %-16s %-3s %-48s %s\n"
+#define	HDR_FMT "%-16s %-16s %-3s %-48s %-48s %s\n"
 
 BEGIN {
-	printf(HDR_FMT, "PORT", "LAYER", "DIR", "FLOW", "RES");
+	printf(HDR_FMT, "PORT", "LAYER", "DIR", "FLOW BEFORE", "FLOW AFTER",
+	    "RES");
 	num = 0;
 }
 
 layer-process-return {
-	this->dir = json(copyinstr(arg0), "ok");
-	this->port = copyinstr(arg1);
-	this->layer = copyinstr(arg2);
-	this->flow = copyinstr(arg3);
+	this->dir = json(copyinstr(arg0), "ok.0");
+	this->port = json(copyinstr(arg0), "ok.1");
+	this->layer = copyinstr(arg1);
+	this->flow_before = copyinstr(arg2);
+	this->flow_after = copyinstr(arg3);
 	this->res = copyinstr(arg4);
 	num++;
 
-	printf(HDR_FMT, this->port, this->layer, this->dir, this->flow,
-	    this->res);
+	printf(HDR_FMT, this->port, this->layer, this->dir, this->flow_before,
+	    this->flow_after, this->res);
 
 	if (num >= 10) {
-		printf(HDR_FMT, "PORT", "LAYER", "DIR", "FLOW", "RES");
+		printf(HDR_FMT, "PORT", "LAYER", "DIR", "FLOW BEFORE",
+		    "FLOW AFTER", "RES");
 		num = 0;
 	}
 }
--- a/opte/src/ddi/kstat.rs
+++ b/opte/src/ddi/kstat.rs
@@ -58,6 +58,7 @@ cfg_if! {
 /// To register a provider see [`KStatNamed`].
 pub trait KStatProvider {
     const NUM_FIELDS: u32;
+    type Snap;
 
     fn init(&mut self) -> Result<(), Error>;
 
@@ -66,6 +67,10 @@ pub trait KStatProvider {
     fn num_fields(&self) -> u32 {
         Self::NUM_FIELDS
     }
+
+    /// Return a snapshot of the stats. This is how you obtain a copy,
+    /// as opposed to the traditional clone().
+    fn snapshot(&self) -> Self::Snap;
 }
 
 /// Initialize and register a [`KStatProvider`].
@@ -209,6 +214,10 @@ impl KStatU64 {
     pub fn new() -> Self {
         Self { inner: kstat_named_t::new() }
     }
+
+    pub fn val(&self) -> u64 {
+        self.inner.val_u64()
+    }
 }
 
 #[cfg(all(not(feature = "std"), not(test)))]
@@ -231,6 +240,10 @@ impl KStatU64 {
 
     pub fn new() -> Self {
         Self { value: 0 }
+    }
+
+    pub fn val(&self) -> u64 {
+        self.value
     }
 }
 
--- a/opte/src/engine/layer.rs
+++ b/opte/src/engine/layer.rs
@@ -5,25 +5,25 @@
 // Copyright 2022 Oxide Computer Company
 
 use super::flow_table::{FlowTable, FlowTableDump};
-use super::headers::{IpAddr, IpMeta, UlpMeta};
 use super::ioctl;
-use super::ip4::Protocol;
-use super::packet::{Initialized, Packet, PacketMeta, PacketRead, Parsed};
+use super::packet::{
+    BodyTransformError, Initialized, InnerFlowId, Packet, PacketMeta,
+    PacketRead, Parsed, FLOW_ID_DEFAULT,
+};
 use super::port::meta::ActionMeta;
+use super::port::Transforms;
 use super::rule::{
     self, flow_id_sdt_arg, ht_probe, Action, ActionDesc, AllowOrDeny,
-    Finalized, HdrTransform, HdrTransformError, Rule, RuleDump,
+    Finalized, GenBtError, HdrTransformError, Rule, RuleDump,
 };
 use crate::ddi::time::Moment;
 use crate::{ExecCtx, LogLevel};
 use core::fmt::{self, Display};
-use core::mem;
 use core::num::NonZeroU32;
 use core::result;
 use cstr_core::CString;
 use illumos_sys_hdrs::c_char;
-use opte_api::{Direction, Ipv4Addr};
-use serde::{Deserialize, Serialize};
+use opte_api::Direction;
 
 cfg_if! {
     if #[cfg(all(not(feature = "std"), not(test)))] {
@@ -40,17 +40,31 @@ cfg_if! {
 
 #[derive(Debug)]
 pub enum LayerError {
+    BodyTransform(BodyTransformError),
     FlowTableFull { layer: String, dir: Direction },
     GenDesc(rule::GenDescError),
-    GenHt(rule::GenHtError),
+    GenBodyTransform(GenBtError),
+    GenHdrTransform(rule::GenHtError),
     GenPacket(rule::GenErr),
     HeaderTransform(HdrTransformError),
     ModMeta(String),
 }
 
+impl From<GenBtError> for LayerError {
+    fn from(e: GenBtError) -> Self {
+        Self::GenBodyTransform(e)
+    }
+}
+
 impl From<HdrTransformError> for LayerError {
     fn from(e: HdrTransformError) -> Self {
         Self::HeaderTransform(e)
+    }
+}
+
+impl From<BodyTransformError> for LayerError {
+    fn from(e: BodyTransformError) -> Self {
+        Self::BodyTransform(e)
     }
 }
 
@@ -231,11 +245,13 @@ impl Layer {
     }
 
     /// Clear all flows from the layer's flow tables.
-    pub fn clear_flows(&mut self) {
+    pub(crate) fn clear_flows(&mut self) {
         self.ft.clear();
     }
 
-    pub fn dump(&self) -> ioctl::DumpLayerResp {
+    /// Dump the contents of this layer. This is used for presenting
+    /// the layer state in a human-friendly manner.
+    pub(crate) fn dump(&self) -> ioctl::DumpLayerResp {
         let rules_in = self.rules_in.dump();
         let rules_out = self.rules_out.dump();
         let ftd = self.ft.dump();
@@ -251,12 +267,12 @@ impl Layer {
     fn gen_desc_fail_probe(
         &self,
         dir: Direction,
-        ifid: &InnerFlowId,
+        flow: &InnerFlowId,
         err: &rule::GenDescError,
     ) {
         cfg_if! {
             if #[cfg(all(not(feature = "std"), not(test)))] {
-                let flow_id = flow_id_sdt_arg::from(ifid);
+                let flow_arg = flow_id_sdt_arg::from(flow);
                 let dir_c = CString::new(format!("{}", dir)).unwrap();
                 let msg_c = CString::new(format!("{:?}", err)).unwrap();
 
@@ -265,22 +281,21 @@ impl Layer {
                         self.port_c.as_ptr() as uintptr_t,
                         self.name_c.as_ptr() as uintptr_t,
                         dir_c.as_ptr() as uintptr_t,
-                        &flow_id as *const flow_id_sdt_arg as uintptr_t,
+                        &flow_arg as *const flow_id_sdt_arg as uintptr_t,
                         msg_c.as_ptr() as uintptr_t,
                     );
                 }
             } else if #[cfg(feature = "usdt")] {
                 let port_s = self.port_c.to_str().unwrap();
                 let name_s = self.name_c.to_str().unwrap();
-                let flow_s = ifid.to_string();
+                let flow_s = flow.to_string();
                 let msg_s = format!("{:?}", err);
 
                 crate::opte_provider::gen__desc__fail!(
                     || (port_s, name_s, dir, flow_s, msg_s)
                 );
             } else {
-                let (_, _, _, _, _) =
-                    (&self.port_c, &self.name_c, dir, ifid, err);
+                let (..) = (&self.port_c, &self.name_c, dir, flow, err);
             }
         }
     }
@@ -288,12 +303,12 @@ impl Layer {
     fn gen_ht_fail_probe(
         &self,
         dir: Direction,
-        ifid: &InnerFlowId,
+        flow: &InnerFlowId,
         err: &rule::GenHtError,
     ) {
         cfg_if! {
             if #[cfg(all(not(feature = "std"), not(test)))] {
-                let flow_id = flow_id_sdt_arg::from(ifid);
+                let flow_arg = flow_id_sdt_arg::from(flow);
                 let dir_c = CString::new(format!("{}", dir)).unwrap();
                 let msg_c = CString::new(format!("{:?}", err)).unwrap();
 
@@ -302,29 +317,31 @@ impl Layer {
                         self.port_c.as_ptr() as uintptr_t,
                         self.name_c.as_ptr() as uintptr_t,
                         dir_c.as_ptr() as uintptr_t,
-                        &flow_id as *const flow_id_sdt_arg as uintptr_t,
+                        &flow_arg as *const flow_id_sdt_arg as uintptr_t,
                         msg_c.as_ptr() as uintptr_t,
                     );
                 }
             } else if #[cfg(feature = "usdt")] {
                 let port_s = self.port_c.to_str().unwrap();
-                let flow_s = ifid.to_string();
+                let flow_s = flow.to_string();
                 let err_s = format!("{:?}", err);
 
                 crate::opte_provider::gen__ht__fail!(
                     || (port_s, &self.name, dir, flow_s, err_s)
                 );
             } else {
-                let (_, _, _) = (dir, ifid, err);
+                let (..) = (dir, flow, err);
             }
         }
     }
 
-    pub fn expire_flows(&mut self, now: Moment) {
+    /// Expire all flows whose TTL has been reached based on the
+    /// passed in moment.
+    pub(crate) fn expire_flows(&mut self, now: Moment) {
         self.ft.expire_flows(now);
     }
 
-    pub fn layer_process_entry_probe(
+    pub(crate) fn layer_process_entry_probe(
         &self,
         dir: Direction,
         ifid: &InnerFlowId,
@@ -358,7 +375,8 @@ impl Layer {
     fn layer_process_return_probe(
         &self,
         dir: Direction,
-        ifid: &InnerFlowId,
+        flow_before: &InnerFlowId,
+        flow_after: &InnerFlowId,
         res: &result::Result<LayerResult, LayerError>,
     ) {
         cfg_if! {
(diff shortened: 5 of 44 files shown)


rzezeski:
I still need to test this.

opte/src/engine/packet.rs
```diff
@@ -1258,10 +1305,89 @@ impl Packet<Parsed> {
...
     }
 
+    /// Run the [`BodyTransform`] against this packet.
+    pub fn body_transform(
+        &mut self,
+        dir: Direction,
+        xform: &Box<dyn BodyTransform>,
+    ) -> Result<(), BodyTransformError> {
+        // We set the flag now with the assumption that the transform
+        // couild fail after modifying part of the body. In the future
```
bnaecker, line 1315: Nit spelling: `couild` -> `could`.

opte/src/engine/packet.rs
```diff
@@ -1241,6 +1247,47 @@ impl Packet<Initialized> {
...
+pub trait BodyTransform: fmt::Display + DynClone {
+    /// Execute the body transformation. The body segments include
+    /// **only** body data, starting directly after the end of the ULP
+    /// header.
+    ///
+    /// # Errors
+    ///
+    /// The transformation can choose to return a
+    /// [`BodyTransformError`] at any time if the body is not
+    /// acceptable. On error, none or some of the bytes may have been
```
bnaecker, line 1266: Is there an argument to be made for running the transform on a copy of the data, and only swapping it into the original if the transform succeeds?

opte/src/engine/packet.rs
```diff
@@ -1258,10 +1305,89 @@ impl Packet<Parsed> {
...
+                body_segs.push(seg.slice());
+            }
+        }
+
+        Some(body_segs)
+    }
+
+    /// Return a list of mutable body segments, or `None` if there is
+    /// no body.
+    pub fn body_segs_mut<'a>(&'a mut self) -> Option<Vec<&'a mut [u8]>> {
```
bnaecker, line 1364: Nit: I _think_ the lifetime can be elided here, since there's only one, but I could be wrong.

opte/src/engine/packet.rs
```diff
@@ -1433,9 +1559,13 @@ impl Packet<Parsed> {
                             .unwrap();
                         let pseudo_csum = ip.pseudo_csum();
                         csum += pseudo_csum;
-                        csum.add(
-                            self.segs[0].slice(tcp_off.seg_pos, tcp.hdr_len()),
-                        );
+                        // Safety: We are slicing with the parse data.
+                        csum.add(unsafe {
```
bnaecker, line 1563: Why use the unsafe / unchecked method here?

opte/src/engine/packet.rs
```diff
@@ -1707,11 +1840,60 @@ impl PacketSeg {
...
+    ///
+    /// It is up to the caller to ensure that `offset` and `offset +
+    /// len` reside within the segment boundaries.
+    unsafe fn slice_unchecked(
+        &self,
+        offset: usize,
+        len: Option<usize>,
+    ) -> &[u8] {
+        let start = (*self.mp).b_rptr.add(offset);
+        slice::from_raw_parts(start, len.unwrap_or(self.len))
```
bnaecker, line 1875: The length argument to `from_raw_parts` seems wrong, unless I'm missing something. I would expect that to be something like `len.unwrap_or(self.len - offset)`. If I provide arguments of `1` and `None`, then the slice will extend beyond the limit of the underlying `mblk`, right?

Also, that subtraction could be something like `self.len.saturating_sub(offset)` to ensure that we never panic on underflow. We'd rather get an empty slice I think.

opte/src/engine/packet.rs
```diff
@@ -1707,11 +1840,60 @@ impl PacketSeg {
...
+    ///
+    /// It is up to the caller to ensure that `offset` and `offset +
+    /// len` reside within the segment boundaries.
+    unsafe fn slice_mut_unchecked(
+        &mut self,
+        offset: usize,
+        len: Option<usize>,
+    ) -> &mut [u8] {
+        let start = (*self.mp).b_rptr.add(offset);
+        slice::from_raw_parts_mut(start, len.unwrap_or(self.len))
```
bnaecker, line 1896: Same comment as above applies here.

opte/src/engine/rule.rs
```diff
@@ -1276,6 +1290,18 @@ impl From<smoltcp::Error> for GenErr {
 
 pub type GenPacketResult = ActionResult<Packet<Initialized>, GenErr>;
 
+/// An error while generating a [`BodyTrannsform`].
```
bnaecker, line 1293: Nit spelling: `BodyTrannsform` -> `BodyTransform`.

opte/src/engine/tcp.rs
```diff
@@ -278,7 +278,7 @@ impl TcpHdr {
             seq: 0,
             ack: 0,
             hdr_len_bytes: 20,
-            flags: 0x02,
+            flags: 0x00,
```
bnaecker, line 281: Seems out of place, just checking you intended to keep this change.

oxide-vpc/src/api.rs
```diff
@@ -47,7 +47,7 @@ pub struct BoundaryServices {
     pub mac: MacAddr,
 }
 
-/// The IPv4 configuration for an OPTE port.
+/// The IPv4 configuration of a VPC guest..
```
bnaecker, line 50: Nit, double periods at the end of the sentence.

oxide-vpc/tests/integration_tests.rs
```diff
@@ -1602,6 +1617,238 @@ fn guest_to_internet() {
...
+            "fw.flows_out",
+            "fw.flows_in",
+            "nat.flows_out",
+            "nat.flows_in",
+            "uft.flows_out"
+        ]
+    );
+
+    // Ether + IPv6 + UDP + Geneve + Ether + IPv4
+    assert_eq!(pkt1.body_offset(), 14 + 40 + 8 + 8 + 14 + 20);
```
bnaecker, line 1672: Nit: Might want to use the `const` sizes we define for these?

oxide-vpc/tests/integration_tests.rs
```diff
@@ -1602,6 +1617,238 @@ fn guest_to_internet() {
...
+        seq_no,
+        &data[..],
+    );
+
+    let res = g1.port.process(In, &mut pkt2, &mut ameta);
+    assert!(matches!(res, Ok(Modified)), "bad result: {:?}", res);
+    incr!(g1, ["uft.flows_in"]);
+
+    // Inner Ether + IP
+    assert_eq!(pkt2.body_offset(), 14 + 20);
```
bnaecker, line 1719: Same nit here.

oxide-vpc/tests/integration_tests.rs
```diff
@@ -1602,6 +1617,238 @@ fn guest_to_internet() {
...
+        &data[..],
+    );
+
+    assert_eq!(g1.port.stats_snap().out_uft_hit, 0);
+    let res = g1.port.process(Out, &mut pkt3, &mut ameta);
+    assert!(matches!(res, Ok(Modified)), "bad result: {:?}", res);
+    assert_port!(g1);
+
+    // Ether + IPv6 + UDP + Geneve + Ether + IPv4
+    assert_eq!(pkt3.body_offset(), 14 + 40 + 8 + 8 + 14 + 20);
```
bnaecker, line 1770: Same nit here.

oxide-vpc/tests/integration_tests.rs
```diff
@@ -1602,6 +1617,238 @@ fn guest_to_internet() {
...
+        &data[..],
+    );
+
+    assert_eq!(g1.port.stats_snap().in_uft_hit, 0);
+    let res = g1.port.process(In, &mut pkt4, &mut ameta);
+    assert!(matches!(res, Ok(Modified)), "bad result: {:?}", res);
+    assert_port!(g1);
+
+    // Inner Ether + IP
+    assert_eq!(pkt4.body_offset(), 14 + 20);
```
bnaecker, line 1821: Same nit here.

oxide-vpc/tests/integration_tests.rs
```diff
@@ -1602,6 +1617,238 @@ fn guest_to_internet() {
...
+            assert_eq!(ip4.dst, g1_cfg.ipv4().private_ip);
+            assert_eq!(ip4.proto, Protocol::ICMP);
+        }
+
+        ip6 => panic!("execpted inner IPv4 metadata, got IPv6: {:?}", ip6),
+    }
+
+    let body = pkt4.body_segs().unwrap()[0];
+    let icmp = Icmpv4Packet::new_checked(body).unwrap();
+    assert_eq!(icmp.echo_ident(), ident);
```
bnaecker, line 1847: This is really a question, but I want to make sure I understand what's going on here.

This asserts that the _incoming_ reply has been modified to use the original `ident` from the outbound echo request, right? So the full scenario is:

- Send an ICMP echo request with some ident
- On the way out, that's modified to use one of the mapped ports from the SNAT configuration
- On the way back in, we're looking for that mapped port, and we modify the packet again from the mapped port back to the original `ident`.

Is that all accurate?

opte/src/engine/port.rs
```diff
@@ -1059,13 +1026,44 @@ impl Port {
 
         Err(OpteError::LayerNotFound(layer_name.to_string()))
     }
+
+    /// Grab a snapshot of the port statistics.
+    pub fn stats_snap(&self) -> PortStatsSnap {
+        self.data.lock().stats.vals.snapshot()
```
bnaecker, line 1032: The locking here makes me a bit worried about deadlocks. I might at least put something in the comment that this acquires the lock and can't be called from any other function that also does so. AFAICT, this is only used in the tests to check that the stats behave as we expect, but still might be a good idea.

opte/src/engine/port.rs
```diff
@@ -1193,17 +1194,25 @@ impl Port {
...
+                        (dir, self.name.as_str()),
+                        (flow_b_s.as_ref(), flow_a_s.as_ref()),
+                        epoch,
+                        pkt.mblk_addr(),
+                        res_str
+                    )
                 );
             } else {
-                let (_, _, _, _, _) = (dir, ifid, epoch, pkt, res);
+                let (_, _, _, _, _, _) =
```
bnaecker, line 1214: I learned recently that you can use `(..)` here to collect all remaining items. Kinda cool.

opte/src/engine/port.rs
```diff
@@ -1334,15 +1343,17 @@ impl Port {
     fn process_in_miss(
         &self,
         data: &mut PortData,
-        ufid_in: &InnerFlowId,
+        ufid_in: &mut InnerFlowId,
```
bnaecker, line 1346: The same comment I made earlier about taking an exclusive reference and modifying applies here. Kind of a style thing, I realize, so maybe not worth the time right now.

opte/src/engine/snat.rs
```diff
@@ -37,30 +44,30 @@ cfg_if! {
 /// A single entry in the NAT pool, describing the public IP and port used to
 /// NAT a private address.
 #[derive(Clone, Copy)]
-pub struct NatPoolEntry {
-    ip: IpAddr,
+pub struct NatPoolEntry<T: ConcreteIpAddr> {
```
bnaecker, line 47: Nice, this is a great way to implement this per-protocol.

opte/src/engine/snat.rs
```diff
@@ -295,109 +418,204 @@ impl ActionDesc for SNatDesc {
...
+    priv_ip: Ipv4Addr,
+    echo_ident: u16,
+}
+
+pub const SNAT_ICMP_ECHO_NAME: &'static str = "SNAT_ICMP_ECHO";
+
+impl ActionDesc for SNatIcmpEchoDesc {
+    fn gen_ht(&self, dir: Direction) -> HdrTransform {
+        match dir {
+            // Outbound traffic needs it's source IP and source port
```
bnaecker, line 440: Spelling nit: `it's` -> `its`.

opte/src/engine/snat.rs
```diff
@@ -295,109 +418,204 @@ impl ActionDesc for SNatDesc {
...
+        self.pool.release(&self.priv_ip.into(), self.nat);
+    }
+}
+
+/// Perform SNAT for ICMP Echo/Reply messages, treating the Identifier
+/// as a source port.
+#[derive(Clone)]
+pub struct SNatIcmpEchoBt {
+    ident: u16,
+    nat: NatPoolEntry<Ipv4Addr>,
```
bnaecker, line 485: I don't think we need it now, but this would be needed for ICMPv6 echoes too, right?

opte/src/engine/snat.rs
```diff
@@ -295,109 +418,204 @@ impl ActionDesc for SNatDesc {
...
+        use crate::engine::ip4::{Ipv4Hdr, Protocol, UlpCsumOpt};
+        use crate::engine::tcp::TcpHdr;
         use opte_api::{Ipv4Addr, MacAddr};
 
         let priv_mac = MacAddr::from([0x02, 0x08, 0x20, 0xd8, 0x35, 0xcf]);
         let dest_mac = MacAddr::from([0x78, 0x23, 0xae, 0x5d, 0x4f, 0x0d]);
-        let priv_ipv4: Ipv4Addr = "10.0.0.220".parse().unwrap();
-        let priv_ip = IpAddr::from(priv_ipv4);
+        let priv_ip: Ipv4Addr = "10.0.0.220".parse().unwrap();
+        // let priv_ip = IpAddr::from(priv_ipv4);
```
bnaecker, line 571: Remove dead code?

bnaecker (commented):
This is a ton of great work, thanks! The body transforms look very useful. I've a bunch of nits, with a few more substantive questions or concerns, but overall looks awesome.

dtrace/opte-layer-process.d
```diff
@@ -9,22 +9,25 @@
 #include "common.h"
 #include "protos.d"
 
-#define	HDR_FMT		"%-16s %-16s %-3s %-48s %s\n"
+#define	HDR_FMT		"%-16s %-16s %-3s %-48s $-48s %s\n"
 
 BEGIN {
-	printf(HDR_FMT, "PORT", "LAYER", "DIR", "FLOW", "RES");
+	printf(HDR_FMT, "PORT", "LAYER", "DIR", "FLOW BEFORE", "FLOW_AFTER",
```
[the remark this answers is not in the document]
rzezeski: Yea there is supposed to be no underscore. Fixed.

dtrace/opte-port-process.d
```diff
@@ -6,31 +6,33 @@
...
 
-#define	HDR_FMT		"%-12s %-3s %-8s %-43s %-5s %s\n"
-#define	LINE_FMT	"%-12s %-3s %-8u %-43s %-5u %s\n"
+#define	HDR_FMT		"%-12s %-3s %-8s %-43s %-43s %-5s %s\n"
+#define	LINE_FMT	"%-12s %-3s %-8u %-43s %-43s %-5u %s\n"
 
 BEGIN {
 	/*
 	 * Use an associative array to stringify the protocol number.
 	 */
```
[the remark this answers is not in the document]
rzezeski: Removed.

dtrace/usdt-opte-layer-process.d
```diff
@@ -3,26 +3,29 @@
...
 
-	printf(HDR_FMT, this->port, this->layer, this->dir, this->flow,
-	    this->res);
+	printf(HDR_FMT, this->port, this->layer, this->dir, this->flow_before,
+	    this->flow_after, this->res);
 
 	if (num >= 10) {
-		printf(HDR_FMT, "PORT", "LAYER", "DIR", "FLOW", "RES");
+		printf(HDR_FMT, "PORT", "LAYER", "DIR", "FLOW BEFORE",
+		    "FLOW AFTTER", "RES");
```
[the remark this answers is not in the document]
rzezeski: Fixed.

opte/src/ddi/kstat.rs
```diff
@@ -66,6 +67,10 @@ pub trait KStatProvider {
     fn num_fields(&self) -> u32 {
         Self::NUM_FIELDS
     }
+
+    /// Return a snapshot of the stats. This is how you obtain a copy,
+    /// as opposed to the tradition clone().
```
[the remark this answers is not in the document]
rzezeski: Fixed.

opte/src/engine/layer.rs
```diff
@@ -369,32 +391,36 @@ impl Layer {
                     Ok(v) => format!("{}", v),
```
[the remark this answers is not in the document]
rzezeski: Yes, I've been meaning to write up an issue about this. I'll do that now. I'm not sure about things like lazy_static in-kernel as that will bring in stuff like spin-rs. We should stick to synchronization objects provided by the kernel.

opte/src/engine/layer.rs
```diff
@@ -369,32 +391,36 @@ impl Layer {
...
-                        &ifid_arg as *const flow_id_sdt_arg as uintptr_t,
+                        &flow_b_arg as *const flow_id_sdt_arg as uintptr_t,
+                        &flow_a_arg as *const flow_id_sdt_arg as uintptr_t,
                         res_c.as_ptr() as uintptr_t,
                     );
                 }
             } else if #[cfg(feature = "usdt")] {
                 let port_s = self.port_c.to_str().unwrap();
-                let ifid_s = ifid.to_string();
+                let flow_b_s = flow_before.to_string();
```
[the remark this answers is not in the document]
rzezeski: For USDT I don't care at all. They are only there to allow tracking probes when running the unit/integration tests.

opte/src/engine/layer.rs
```diff
@@ -424,83 +450,93 @@ impl Layer {
...
 
         // Do we have a FlowTable entry? If so, use it.
         if let Some(desc) = self.ft.get_in(&ifid) {
+            let ifid_before = ifid.clone();
             let ht = desc.gen_ht(Direction::In);
-            hts.push(ht.clone());
             ht.run(pkt.meta_mut())?;
-            let ifid_after = InnerFlowId::from(pkt.meta());
+            xforms.hdr.push(ht);
+            *ifid = InnerFlowId::from(pkt.meta());
```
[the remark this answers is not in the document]
rzezeski: As we talked about offline, this was me thinking too much in terms of C ABI. I was then going to have this as argument that takes ownership and return new value, but then I realized this really should just live in Packet itself. And anytime something runs a header transformation, the packet's flow should be updated to reflect the change. So I've done that now. However, while working on that I found some other bugs. I'm undecided on if I'll keep holding up this PR or just do a follow up.

opte/src/engine/layer.rs
```diff
@@ -554,19 +590,20 @@ impl Layer {
...
+                        return Err(LayerError::GenHdrTransform(e));
                     }
                 };
 
-                hts.push(ht.clone());
+                let ifid_before = ifid.clone();
                 ht.run(pkt.meta_mut())?;
-                let ifid_after = InnerFlowId::from(pkt.meta());
+                xforms.hdr.push(ht);
+                *ifid = InnerFlowId::from(pkt.meta());
```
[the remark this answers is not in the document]
rzezeski: See reply to first comment about this.

opte/src/engine/layer.rs
```diff
@@ -609,21 +655,16 @@ impl Layer {
...
-                    &ifid_after,
                 );
 
-                self.ft.add_pair(desc, ifid.clone(), out_ifid);
-
-                // if let Some(ctx) = ra_in.ctx {
-                //     ctx.exec(flow_id, &mut [0; 0]);
-                // }
-
+                self.ft.add_pair(desc, ifid_before.clone(), out_ifid);
```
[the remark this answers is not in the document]
rzezeski: See reply to first comment about this.

opte/src/engine/packet.rs
rzezeski: OPTE should avoid copying the body. It's okay that the modification status is unknown on failure, the packet is going to be dropped anyways.

opte/src/engine/packet.rs
rzezeski: Because it's the only option available here, we don't know at compile time that the offset is valid (though practically speaking we know it's fine because it came from the parsed data). Anyways, this is moot now. I changed these functions to no longer be unsafe, but instead panic if you slice out of bound of the segment.

opte/src/engine/packet.rs
rzezeski: You are correct, this was wrong. I dabbled with your idea, but we also had the issue that offset could be past the end of the segment. Then I dabbled with the idea of silently pinning `offset` to `b_wptr` when its too large, and pinning `offset + len` to `b_wptr` when it goes past, effectively making the function always safe...but then I realized this is a bad idea because it's no longer correct, and it's pretty unsporting for it to silently reduce the size of the expected slice (and this would also result in very silent/hard-to-debug issues in production). So yea, I opted for panic. If we can to eliminate all possibility of panic either the parsing/packet code needs to change to avoid this slicing all together or we make a version that returns Option/Result and deal with that.

opte/src/engine/packet.rs
rzezeski: See reply to other comment.

opte/src/engine/tcp.rs
rzezeski: This was on purpose: I figured the flags should default to 0, not SYN (and it was causing some test failures IIRC).

oxide-vpc/tests/integration_tests.rs
rzezeski: Yes, the steps you described are exactly the purpose of the SNAT/ICMP body transform. On outbound we need to map the Echo Identifier into the SNAT port space. On inbound we need to do the reverse, and map it back to the Echo Identifier that the guest is expecting to see.

opte/src/engine/port.rs
rzezeski: Yea this is nice. I'll start changing places as I come across  them.

opte/src/engine/snat.rs
rzezeski: I'm 50/50 if I like this, honestly. Don't get me wrong, generics are great, I used them in places. But I'm undecided on their value in this case as you still end up with lots of duplicated code but some of it becomes a little more convoluted because of the generics. I'm obviously leaving this as-is. But just felt like giving my 2 cents.

opte/src/engine/snat.rs
rzezeski: Yea I decided to just do ICMPv4 here. And I'm not sure if I would make this generic. I'd probably just make a separate type. But I imagine you'll be implementing the ICMPv6 so you can decide.

opte/src/engine/layer.rs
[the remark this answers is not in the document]
bnaecker: > this really should just live in Packet itself

Nice, that makes a lot of sense. It's really a property of the packet, and changes whenever the packet is modified. I like it! Up to you about making that change now.

opte/src/engine/packet.rs
bnaecker: Yeah, I know we want to avoid making copies, I was asking if there was a reason not to muck with the original body. Your point about it being dropped is a solid "no", so thanks!

opte/src/engine/packet.rs
bnaecker: Yeah, those are both good points, that we have to deal with limiting original offset and silently returning something "else" sounds bad. That does lead me to wonder though, why not make this return an optional type now? It looks like the only callsites that do _not_ currently have to handle errors are the `Packet::slice` and `Packet::slice_mut` functions, which are implemented in terms of `slice_unchecked`. For example, `Packet::body_segs` already returns an `Option`, so we could just propagate that if the slice fails. Same thing for `Packet::emit_headers`.

I'm not necessarily advocating we do that -- I know this PR has already dragged on much longer than you want to. But I know we've talked about not introducing further panics if possible. I guess one argument is that this really is a bug: if we can't take the slice we ask for, we've parsed the packet incorrectly. Anyway, just want to make sure we're on the same page about that tradeoff, adding a new possible panic site in exchange for not dealing with an optional value.

opte/src/engine/snat.rs
bnaecker: Sounds good. I think we've been generally going towards the idea of keeping IPv4 and IPv6 separate whenever possible, to avoid all the BS you've hit around checking for impossible conditions.

oxide-vpc/src/api.rs
rzezeski: Done.

oxide-vpc/tests/integration_tests.rs
rzezeski: Great point. I started making const values for these. While doing this I also moved some of the `<TYPE>_HDR_SZ` constants into their respective header types: e.g. `EtherHdr::SIZE`. This lead to more refactoring of other files but I do feel like it's more idiomatic rust code so just did it while I was here.

oxide-vpc/tests/integration_tests.rs
rzezeski: Done.

oxide-vpc/tests/integration_tests.rs
rzezeski: Done

oxide-vpc/tests/integration_tests.rs
rzezeski: Done

opte/src/engine/snat.rs
rzezeski: Done

opte/src/engine/snat.rs
rzezeski: Done

opte/src/engine/packet.rs
rzezeski: Done.

opte/src/engine/packet.rs
rzezeski: The problem is the other call sites of `slice_unchecked()` like `emit_headers()`, they are currently not prepared to deal with Result/Option from that call. And really they NEVER should be slicing outside the bounds of the segment, because they are slicing based on the `Parsed` data of the packet. So it just didn't seem like a can of worms to break open in this PR. Yes, we could add a new WriteError value and propagate up from new "unexpected" error or something (or maybe it could reuse OutOfRange)...anyways I just didn't want to have to think about all this in an already large PR.

And yea I wrote `slice`/`slice_mut` that way because they are literally just asking the segment for the entire buffer, using the pointers the segment itself deems as the bounds.

opte/src/engine/port.rs
rzezeski: Good news, in the kernel recursive mutex entries panic the box! So no worry about deadlocks. This is kind of a problem in general with all the Port methods, you have to know if it's expecting to grab the lock or already hold it. I achieved this my having some methods take a `&mut PortData`, which is only possible if you hold the lock. Of course, this doesn't stop you from calling another method on `self` that might also acquire said lock. This might be an argument for making "internal" methods be static so that `self` and `&mut PortData` cannot coexist. Or perhaps moving some methods on `PortData` instead. Until then, I'm just going to put this behind test cfg, because that's all it's used for currently. If some userland process needs these stats they can use kstats. I'll also make a note about the locking and write up an issue.

The good news is, practically speaking, we'll hit this in dev immediately by seeing a panic. So it's also worth not over rotating on this. The main way it could bite is if we have some API that is rarely used and has a recursive entry. Which gives me another thought: can we configure the std Mutex impl to panic on recursive entry? If so, we could also catch this when running tests. If it _doesn't_ support that feature, perhaps we can implement the `std` version of KMutex with our own direct bindings to the pthread mutex.

bnaecker: Sounds good, thanks for that.

> can we configure the std Mutex impl to panic on recursive entry?

Nope, I just ran into this last week. There is no support for configuring what happens in this case, and `std::sync::Mutex` is set up to just deadlock. Bummer. We may want to do make our own mutex over pthreads that's designed to panic here too, if we want parity between the userland and kernel implementations.

bnaecker (approved):
A couple of small things, but looks good to me otherwise!

oxide-vpc/src/engine/arp.rs
```diff
@@ -43,20 +43,36 @@ pub fn setup(
...
-                // Reuse the same MAC address for both IPs. This should be
-                // fine as the VIP is contained solely to the guest
-                // instance.
-                actions.push(Action::Hairpin(Arc::new(ArpReply::new(
-                    *ip,
-                    cfg.private_mac,
-                ))));
-            }
-        }
+        // if let Some(ip) = ip_cfg.external_ips.as_ref() {
```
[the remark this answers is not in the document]
rzezeski: Removed.

rzezeski:
Okay, after wandering through the desert for 40 days, one can finally ping the outside world via SNAT (no 1:1 NAT).

```
localhost:~# ip addr show eth0
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether a8:40:25:f7:95:36 brd ff:ff:ff:ff:ff:ff
    inet 172.30.0.5/32 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::aa40:25ff:fef7:9536/64 scope link 
       valid_lft forever preferred_lft forever

localhost:~# ping -c 3 10.0.0.1
PING 10.0.0.1 (10.0.0.1): 56 data bytes
64 bytes from 10.0.0.1: seq=0 ttl=64 time=4.229 ms
64 bytes from 10.0.0.1: seq=1 ttl=64 time=2.784 ms
64 bytes from 10.0.0.1: seq=2 ttl=64 time=2.797 ms

--- 10.0.0.1 ping statistics ---
3 packets transmitted, 3 packets received, 0% packet loss
round-trip min/avg/max = 2.784/3.270/4.229 ms
localhost:~# 
localhost:~# ping -c 3 1.1.1.1
PING 1.1.1.1 (1.1.1.1): 56 data bytes
64 bytes from 1.1.1.1: seq=0 ttl=57 time=20.585 ms
64 bytes from 1.1.1.1: seq=1 ttl=57 time=20.064 ms
64 bytes from 1.1.1.1: seq=2 ttl=57 time=19.220 ms

--- 1.1.1.1 ping statistics ---
3 packets transmitted, 3 packets received, 0% packet loss
round-trip min/avg/max = 19.220/19.956/20.585 ms
```


```
rpz@kalm:~/oxidecomputer/opte/xde$ pfexec opteadm list-ports                     
LINK                             MAC ADDRESS              IPv4 ADDRESS     EXTERNAL IPv4    IPv6 ADDRESS                             EXTERNAL IPv6              
              STATE   
opte5                            A8:40:25:F7:95:36        172.30.0.5       None             fd00:abcd:abcd:abcd::5                   None                       
              running

rpz@kalm:~/oxidecomputer/opte/xde$ pfexec opteadm dump-layer -p opte5 arp                                           
Layer arp                                                                                                                                                       
Inbound Flows          
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                 
                                                                                                                                                                
Outbound Flows                                                                                                                                                  
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                 
                                        
Inbound Rules                                                                   
ID       PRI    PREDICATES                                       ACTION             
0        1      inner.ether.ether_type=0x806 inner.ether.dst=FF:FF:FF:FF:FF:FF inner.arp.htype=1 inner.arp.ptype=2048 inner.arp.op=Request inner.arp.data.tpa=10
.0.0.226 "HAIRPIN: ArpReply 10.0.0.226 => A8:40:25:F7:95:36"          
1        2      inner.ether.ether_type=0x806                     "DENY"                                                                                         
                                                                                                                                                                
Outbound Rules                                                                                                                                                  
ID       PRI    PREDICATES                                       ACTION             
0        1      inner.ether.ether_type=0x806 inner.ether.dst=FF:FF:FF:FF:FF:FF inner.arp.htype=1 inner.arp.ptype=2048 inner.arp.op=Request inner.ether.src=A8:40
:25:F7:95:36 inner.arp.data.tpa=172.30.0.1 "HAIRPIN: ArpReply 172.30.0.1 => A8:40:25:FF:77:77"
1        2      inner.ether.ether_type=0x806                     "DENY"  

rpz@kalm:~/oxidecomputer/opte/xde$ pfexec opteadm dump-uft -p opte5              
Unified Flow Table
Inbound Flows [3/8096]
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                 
ICMP   1.1.1.1          0      10.0.0.226       0      2        hdr: SNAT, body: ICMP Echo Ident/SNAT 1834 <=> 16382
ICMP   10.0.0.1         0      10.0.0.226       0      2        hdr: SNAT, body: ICMP Echo Ident/SNAT 1833 <=> 16383
UDP    10.0.0.39        57621  10.0.0.255       57621  13       hdr: , body:           

Outbound Flows [2/8096]
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                 
ICMP   172.30.0.5       0      1.1.1.1          0      2        hdr: SNAT, body: ICMP Echo Ident/SNAT 1834 <=> 16382
ICMP   172.30.0.5       0      10.0.0.1         0      2        hdr: SNAT, body: ICMP Echo Ident/SNAT 1833 <=> 16383

rpz@kalm:~/oxidecomputer/opte/xde$ 
(reverse-i-search)`': ^C
rpz@kalm:~/oxidecomputer/opte/xde$ pfexec opteadm dump-layer -p opte5 nat
Layer nat
Inbound Flows
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                 
ICMP   1.1.1.1          0      10.0.0.226       0      1        SNAT_ICMP_ECHO         
ICMP   10.0.0.1         0      10.0.0.226       0      1        SNAT_ICMP_ECHO         

Outbound Flows
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                 
ICMP   172.30.0.5       0      1.1.1.1          0      0        SNAT_ICMP_ECHO         
ICMP   172.30.0.5       0      10.0.0.1         0      0        SNAT_ICMP_ECHO         

Inbound Rules
ID       PRI    PREDICATES                                       ACTION             

Outbound Rules
ID       PRI    PREDICATES                                       ACTION             
1        100    inner.ether.ether_type=0x86DD meta: router-target=ig  "STATEFUL: [fd00:abcd:abcd:abcd::2]:0-16384"
0        100    inner.ether.ether_type=0x800 meta: router-target=ig  "STATEFUL: 10.0.0.226:0-16383"
```

Issue #244, opened 2022-08-30
Want support for IPv6 Neighbor Discovery Protocol


bnaecker:
OPTE needs to be set up to respond to Router Solicitations (on both the all-routers broadcast and its unicast address). The Router Advertisements it shoots back need to instruct the host that OPTE is the default (well, only) router, and that the host needs to use DHCPv6 to get both its VPC-private IPv6 address and the DNS servers.

There shouldn't be much else, OPTE doesn't need to provide any on-link prefixes. OPTE may need to make sure the host's SLAAC procedure doesn't somehow pick the same IP as OPTE itself, but that's very unlikely since the host _should_ be using an EUI-64 transform on its (Nexus-provided) MAC address to derive it's link-local IPv6 address. Since OPTE sets itself up as the only router, and there's nothing else on-link, there shouldn't be any address resolution needed.

bnaecker:
Addressed in #260.

Issue #42, opened 2022-02-10
Want anti-spoof enforcement


rzezeski:
Add anti-spoof protection to OPTE to make sure a guest can only send out frames/packets that match its prescribed MAC address + IP address for the given interface.

Along with this there should be a dedicated spoof stat(s). Also might want a dedicated SDT probe for this case as well.

rzezeski:
This anti-spoof enforcement was addressed in 717aaf1283a84fe5797fb733a6448fac04dfe3a3. However, before marking this done we should

- [ ] Add a regression test.
- [ ] Add spoof stats.
- [ ] Add spoof probe.

rzezeski:
It turns out that having layer-specific stats like a "spoof stat" is harder than I thought.

First, I see no easy way to attach a new kstats structure to each layer. If we make it a generic on `Layer`, then that breaks the `Vec<Layer>` in `Port`. If we change the layers list to be `Vec<dyn Layer>`, then that means each layer impl could propose its own process method, which is not at all how opte was designed to work (the processing is generic, the customization comes in the form of rules and their associated actions). We could probably do something with raw pointers and casting, but that feels wrong. I just don't see the easy/right way to do this.

Second, the more I think about it the more I'm not sure such stats need to exist anyways. To implement a "spoof stat", the gateway layer impl somehow needs to know when no rule matched and hook into that. Then it needs to inspect the packet/flow to know if it didn't match any rules because it is spoofed or for some other reason. This seems a bit odd of a way to go about things, and perhaps a bit heavyweight in terms of code and runtime efficiency.

Instead, I think #207 gets us most of the benefit without having to do a bunch of remodeling. It tells us which rules are matching, and when no rules are matching at all (assuming we also add "default rule match" stats). With those stats in place, we can then use the more generic flow/packet-based probes to delve into the traffic to determine why it isn't matching. We wouldn't have historical counts of "this guest tried to spoof a packet X times", but I think that's a fine tradeoff for now given the complexity vibes I'm getting from trying to implement these stats.

With that said I'm going to pass on making specific "spoof stats" and a "spoof probe". I still want a test, but I'll make that a new issue to stand on its own.

rzezeski:
It turns out that having layer-specific stats like a "spoof stat" is harder than I thought.

[...]

PR #260, opened 2022-09-22
Implements core NDP support

bnaecker:
- Adds types for handling expected NDP messages, including Router and Neigbor Solicitations and Advertisements
- Reworks the ICMPv6 layer for `oxide-vpc` to include better Echo support, and adds new NDP handling.

### DIFF
--- a/opte-api/src/ip.rs
+++ b/opte-api/src/ip.rs
@@ -21,6 +21,9 @@ cfg_if! {
 }
 
 /// Generate an ICMPv6 Echo Reply message.
+///
+/// This maps an ICMPv6 Echo Request message from `src` to `dst` into an ICMPv6
+/// Echo Reply message from `dst` to `src`.
 #[derive(Debug, Clone, Copy)]
 pub struct Icmpv6EchoReply {
     /// The MAC address of the Echo Request source.
@@ -32,7 +35,7 @@ pub struct Icmpv6EchoReply {
     /// The MAC address of the Echo Request destination.
     pub dst_mac: MacAddr,
 
-    /// The IP address of the Echo source destination.
+    /// The IP address of the Echo Request destination.
     pub dst_ip: Ipv6Addr,
 }
 
@@ -481,7 +484,79 @@ pub struct Ipv6Addr {
 }
 
 impl Ipv6Addr {
+    /// The unspecified IPv6 address, i.e., `::` or all zeros.
     pub const ANY_ADDR: Self = Self { inner: [0; 16] };
+
+    /// The All-Routers multicast address, used in the Neighbor Discovery
+    /// Protocol.
+    pub const ALL_ROUTERS: Self =
+        Self::from_const([0xff02, 0, 0, 0, 0, 0, 0, 2]);
+
+    /// The All-Nodes multicast address, used in the Neighbor Discovery
+    /// Protocol.
+    pub const ALL_NODES: Self = Self::from_const([0xff02, 0, 0, 0, 0, 0, 0, 1]);
+
+    /// Generate an IPv6 address via an EUI-64 transform, from a MAC address.
+    /// The generated address has link-local scope.
+    ///
+    /// See https://www.rfc-editor.org/rfc/rfc4291#page-20 for details of the
+    /// transformation applied.
+    pub fn from_eui64(mac: &MacAddr) -> Self {
+        let mac = mac.bytes();
+        // Invert the locally-administered bit in the first octet of the MAC
+        let mac0 = mac[0] ^ 0b10;
+        let bytes: [u8; 16] = [
+            0xfe, 0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, mac0, mac[1],
+            mac[2], 0xff, 0xfe, mac[3], mac[4], mac[5],
+        ];
+        Self::from(bytes)
+    }
+
+    /// Return the multicast MAC address associated with this multicast IPv6
+    /// address. If the IPv6 address is not multicast, None will be returned.
+    ///
+    /// See https://www.rfc-editor.org/rfc/rfc2464, section 7 for details.
+    pub const fn multicast_mac(&self) -> Option<MacAddr> {
+        if self.is_multicast() {
+            Some(self.unchecked_multicast_mac())
+        } else {
+            None
+        }
+    }
+
+    /// Return the multicast MAC address associated with this multicast IPv6
+    /// address, without checking if this IP address is a multicast address.
+    ///
+    /// See https://www.rfc-editor.org/rfc/rfc2464, section 7 for details.
+    pub const fn unchecked_multicast_mac(&self) -> MacAddr {
+        let bytes = &self.inner;
+        MacAddr::from_const([
+            0x33, 0x33, bytes[12], bytes[13], bytes[14], bytes[15],
+        ])
+    }
+
+    /// Return the solicited-node multicast IPv6 address corresponding to
+    /// `self`.
+    ///
+    /// See https://www.rfc-editor.org/rfc/rfc4291#section-2.7.1 for details.
+    pub const fn solicited_node_multicast(&self) -> Ipv6Addr {
+        let bytes = &self.inner;
+        let w0 = u16::from_be_bytes([0xff, bytes[13]]);
+        let w1 = u16::from_be_bytes([bytes[14], bytes[15]]);
+        Self::from_const([0xff02, 0, 0, 0, 0, 1, w0, w1])
+    }
+
+    /// Return `true` if this is a solicited node multicast address.
+    pub fn is_solicited_node_multicast(&self) -> bool {
+        const EXPECTED: &[u8] =
+            &[0xff, 0x02, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x01, 0xff];
+        &self.inner[..EXPECTED.len()] == EXPECTED
+    }
+
+    /// Return `true` if this is a multicast IPv6 address, and `false` otherwise
+    pub const fn is_multicast(&self) -> bool {
+        self.inner[0] == 0xFF
+    }
 
     /// Return the bytes of the address.
     pub fn bytes(&self) -> [u8; 16] {
@@ -524,6 +599,23 @@ impl Ipv6Addr {
     pub fn safe_mask(self, mask: Ipv6PrefixLen) -> Self {
         self.mask(mask.val()).unwrap()
     }
+
+    pub const fn from_const(words: [u16; 8]) -> Self {
+        let w0 = words[0].to_be_bytes();
+        let w1 = words[1].to_be_bytes();
+        let w2 = words[2].to_be_bytes();
+        let w3 = words[3].to_be_bytes();
+        let w4 = words[4].to_be_bytes();
+        let w5 = words[5].to_be_bytes();
+        let w6 = words[6].to_be_bytes();
+        let w7 = words[7].to_be_bytes();
+        Self {
+            inner: [
+                w0[0], w0[1], w1[0], w1[1], w2[0], w2[1], w3[0], w3[1], w4[0],
+                w4[1], w5[0], w5[1], w6[0], w6[1], w7[0], w7[1],
+            ],
+        }
+    }
 }
 
 impl fmt::Display for Ipv6Addr {
@@ -546,6 +638,13 @@ impl From<smoltcp::wire::Ipv6Address> for Ipv6Addr {
         // octets in the correct order.
         let bytes: [u8; 16] = ip.as_bytes().try_into().unwrap();
         Self::from(bytes)
+    }
+}
+
+impl From<Ipv6Addr> for smoltcp::wire::Ipv6Address {
+    fn from(ip: Ipv6Addr) -> Self {
+        // Safety: This panics, but we know bytes is exactly 16 octets.
+        Self::from_bytes(&ip.bytes())
     }
 }
 
@@ -835,6 +934,12 @@ impl Ipv6PrefixLen {
 }
 
 impl Ipv6Cidr {
+    /// The IPv6 link-local prefix, `fe80::/64`.
+    pub const LINK_LOCAL: Self = Self {
+        ip: Ipv6Addr::from_const([0xfe80, 0, 0, 0, 0, 0, 0, 0]),
+        prefix_len: Ipv6PrefixLen(64),
+    };
+
     pub fn new(ip: Ipv6Addr, prefix_len: Ipv6PrefixLen) -> Self {
         let ip = ip.safe_mask(prefix_len);
         Ipv6Cidr { ip, prefix_len }
@@ -1074,5 +1179,50 @@ mod test {
             ),
             "fd00::1/64".parse().unwrap(),
         );
+    }
+
+    #[test]
+    fn test_ipv6_from_eui64() {
+        let mac = MacAddr::from_const([0x01, 0x02, 0x03, 0x04, 0x05, 0x06]);
+        let actual = Ipv6Addr::from_eui64(&mac);
+        // The locally-administered bit should be set, but otherwise the MAC is
+        // just transcribed.
+        assert_eq!(actual, "fe80::0302:03ff:fe04:0506".parse().unwrap());
+    }
+
+    #[test]
+    fn test_ipv6_from_const() {
+        assert_eq!(
+            Ipv6Addr {
+                inner: [0xfe, 0x80, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+            },
+            Ipv6Addr::from_const([0xfe80, 0, 0, 0, 0, 0, 0, 0]),
+        );
+    }
+
+    fn to_ipv6(s: &str) -> Ipv6Addr {
+        s.parse().unwrap()
+    }
+
+    #[test]
+    fn test_ipv6_is_multicast() {
+        assert!(!to_ipv6("fd00::1").is_multicast());
+        assert!(to_ipv6("ff00::1").is_multicast());
+    }
+
+    #[test]
+    fn test_ipv6_multicast_mac() {
+        assert!(to_ipv6("fd00::1").multicast_mac().is_none());
+        assert_eq!(
+            to_ipv6("ff00::0001:0203").multicast_mac().unwrap(),
+            MacAddr::from([0x33, 0x33, 0, 1, 2, 3]),
+        );
+    }
+
+    #[test]
+    fn test_ipv6_solicited_node_multicast() {
+        let addr = to_ipv6("fd00:abcd:abcd:abcd:abcd:abcd:abcd:abcd");
+        let expected = to_ipv6("ff02::1:ffcd:abcd");
+        assert_eq!(addr.solicited_node_multicast(), expected);
     }
 }
--- a/opte-api/src/ndp.rs
+++ b/opte-api/src/ndp.rs
@@ -1,0 +1,125 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Copyright 2022 Oxide Computer Company
+
+//! Types for working with the IPv6 Neighbor Discovery Protocol
+
+use super::Ipv6Addr;
+use super::MacAddr;
+use core::fmt::{self, Debug, Display};
+
+/// A Neighbor Discovery Protocol Router Advertisement, generated in response to
+/// a Router Solicitation.
+#[derive(Clone, Copy, Debug)]
+pub struct RouterAdvertisement {
+    /// The expected MAC address of the client whose Router Solicitations we
+    /// respond to.
+    pub src_mac: MacAddr,
+
+    /// The MAC address advertised by the router.
+    pub mac: MacAddr,
+
+    // The IPv6 address advertised by the router, which is the EUI-64 transform
+    // of the router MAC address.
+    ip: Ipv6Addr,
+
+    /// Managed address configuration, indicating that the peer can use DHCPv6
+    /// to acquire an IPv6 address.
+    pub managed_cfg: bool,
+}
+
+impl RouterAdvertisement {
+    /// Create new `RouterAdvertisement`.
+    ///
+    /// The `src_mac` is the expected source MAC address a Router Solicitation
+    /// should come from. There are no restrictions on the source IP address,
+    /// other than that it be link-local, in `fe80::/10`.
+    ///
+    /// `mac` is the MAC address of the router, to which Solicitations are
+    /// expected to be addressed, and from which Advertisements are sent. The
+    /// source IPv6 address of the Advertisement is derived from this, using the
+    /// EUI-64 transform.
+    ///
+    /// `managed_cfg` is set to `true` to indicate that the host can get further
+    /// configuration from a DHCPv6 server running on the network.
+    pub fn new(src_mac: MacAddr, mac: MacAddr, managed_cfg: bool) -> Self {
+        let ip = Ipv6Addr::from_eui64(&mac);
+        Self { src_mac, mac, ip, managed_cfg }
+    }
+
+    /// Return the IPv6 address the router sends advertisements from.
+    pub fn ip(&self) -> &Ipv6Addr {
+        &self.ip
+    }
+}
+
+impl Display for RouterAdvertisement {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "NDP RA IPv6={} MAC={}", self.ip, self.mac)
+    }
+}
+
+/// A Neighbor Discovery Protocol Neighbor Advertisement, generated in response to
+/// a Neighbor Solicitation.
+#[derive(Clone, Copy, Debug)]
+pub struct NeighborAdvertisement {
+    /// The expected MAC address of the client whose Neighbor Solicitations we
+    /// respond to.
+    pub src_mac: MacAddr,
+
+    /// The MAC address advertised by the neighbor.
+    pub mac: MacAddr,
+
+    // The advertised IPv6 address of the neighbor, which is the EUI-64
+    // transform of the source MAC address, in `mac`.
+    ip: Ipv6Addr,
+
+    /// If true, advertise that this neighbor is a router.
+    pub is_router: bool,
+
+    /// If true, respond to Neighbor Solicitations sent from the unspecified
+    /// address `::`, in addition to those from a link-local address
+    /// `fe80::/10`.
+    pub allow_unspec: bool,
+}
+
+impl NeighborAdvertisement {
+    /// Create new `NeighborAdvertisement`.
+    ///
+    /// The `src_mac` is the expected source MAC address a Neighbor Solicitation
+    /// should come from. There are no restrictions on the source IP address,
+    /// other than that it be link-local, in `fe80::/10`.
+    ///
+    /// `mac` is the MAC address of the neighbor, to which Solicitations are
+    /// expected to be addressed, and from which Advertisements are sent. The
+    /// source IPv6 address of the Advertisement is derived from this, using the
+    /// EUI-64 transform.
+    ///
+    /// `is_router` is `true` if the advert should be marked as coming from a
+    /// router.
+    ///
+    /// `allow_unspec` is `true` if the advertisement is generated in response
+    /// to Neigbor Solicitations from the unspecified address.
+    pub fn new(
+        src_mac: MacAddr,
+        mac: MacAddr,
+        is_router: bool,
+        allow_unspec: bool,
+    ) -> Self {
+        let ip = Ipv6Addr::from_eui64(&mac);
+        Self { src_mac, mac, ip, is_router, allow_unspec }
+    }
+
+    /// Return the IPv6 address the neighbor sends advertisements from.
+    pub fn ip(&self) -> &Ipv6Addr {
+        &self.ip
+    }
+}
+
+impl Display for NeighborAdvertisement {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "NDP NA IPv6={} MAC={}", self.ip, self.mac)
+    }
+}
--- a/opte/src/engine/icmpv6.rs
+++ b/opte/src/engine/icmpv6.rs
@@ -15,18 +15,26 @@ use super::rule::{
     HairpinAction, IpProtoMatch, Ipv6AddrMatch, Predicate,
 };
 use core::fmt::{self, Display};
-pub use opte_api::ip::{Icmpv6EchoReply, Protocol};
+pub use opte_api::ip::Ipv6Cidr;
+pub use opte_api::ip::{Icmpv6EchoReply, Ipv6Addr, Protocol};
+use opte_api::mac::MacAddr;
+pub use opte_api::ndp::NeighborAdvertisement;
+pub use opte_api::ndp::RouterAdvertisement;
 use serde::{Deserialize, Serialize};
 use smoltcp::phy::{Checksum, ChecksumCapabilities as Csum};
+use smoltcp::wire::NdiscNeighborFlags;
+use smoltcp::wire::RawHardwareAddress;
 use smoltcp::wire::{
-    Icmpv6Message, Icmpv6Packet, Icmpv6Repr, IpAddress, Ipv6Address,
+    Icmpv6Message, Icmpv6Packet, Icmpv6Repr, IpAddress, Ipv6Address, NdiscRepr,
 };
 
 cfg_if! {
     if #[cfg(all(not(feature = "std"), not(test)))] {
         use alloc::vec::Vec;
+        use alloc::string::String;
     } else {
         use std::vec::Vec;
+        use std::string::String;
     }
 }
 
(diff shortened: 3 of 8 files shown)


bnaecker:
This adds the basic NDP support, which is mostly responding to Router Solicitations with Router Adverts, and Neighbor Solicitations with Neighbor Adverts. Using a Linux (Ubuntu) guest, we can see that the guest correctly sends an RS, gets back the RA from OPTE, and uses that to update its neighbor cache and list of default routers.

This shows the first NS, which the guest uses for DAD to verify it can use its chosen link-local address. Then we see the RS and RA reply from OPTE.

```
bnaecker@feldspar : ~/opte/xde $ snoop -i boot.snoop
  1   0.00000  UNSPECIFIED -> ff02::1:fffb:e191 ICMPv6 Neighbor solicitation
  2   1.02410 fe80::aa40:25ff:fefb:e191 -> ip6-allrouters ICMPv6 Router solicitation
  3   0.00003 fe80::aa40:25ff:feff:7777 -> fe80::aa40:25ff:fefb:e191 ICMPv6 Router advertisement
  4 132.81956 fe80::aa40:25ff:fefb:e191 -> fe80::aa40:25ff:feff:7777 ICMPv6 Echo request (ID: 25619 Sequence number: 1)
  5   0.00004 fe80::aa40:25ff:feff:7777 -> fe80::aa40:25ff:fefb:e191 ICMPv6 Echo reply (ID: 25619 Sequence number: 1)
  6   5.22836 fe80::aa40:25ff:fefb:e191 -> fe80::aa40:25ff:feff:7777 ICMPv6 Neighbor solicitation
  7   0.00009 fe80::aa40:25ff:feff:7777 -> fe80::aa40:25ff:fefb:e191 ICMPv6 Neighbor advertisement
  8  37.54866 fe80::aa40:25ff:fefb:e191 -> fd00:abcd:abcd:abcd::5 ICMPv6 Echo request (ID: 18020 Sequence number: 1)
  9   5.20343 fe80::aa40:25ff:fefb:e191 -> fe80::aa40:25ff:feff:7777 ICMPv6 Neighbor solicitation
 10   0.00007 fe80::aa40:25ff:feff:7777 -> fe80::aa40:25ff:fefb:e191 ICMPv6 Neighbor advertisement
 11  55.39297  UNSPECIFIED -> ff02::1:ff00:5 ICMPv6 Neighbor solicitation
 12  75.14530 fd00:abcd:abcd:abcd::5 -> fd00:abcd:abcd:abcd::1 ICMPv6 Echo request (ID: 26831 Sequence number: 1)
 13   0.00006 fd00:abcd:abcd:abcd::1 -> fd00:abcd:abcd:abcd::5 ICMPv6 Echo reply (ID: 26831 Sequence number: 1)
 14   5.14363 fe80::aa40:25ff:fefb:e191 -> fe80::aa40:25ff:feff:7777 ICMPv6 Neighbor solicitation
 15   0.00008 fe80::aa40:25ff:feff:7777 -> fe80::aa40:25ff:fefb:e191 ICMPv6 Neighbor advertisement
```

That's generated at the startup of the guest. The later packets are after I've SSH'd in and am trying to ping various things. I manually assigned the guest the fake VPC-private IPv6 address `fd00:abcd:abcd:abcd::5`. In the middle, you can see the DAD running in packet 11, where the guest is checking if any neighbor has that address already. Once that's done, I'm able to ping the OPTE virtual gateway, both at its link-local address and the VPC-private `fd00:abcd:abcd:abcd::1`. It's using the link-local address of the OPTE router, which we can see from the guest itself:

```bash
debian@myinst:~$ ip -6 neigh
fe80::aa40:25ff:feff:7777 dev enp0s8 lladdr a8:40:25:ff:77:77 router STALE
debian@myinst:~$ ip -6 route
::1 dev lo proto kernel metric 256 pref medium
fd00:abcd:abcd:abcd::5 dev enp0s8 proto kernel metric 256 pref medium
fe80::/64 dev enp0s8 proto kernel metric 256 pref medium
default via fe80::aa40:25ff:feff:7777 dev enp0s8 proto ra metric 1024 expires 8170sec hoplimit 255 pref medium
debian@myinst:~$
```

There's a bit of an interesting note here. RFD 21 describes that virtual gateway address, which makes a lot of sense coming from IPv4 -- the guest would use DHCP to get that as the gateway. But IPv6 works a bit differently. The list of usable gateways is populated from NDP. Furthermore, the fact that we give OPTE that address is a funky, because that virtual address is "part of" OPTE, but it's not a neighbor. In particular, we don't want to respond to NSs for that address, because the guest might use that to determine which addresses are on-link. I'm not 100% this is the right behavior, but I'm trying to be conservative and ensure that the guest sends all traffic through its only default router, OPTE, via its link-local address.

I've also updated the Echo support. Previously OPTE only responded to pings at that virtual gateway address. Now it responds to pings at the virtual gateway or its link-local address, and from either the link-local or VPC-private IP address of the guest (4 possible situations).

Also note that both the NDP support and the ping support assume the guest creates their link-local IPv6 addresses at startup time through the EUI-64 transform, applied to their MAC address.

bnaecker:
This should resolve #244

bnaecker:
Just tested this with an Ubuntu guest, and we get the same behavior.

opte-api/src/ndp.rs
```diff
@@ -0,0 +1,134 @@
...
+    ///
+    /// `mac` is the destination MAC the Solitications are expected to be
+    /// delivered to, and the source MAC of the generated Advertisement. The
+    /// IPv6 address is also derived from this.
+    ///
+    /// `managed_cfg` is set to `true` to indicate that the host can get further
+    /// configuration from a DHCPv6 server running on the network.
+    pub fn new(src_mac: MacAddr, mac: MacAddr, managed_cfg: bool) -> Self {
+        let ip = Ipv6Addr::from_eui64(&mac);
+        let src_ip = Ipv6Addr::from_eui64(&src_mac);
```
bnaecker, line 52: I've been thinking more about this. It seems like we should not expect that guests use the EUI-64 transform to generate their link-local addresses. @rcgoodfellow pointed out that some versions of Fedora and RHEL may be configured to do something else. It seems sufficient for us to respond to any inbound RS from the guest's MAC, with any link-local address in `fe80::/10`.

rcgoodfellow:
> There's a bit of an interesting note here. RFD 21 describes that virtual gateway address, which makes a lot of sense coming from IPv4 -- the guest would use DHCP to get that as the gateway. But IPv6 works a bit differently. The list of usable gateways is populated from NDP. Furthermore, the fact that we give OPTE that address is a funky, because that virtual address is "part of" OPTE, but it's not a neighbor. In particular, we don't want to respond to NSs for that address, because the guest might use that to determine which addresses are on-link. I'm not 100% this is the right behavior, but I'm trying to be conservative and ensure that the guest sends all traffic through its only default router, OPTE, via its link-local address.

I believe what you have done is the desired/expected way to go about this. I don't see a reason to have anything beyond a link-local address as the v6 gateway. We should update RFD 21 to reflect this. In v4, a gateway IP address is a means to get a MAC address via ARP in order to send off-network packets to that gateway at layer 2. In v6, we have NDP which operates over link-local addresses for this purpose.

opte-api/src/ndp.rs
rcgoodfellow: RFC 4291 IP Version 6 Addressing Architecture, which originally specified that interface identifiers be constructed in a Modified EUI-64 format, was updated by RFC 7136 Significance of IPv6 Interface Identifiers. That RFC modifies the statements of RFC 4291 to specify that Modified EUI-64 construction is required only when an interface identifier is derived from a MAC address.

While the motivations for using an interface identifier other than one derived from a MAC address are completely vacuous for link-local addresses (the MAC address from which the interface identifier is derived is by definition in the packet already) - no specific provisions were made in the RFC 7136 for link-local addresses. Because of that, there can be no realistic expectation of any sort of deterministic structure for interface identifiers within link-local addresses, including the Modified EUI-64 that was originally specified.

bnaecker:
Thanks for that @rcgoodfellow. I'm a bit confused, though, by the "what you have done" part. I'm currently faking up the first address in the VPC subnet as something that can be pinged basically. That address doesn't "do" anything else. It sounds like you're suggesting we _remove_ that, and update RFD 21 to match, and _only_ use the EUI-64 link-local as the gateway for guests populated by NDP. Is that right?

opte-api/src/ndp.rs
bnaecker: Thanks Ry. The summary then is we need to just use `fe80::/10`, and assume nothing else about the source IP address, yeah?

rcgoodfellow:
Ah, got it. Yes, I'm suggesting we remove that address, update RFD 21 to match, and only use a link-local address as the gateway.

bnaecker:
Sounds good, thanks. I'll remove it now, it's very simple to add it back if we want to keep it. CC @rmustacc, I'm curious if you had other motivations for supporting the first address in the network as the IPv6 gateway, before I go on to modify RFD 21.

opte-api/src/ndp.rs
rcgoodfellow: Yup. I'm afraid so.

bnaecker: Thanks for confirming. It's no big deal, @rzezeski has made it trivial to change that kinda thing in OPTE. I'll update this and re-push.

opte/src/engine/icmpv6.rs
```diff
@@ -172,3 +176,296 @@ impl HairpinAction for Icmpv6EchoReply {
...
+        Ok(AllowOrDeny::Allow(Packet::copy(&pkt_bytes)))
+    }
+}
+
+impl HairpinAction for NeighborAdvertisement {
+    fn implicit_preds(&self) -> (Vec<Predicate>, Vec<DataPredicate>) {
+        let hdr_preds = vec![
+            // We expect that the source MAC is the MAC provided to the Guest.
+            Predicate::InnerEtherSrc(vec![EtherAddrMatch::Exact(self.src_mac)]),
+            // It's directed to the virtual gateway's MAC address.
```
rzezeski, line 351: I thought NS messages were directed to the all-nodes multicast address?

rzezeski: Okay, so reading more I see there is unicast and multicast solicitation. So this kind of comes back to my various comments about the general vs VPC-specific wording in the comments. Now I think we might also be starting to creep into VPC-specific implementation. That said, I don't want to cause a stir in this PR and stop momentum. I just want to share that my intention (which is probably not clear outside of my own mind) is that opte-api holds actions that are mostly generic, and anything that starts getting fairly specific to oxide-vpc lives in there. I realize that you are just following the pattern I've laid out and that this action is borderline general vs specific. You could also argue that we only have the one implementation at the moment, so maybe this is all fine for now. I just wanted to point out this division that I try to keep in my head and that I've been trying to enforce more and more in the code.

oxide-vpc/src/engine/icmpv6.rs
```diff
@@ -9,41 +9,169 @@
...
+// - Respond to echo requests from the guest to the gateway. The source address
+// may be either any link-local address in guest (since we can't know how they
+// generate that) or its assigned VPC-private address. The destination address
+// must be the link-local address we derive for OPTE, from the EUI-64 transform
+// on its MAC address.
+//
+// - Respond to NDP Router Solicitations from the guest to the gateway.
+//
+// - Respond to NDP Neighbor Solicitations from the guest to the gateway, with a
+// Target Address of both the link-local and VPC-private IPv6 address.
```
rzezeski, line 83: This comment has me confused. We are responding to a unicast solicitation for the Target Address (which I think is always going to be the router's link-local address if I'm reading things correctly) and we respond with the corresponding target link-layer (MAC) address.

oxide-vpc/src/engine/icmpv6.rs
```diff
@@ -9,41 +9,169 @@
...
+    let mut icmp = Layer::new("icmpv6", pb.name(), actions, ft_limit);
+
+    // Add rules for the above actions.
+    for i in 0..n_actions {
+        let priority = u16::try_from(i + 1).unwrap();
+        let rule = Rule::new(priority, icmp.action(i).unwrap().clone());
+        icmp.add_rule(Direction::Out, rule.finalize());
+    }
+
+    // Drop any other NDP traffic in either direction, both in and out.
```
rzezeski, line 147: So I think I'd approach this a different way.

First off, at the max priority, place a deny for **any** ICMPv6 traffic.

At `n_actions + 1` place an allow action (I think the static `Identity` action would do) for any ICMPv6 Echo/Reply. As we need other ICMPv6 traffic (like perhaps inbound DU) we can add it to the list of allowed traffic.

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2024,3 +2046,369 @@ fn gateway_icmpv6_ping() {
...
+// Advertisement from the gateway.
+#[test]
+fn gateway_router_advert_reply() {
+    use smoltcp::time::Duration;
+    use smoltcp::wire::Icmpv6Packet;
+    use smoltcp::wire::Icmpv6Repr;
+    use smoltcp::wire::Ipv6Address;
+    use smoltcp::wire::NdiscRepr;
+    use smoltcp::wire::NdiscRouterFlags;
+    use smoltcp::wire::RawHardwareAddress;
```
rzezeski, line 2114: I know I've done this myself in some places, but we might as well move these imports to module level.

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2024,3 +2046,369 @@ fn gateway_icmpv6_ping() {
         _ => panic!("expected Echo Reply, got {:?}", reply_icmp),
     }
 }
+
+// Generate a packet containing an NDP Router Solicitation.
+//
+// The source MAC is used to generate the source IPv6 address, using the EUI-64
+// transform. The resulting packet is has a multicast MAC address, and the
```
rzezeski, line 2053:
```suggestion
// transform. The resulting packet has a multicast MAC address, and the
```

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2024,3 +2046,369 @@ fn gateway_icmpv6_ping() {
...
+// All-Routers destination IPv6 address.
+fn gen_router_solicitation(src_mac: &MacAddr) -> Packet<Parsed> {
+    use smoltcp::wire::Icmpv6Packet;
+    use smoltcp::wire::Icmpv6Repr;
+    use smoltcp::wire::IpAddress;
+    use smoltcp::wire::NdiscRepr;
+    use smoltcp::wire::RawHardwareAddress;
+
+    // The source IPv6 address is the EUI-64 transform of the source MAC
```
rzezeski, line 2062:
```suggestion
    // The source IPv6 address is the EUI-64 transform of the source MAC.
```

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2024,3 +2046,369 @@ fn gateway_icmpv6_ping() {
...
+    let src_ip = Ipv6Addr::from_eui64(src_mac);
+
+    // Must be destined for the All-Routers IPv6 address, and the corresponding
+    // multicast Ethernet address
```
rzezeski, line 2066:
```suggestion
    // multicast Ethernet address.
```

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2024,3 +2046,369 @@ fn gateway_icmpv6_ping() {
...
+    let dst_ip: Ipv6Addr = "ff02::2".parse().unwrap();
```
rzezeski, line 2067: If you want might be nice to add pub consts to Ipv6Addr for these addresses.

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2024,3 +2046,369 @@ fn gateway_icmpv6_ping() {
...
+                Ipv6Address::from_bytes(ip6.src.bytes().as_slice()),
+                Ipv6Address::from_bytes(expected_dst.bytes().as_slice()),
+            )
+        }
+        ip4 => panic!("expected inner IPv6 metadata, got IPv4: {:?}", ip4),
+    };
+
+    let mut rdr = PacketReader::new(&reply, ());
+    // Need to seek to body.
+    rdr.seek(ETHER_HDR_SZ + IPV6_HDR_SZ).unwrap();
```
rzezeski, line 2196: You can replace this with `reply.get_body_rdr()`.

rzezeski (commented):
Bunch of great work. Nice to see IPv6 coming together. I think the bulk of my comments are around generic vs VPC-specific comments/code in opte-api. This might be worth a live discussion if what I'm saying isn't clear. It's also not a huge deal to get all figured out in this PR, as we can just rewrite comments/move code in future commits. I mainly want to make sure I get these ideas out in the public so it's better understood why I've tried to structure opte the way I did.

opte/src/engine/icmpv6.rs
```diff
@@ -172,3 +176,296 @@ impl HairpinAction for Icmpv6EchoReply {
         Ok(AllowOrDeny::Allow(Packet::copy(&pkt_bytes)))
     }
 }
+
+impl HairpinAction for RouterAdvertisement {
+    fn implicit_preds(&self) -> (Vec<Predicate>, Vec<DataPredicate>) {
+        const ALL_ROUTERS_ADDR: Ipv6Addr = Ipv6Addr::from_const([
+            0xff, 0x02, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2,
+        ]);
+        const ALL_ROUTERS_MAC: MacAddr =
```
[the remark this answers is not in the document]
bnaecker: Yeah, I'm adding a method like that now as part of DHCPv6 support. I'll go ahead and move it to this PR.

opte-api/src/ndp.rs
```diff
@@ -0,0 +1,114 @@
...
+
+use super::Ipv6Addr;
+use super::MacAddr;
+use core::fmt::{self, Debug, Display};
+
+/// A Neighbor Discovery Protocol Router Advertisement, generated in response to
+/// a Router Solicitation.
+#[derive(Clone, Copy, Debug)]
+pub struct RouterAdvertisement {
+    /// The expected MAC address of the guest whose Router Solicitations we
```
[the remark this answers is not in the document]
bnaecker: Yeah, I struggled with how specific to make this about actual VM instances too. I like "client," gets the point across nicely. I'll update here.

opte/src/engine/icmpv6.rs
```diff
@@ -172,3 +176,296 @@ impl HairpinAction for Icmpv6EchoReply {
...
+        let flags = if self.managed_cfg {
+            NdiscRouterFlags::MANAGED
+        } else {
+            NdiscRouterFlags::empty()
+        };
+        let advert = NdiscRepr::RouterAdvert {
+            hop_limit: u8::MAX,
+            flags,
+            // Use the maximum advertised lifetime as a default router.
+            router_lifetime: Duration::from_secs(9_000),
```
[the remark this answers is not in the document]
bnaecker: That's the recommended default, but I agree we can pull this and other configuration data out if it'd be helpful. Or at least make it a `const`.

opte-api/src/ip.rs
```diff
@@ -524,6 +543,10 @@ impl Ipv6Addr {
     pub fn safe_mask(self, mask: Ipv6PrefixLen) -> Self {
         self.mask(mask.val()).unwrap()
     }
+
+    pub const fn from_const(bytes: [u8; 16]) -> Self {
```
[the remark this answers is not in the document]
bnaecker: Ah, I remember why I didn't do that. It's pretty verbose, since you need to call `u16::to_be_bytes()` on each word in that array. But I agree it's probably worth it. I'm adding it now.

opte-api/src/ndp.rs
```diff
@@ -0,0 +1,114 @@
...
+impl RouterAdvertisement {
+    /// Create new `RouterAdvertisement`.
+    ///
+    /// The `src_mac` is the expected source MAC address a Router Solicitation
+    /// should come from. There are no restrictions on the source IP address,
+    /// other than that it be link-local, in `fe80::/10`.
+    ///
+    /// `mac` is the destination MAC the Solitications are expected to be
+    /// delivered to, and the source MAC of the generated Advertisement. The
+    /// IPv6 address is also derived from this.
```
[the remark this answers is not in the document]
bnaecker: Thanks, that's clearer.

opte-api/src/ndp.rs
```diff
@@ -0,0 +1,114 @@
...
+    // The IPv6 address the advertisement is sent from, which is the EUI-64
+    // transform of the source MAC address.
+    ip: Ipv6Addr,
+
+    /// If true, advertise that this neighbor is a router.
+    pub is_router: bool,
+}
+
+impl NeighborAdvertisement {
+    /// Create new `RouterAdvertisement`.
```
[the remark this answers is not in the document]
bnaecker: Good catch, thanks.

opte/src/engine/icmpv6.rs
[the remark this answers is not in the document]
bnaecker: Done, added some `const fn`s to make this all work. Good idea, thank you.

opte/src/engine/icmpv6.rs
```diff
@@ -172,3 +176,296 @@ impl HairpinAction for Icmpv6EchoReply {
...
+        if src_pkt.msg_code() != 0 {
+            return Ok(AllowOrDeny::Deny);
+        }
+        if !src_pkt.verify_checksum(&src_ip, &dst_ip) {
+            return Ok(AllowOrDeny::Deny);
+        }
+        // NOTE: The router is required to check that there is no Link-Layer
+        // Address Option, if the solicitation is sent from the unspecified
+        // address. The predicates here are that the solicitation came from the
+        // guest's EUI-64 transform link-local, so that is checked.
```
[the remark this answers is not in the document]
bnaecker: That's better, thank you.

opte/src/engine/icmpv6.rs
```diff
@@ -172,3 +176,296 @@ impl HairpinAction for Icmpv6EchoReply {
...
+        //
+        // TODO-completeness: Check IP Hop Limit and ICMP length / option length
```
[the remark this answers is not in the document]
bnaecker: I might have misunderstood, but I thought by the time we get here in the code, the IPv6 header has been parsed and extracted from the `Packet`. That is, we don't really have access to the bytes of the header anymore. Is that incorrect? I also thought it might be better to avoid bloating this PR by adding the hop limit to the IPv6 header, but that could be worth it. Thoughts?

opte/src/engine/icmpv6.rs
```diff
@@ -172,3 +176,296 @@ impl HairpinAction for Icmpv6EchoReply {
...
+            hop_limit: u8::MAX,
+            flags,
+            // Use the maximum advertised lifetime as a default router.
+            router_lifetime: Duration::from_secs(9_000),
+            // Do not specify the reachable or retrans time. Guests will decide
+            // that for themselves at this point.
+            reachable_time: Duration::from_millis(0),
+            retrans_time: Duration::from_millis(0),
+            lladdr: Some(RawHardwareAddress::from_bytes(&self.mac.bytes())),
+            mtu: Some(1500),
```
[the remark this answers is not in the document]
bnaecker: Ok, thanks. I'll refer to that issue, and I agree adding the MTU into the `VpcCfg` seems like a pretty easy and valuable solution.

To make sure I understand the point about headroom though. We can advertise 1500 to clients, and I recall agreeing at some point that we were sticking to that in the MVP. In that case, ensuring headroom means using a larger MTU in `xde`, right? I'm just checking that we want to advertise 1500 to _clients_, rather than use 1500 ourselves and advertise a smaller MTU to clients.

opte/src/engine/icmpv6.rs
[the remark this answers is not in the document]
bnaecker: Sorry, I think I got confused. 9000 isn't a default, it's maximum time allowed for the `AdvDefaultLifetime` configuration variable (see https://www.rfc-editor.org/rfc/rfc4861.html#section-6.2.1). I'm going to make this a constant -- the other times (reachable and retrans) are about the client's reachability detection, and we're basically saying here that clients decide how often to resend solicitations to implement that.

opte/src/engine/icmpv6.rs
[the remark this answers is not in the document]
rzezeski: Oh right, I'm sorry I was thinking that these are fields in the ICMP body, which is obviously not correct. You are correct that this stuff is not available here, and not really generally available at the moment. I agree skipping it in this PR is the right choice as it opens up a bigger can of worms around meta vs headers, and whether we should instead just start passing the entire headers around (since we copy them anyways) to allow for more complete inspection of a given packet. Something worth chatting about a bit next time we talk.

opte/src/engine/icmpv6.rs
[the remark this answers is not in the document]
rzezeski: Awesome, that clears things up. Sounds like a good plan.

opte/src/engine/icmpv6.rs
bnaecker: It's a good question about this particular case. I had assumed that an NS would only be received _after_ an RS, in which case the client would have a MAC address for the virtual gateway in their cache. At that point, clients are allowed to unicast the following NS or multicast it, as you pointed out.

The other reason I ignored the multicast is that we will drop any NS for _another_ neighbor. That is, we only want to respond to NSs with a target node of the gateway, and everything else needs to be determined as _off-link_. So the gateway could respond to solicited-node multicast NSs, assuming that the target node in that message is the gateway IPv6 address.

That could be an optimization though. I mean, clients could still be multicasting NSs, and the gateway _should_ respond if the target node is its IPv6 address.

This actually brings up a second point, something I think I missed. During Duplicate Address Detection, the client will be sending out NSs that _are_ multicast to the solicited-node multicast group defined by their candidate link-local IPv6 address. OPTE should mostly ignore those, with one exception: On the off-chance a client tries to claim _our_ IPv6 address, we need to wave them off!

All this is to say, I think we should probably add support for responding with NAs when the client sends a multicast NS that has our IPv6 address as the target node. This will force the client to pick a new link-local there's a duplicate, and is probably the expected behavior, even _if_ the client has our MAC in their cache.

oxide-vpc/src/engine/icmpv6.rs
bnaecker: Yeah, that's just wrong. I had something similar when I supported pings from the VPC-private IPv6 address of the client. I'll clean it up now.

oxide-vpc/src/engine/icmpv6.rs
bnaecker: Yeah, that's nice. I'll update it now, thanks.

opte/src/engine/icmpv6.rs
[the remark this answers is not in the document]
rzezeski: Yes, 1500 to clients/instances at this time. In our VPC case we decided we'd stick with 1500 for now but have the machinery to allow configuration in the future when we deem it's needed/wanted.

https://rfd.shared.oxide.computer/rfd/0021#design-mtu

So with decision in place we just need to make sure our transit links (T6 NICs and the switches) are configured with a large enough MTU to fit the VPC encap headers. This shouldn't affect traffic reaching out to the external network as the VPC encap only exists inside the rack/fleet -- so the external network shouldn't have to make any adjustments to handle the frames coming out of our switch's upstream ports.

oxide-vpc/src/engine/icmpv6.rs
bnaecker: I'm not sure I follow the Echo Request / Reply bit, actually. I think you're recommending just allowing _all_ ICMPv6 traffic, regardless of destination. I assume that means you're expecting the router layer to filter those on the basis of the source / destination IP addresses, is that right?

oxide-vpc/tests/integration_tests.rs
bnaecker: Sure, we can do that. I'm kinda playing fast-and-loose on the assumption we'll do a bulk format at some point.

oxide-vpc/tests/integration_tests.rs
bnaecker: Oh, nice didn't know about that. Thanks!

oxide-vpc/src/engine/icmpv6.rs
rzezeski: No, basically I'm saying to create an allowed list of ICMP types as opposed to your deny list. I think it would look something like the following (note I didn't actually verify this compiles, might not be 100% right).


```Rust
let mut echo = Rule::new(nactions + 1, Action::Static(Identity::new());
echo.add_predicate(Predicate::InnerEtherType(vec![EtherTypeMatch::Exact(
     ETHER_TYPE_IPV6,
 )]));
 echo.add_predicate(Predicate::InnerIpProto(vec![IpProtoMatch::Exact(
     Protocol::ICMPv6,
 )]));
// Actually it would be nice to make this a Vec of message types so you could just apply this rule to both echo and reply.
echo.add_data_predicate(DataPredicate::Icmpv6MsgType(MessageType::from(
     Icmpv6Message::Echo,
 )));
icmp.add_rule(Direction::Out, echo.clone().finalize());
icmp.add_rule(Direction::In, echo.finalize());

let deny = Rule::new(u16::MAX - 1, Action::Deny);
deny.add_predicate(Predicate::InnerEtherType(vec![EtherTypeMatch::Exact(
     ETHER_TYPE_IPV6,
 )]));
 deny.add_predicate(Predicate::InnerIpProto(vec![IpProtoMatch::Exact(
     Protocol::ICMPv6,
 )]));
icmp.add_rule(Direction::Out, deny.clone().finalize());
imcp.add_rule(Direction::Out, deny.finalize());
```

oxide-vpc/tests/integration_tests.rs
rzezeski: Yes, once we both have a pause point let's do bulk reformat. In this case I was just saying that the imports are used in more than one function, so might as well do it at module level.

bnaecker:
Ok, first round of review feedback is addressed. The big outstanding question is about adding NAs for NSs that are multicast to OPTE's solicited-node group. @rzezeski Let me know if you have any more thoughts on that, and I can address the rest of the questions. Looks like I'll need to rebase manually, so I'll probably squash everything then too, once the review is completed.

opte/src/engine/icmpv6.rs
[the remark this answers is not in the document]
bnaecker: Sounds good. I'll ignore this fow now, and I'm up for talking more about it next week.

opte/src/engine/icmpv6.rs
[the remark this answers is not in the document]
bnaecker: Yeah, that all makes sense, thanks.

oxide-vpc/src/engine/icmpv6.rs
bnaecker: So that all makes sense, and that's what I attempted to do in the follow-up commit. (Might have mucked that up, though.) I think I was trying to clarify whether there are predicates here on the destination or source IP addresses. For example, should we deny inbound ICMPv6 that's not destined for the client IP addresses? Or is the expectation that that denial would happen at the router or firewall?

Does that make sense?

bnaecker: BTW, I see the confusion. I wrote "allowing _all_ ICMPv6 traffic" in my reply. I meant "denying _all_ ICMPv6 traffic". So...yeah, that would be confusing :)

oxide-vpc/tests/integration_tests.rs
bnaecker: Ohhh, I interpreted that as the "module-level" grouping, the option to `rustfmt` that groups imports at the level of the module. You mean put the imports in the integration test module itself, rather than inside the function. My bad, yeah I get what you're saying now. I'll move them up.

opte/src/engine/icmpv6.rs
rzezeski: > This actually brings up a second point, something I think I missed. During Duplicate Address Detection, the client will be sending out NSs that are multicast to the solicited-node multicast group defined by their candidate link-local IPv6 address. OPTE should mostly ignore those, with one exception: On the off-chance a client tries to claim our IPv6 address, we need to wave them off!

So I'm not to worried about DAD here, as I would expect all guest OSs we care about use the EUI-64 of the NIC's MAC address to compute the link-local, right? However, reading 7.2.2, I get the feeling that a node is going to use the solicited-node multicast address to send an NS to the gateway (opte).

```
[7.2.2](https://www.rfc-editor.org/rfc/rfc2461#section-7.2.2).  Sending Neighbor Solicitations

   When a node has a unicast packet to send to a neighbor, but does not
   know the neighbor's link-layer address, it performs address
   resolution.  For multicast-capable interfaces this entails creating a
   Neighbor Cache entry in the INCOMPLETE state and transmitting a
   Neighbor Solicitation message targeted at the neighbor.  The
   solicitation is sent to the solicited-node multicast address
   corresponding to the target address.
```

Granted, I suppose our guest could choose to not have a multicast-capable interface, but I'm assuming that's always on by default (I could be wrong about that). Though, if you tested this, it must have worked, right? So I guess the guest did use unicast to send the NS to the gateway.

To me it seems like the safest bet is to look for a NS on both the unicast and solicited-node multicast address. That way we reply to the guest no matter how it decides to perform NS for its gateway.

bnaecker: > So I'm not to worried about DAD here, as I would expect all guest OSs we care about use the EUI-64 of the NIC's MAC address to compute the link-local, right?

My understanding that while this is the most common, there are some flavors of Linux (RHEL and Fedora) that may be configured to use a "privacy-preserving" link-local address. I really don't understand how that preserves any privacy, since the Ethernet frames will by definition have the MAC address in them, so you don't lose any more privacy by using that in an EUI-64 transform. But I gather it could happen.

> To me it seems like the safest bet is to look for a NS on both the unicast and solicited-node multicast address.

Agreed. I'll add the solicited-node multicast group to the predicates of OPTE's NA. Guests should get back the right NA no matter how they try to resolve it (unicast or multicast). Thanks for walking through that.

oxide-vpc/src/engine/icmpv6.rs
rzezeski: Okay, yea that typo had me very confused. And yes, I'm saying **deny all** ICMPv6 traffic that doesn't match the allow list.

The router doesn't even do any checks on inbound (maybe something to change in the future but I don't want to rat hole on that right now).

The firewall is user controlled and is separate from OPTE's own enforcement.

In this case the ICMP layer should make sure only the explicit ICMP traffic we allow is allowed to pass. For now let's just keep that to echo/reply (but we will probably need inbound DU support soonish).

There is still the larger issue of #42, where we make sure early that the guest doesn't have some malicious program using a raw socket/ip/L2 API to send out bogus traffic. That should mostly just be an outbound concern I think, but also separate from this PR as it applies to all traffic, not just ICMP. Though there is no harm tacking on the extra predicates to the rules in the this layer as well to make sure things are coming from the expected MAC address and IP.

opte/src/engine/icmpv6.rs
bnaecker: This ended up being a bit bigger than I expected. I beefed up the validation of NS messages from the client, and the generation of the NA from us on the basis of that NS data. I really just missed this in the initial implementation, and did not read the relevant RFC 2841 sections closely enough.

I pulled the NS validation into [this function](https://github.com/oxidecomputer/opte/blob/ndp-support/opte/src/engine/icmpv6.rs#L359), which is based off [section 7.1.1](https://www.rfc-editor.org/rfc/rfc4861.html#section-7.1.1).

I pulled the NA generation into [this function](https://github.com/oxidecomputer/opte/blob/ndp-support/opte/src/engine/icmpv6.rs#L424), which is based of [section 7.2.4](https://www.rfc-editor.org/rfc/rfc4861.html#section-7.2.4).

oxide-vpc/src/engine/icmpv6.rs
```diff
@@ -155,24 +159,74 @@ fn add_icmpv6_rules(
...
+        cfg.private_mac,
+    )]));
+    // To OPTE MAC.
+    rule.add_predicate(Predicate::InnerEtherDst(vec![EtherAddrMatch::Exact(
+        cfg.gateway_mac,
+    )]));
+    // From client IPs.
+    rule.add_predicate(Predicate::InnerSrcIp6(vec![
+        Ipv6AddrMatch::Exact(src_ips[0]),
+        Ipv6AddrMatch::Exact(src_ips[1]),
```
rzezeski, line 187: I don't think an ICMP echo going to another guest or external network can use the link-local address.

bnaecker: Oh duh, will remove.

oxide-vpc/src/engine/icmpv6.rs
```diff
@@ -155,24 +159,74 @@ fn add_icmpv6_rules(
...
+        ETHER_TYPE_IPV6,
+    )]));
+    // To client MAC. Note we don't have a predicate on the source MAC.
+    rule.add_predicate(Predicate::InnerEtherDst(vec![EtherAddrMatch::Exact(
+        cfg.private_mac,
+    )]));
+    // To client IPs.
+    rule.add_predicate(Predicate::InnerDstIp6(vec![
+        Ipv6AddrMatch::Exact(src_ips[0]),
+        Ipv6AddrMatch::Exact(src_ips[1]),
```
rzezeski, line 218: Same here, I don't see how a link-local address could reach from guest-to-guest or external-to-guest.

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2216,17 +2218,28 @@ fn gateway_router_advert_reply() {
...
 //
 // The MAC addresses are used to generate the IPv6 link-local addresses using
 // the EUI-64 transformation.
 fn gen_neighbor_solicitation(
     src_mac: &MacAddr,
     dst_mac: &MacAddr,
+    unicast: bool,
 ) -> Packet<Parsed> {
-    // The IPv6 addresses are the EUI-64 transform of the MAC addresses
+    // The source IPv6 addresses the EUI-64 transform of the MAC address.
```
rzezeski, line 2233: "is the"

opte-api/src/ip.rs
```diff
@@ -538,6 +546,11 @@ impl Ipv6Addr {
         Self::from_const([0xff02, 0, 0, 0, 0, 1, w0, w1])
     }
 
+    /// Return `true` if this is a solicited node multicast address.
+    pub fn is_solicited_node_multicast(&self) -> bool {
+        &self.inner[..12] == &[0xff, 0x02, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x01]
```
rzezeski, line 551: Does this also have to verify the `0xFF` in octet 13?

```
   The above multicast address is computed as a function of a node's
   unicast and anycast addresses.  The solicited-node multicast address
   is formed by taking the low-order 24 bits of the address (unicast or
   anycast) and appending those bits to the prefix
   FF02:0:0:0:0:1:FF00::/104 resulting in a multicast address in the
   range

      FF02:0:0:0:0:1:FF00:0000

   to

      FF02:0:0:0:0:1:FFFF:FFFF
```

bnaecker: It does yeah, thanks.

opte/src/engine/icmpv6.rs
```diff
@@ -346,18 +349,141 @@ impl HairpinAction for RouterAdvertisement {
...
+                src_pkt.msg_type(),
+                src_pkt.msg_code()
+            )));
+        }
+    };
+
+    // The target cannot be a multicast address.
+    if target_addr.is_multicast() {
+        return Err(GenErr::Unexpected(String::from(
+            "Received NS with multicast target address, dropping.",
```
rzezeski, line 392: We don't have to include the "dropping" in error messages. They bubble up as a `ProcessResult::Drop`.

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2360,7 +2362,12 @@ fn test_gateway_neighbor_advert_reply(
                 gateway's link-local IPv6 address, generated \
                 from the EUI-64 transform of its MAC",
             );
-            let expected_dst = Ipv6Addr::from_eui64(&g1_cfg.private_mac);
+            let expected_dst = if unicast {
+                Ipv6Addr::from_eui64(&g1_cfg.private_mac)
+            } else {
+                Ipv6Addr::from_eui64(&g1_cfg.private_mac)
+            };
```
rzezeski, line 2369: Something is off here. You switch on `unicast` but return same result. I think you meant for the dst to be either the guest's link-local address or the all-nodes multicast? Which makes it curious why this test isn't failing.

bnaecker: Yeah, I was looking at that myself just now. I must have lost some history or forgotten to stash something, since I remember changing that. I'm going to try to reconstruct it, thanks for flagging.

opte/src/engine/icmpv6.rs
```diff
@@ -346,18 +349,141 @@ impl HairpinAction for RouterAdvertisement {
...
+    }
+
+    // Set the ROUTER flag, if required.
+    let mut flags = NdiscNeighborFlags::empty();
+    flags.set(NdiscNeighborFlags::ROUTER, na.is_router);
+
+    // Even though this is NA is in response to an NS, if the source IP is
+    // UNSPEC, we must _not_ set the SOLICITED flag.
+    let src_is_unspec = src_ip == &Ipv6Addr::ANY_ADDR;
+    flags.set(NdiscNeighborFlags::SOLICITED, !src_is_unspec);
```
rzezeski, line 442: So I'm reading RFC 2461 4.4, and my reading of the `S` flag is that you don't set it if the **advertisement** is multicast, I don't think it care about how the NS came in. In fact, if a node sends an NS, isn't that by definition a solicitation for the target address? That is, the NA is in response to an NS, it was solicited. Am I missing something?

bnaecker: So, I'm basing all this off RFC 4861, which I think is the updated version of 2461. That said, I wrote the code from this paragraph:

> If the source of the solicitation is the unspecified address, the
   node MUST set the Solicited flag to zero and multicast the
   advertisement to the all-nodes address.  Otherwise, the node MUST set
   the Solicited flag to one and unicast the advertisement to the Source
   Address of the solicitation.

Which is paragraph 3 of this section: https://www.rfc-editor.org/rfc/rfc4861.html#section-7.2.4. I read that as requiring us to send an NA to all-nodes, if we got an NS from the unspecified address. Am I misinterpreting that?

bnaecker: I think actually that's the same thing as what you said:

> you don't set it if the advertisement is multicast

That's because we only multicast the NA if we got the NS from the unspecified address. So checking if the source is unspec is equivalent to checking if the destination is multicast. Does that sound right?

opte/src/engine/icmpv6.rs
```diff
@@ -172,3 +180,405 @@ impl HairpinAction for Icmpv6EchoReply {
...
+            ]
+        } else {
+            vec![Ipv6AddrMatch::Prefix(Ipv6Cidr::LINK_LOCAL)]
+        };
+
+        let hdr_preds = vec![
+            // We expect that the source MAC is the MAC provided to the client.
+            Predicate::InnerEtherSrc(vec![EtherAddrMatch::Exact(self.src_mac)]),
+            // It's directed to the virtual gateway's MAC address.
+            Predicate::InnerEtherDst(vec![EtherAddrMatch::Exact(self.mac)]),
```
bnaecker, line 485: This is not correct. We're only allowing NSs unicast to the gateway's MAC, which is clearly not correct in some cases. This could be multicast to all-nodes, or to the multicast MAC address derived solicited-node multicast group. In the first case that would be `33:33:0:0:0:2`, and in the second it would be `33:33:ff:ff:77:77`.

IPv6 is _so_ complicated.

opte/src/engine/icmpv6.rs
rzezeski: Oh right, duh. I believe you are correct. Honestly IPv6 makes all this very hard to think about. At this point if guests are resolving the gateway correctly I'm good to go. My eyes are bleeding reading these RFCs.

bnaecker:
Ok, I'm pretty sure all the comments have been addressed, and that the tests are much improved. There's some work in here I'm deferring, specifically, we currently may not be able to prevent guests from self-assigning OPTE's link-local IPv6 address. I've seen that happen in Debian, which tries to use the Secure Neighbor Discovery Protocol, which `smoltcp` does not support. See #265 for more details, but I'm not going to block this PR on that.

opte-api/src/ip.rs
```diff
@@ -932,10 +934,10 @@ impl Ipv6PrefixLen {
 }
 
 impl Ipv6Cidr {
-    /// The IPv6 link-local prefix, `fe80::/10`.
+    /// The IPv6 link-local prefix, `fe80::/64`.
```
rzezeski, line 937: Good catch.

opte/src/engine/icmpv6.rs
```diff
@@ -433,22 +431,26 @@ fn construct_neighbor_advert<'a>(
...
-    let mut flags = NdiscNeighborFlags::empty();
+    //
+    // Note from RFC 4861 Section 7.2.4 paragraph 2, we start with the OVERRIDE
+    // flag set. That says:
+    //
+    // > If the Target Address is either an anycast address or a unicast
+    // > address for which the node is providing proxy service, or the Target
+    // > Link-Layer Address option is not included, the Override flag SHOULD
+    // > be set to zero.  Otherwise, the Override flag SHOULD be set to one.
+    //
```
rzezeski, line 442: Thanks for adding this comment with a callout to the RFC. Stuff like this helps greatly in future debugging when some pour soul is debugging an IPv6 support ticket.

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2218,204 +2217,356 @@ fn gateway_router_advert_reply() {
...
+                src_ip: Ipv6Addr::ANY_ADDR,
+                dst_ip: Ipv6Addr::from_eui64(&cfg.private_mac)
+                    .solicited_node_multicast(),
+                target_addr: Ipv6Addr::from_eui64(&cfg.private_mac),
+                lladdr: None,
+            },
+            na: None,
+        },
+        // In this case, the client happens to pick our gateway's EUI-64
+        // link-local address. We _must_ sent back an NA, so that they configure
```
rzezeski, line 2385: s/sent/send/

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2218,204 +2217,356 @@ fn gateway_router_advert_reply() {
...
+        let res_dbg = format!("{:?}", &res);
+
+        match (res, d.na) {
+            (Ok(ProcessResult::Drop { .. }), None) => {
+                // Dropped the packet, as we expected
+                continue;
+            },
+            (Ok(Hairpin(hp)), Some(na)) => {
+                assert_port!(g1);
+                println!("\n{}\n{}", d.ns, na);
```
rzezeski, line 2563: Remove the println

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2218,204 +2217,356 @@ fn gateway_router_advert_reply() {
...
+    set_state!(g1, PortState::Running);
+    let mut pcap = PcapBuilder::new("gateway_neighbor_advert_reply.pcap");
+
+    let data = generate_solicit_test_data(&g1_cfg);
+    for d in data.into_iter().skip(1).take(1) {
+        let mut pkt = generate_neighbor_solicitation(&d.ns);
+        pcap.add_pkt(&pkt);
+        let res = g1.port.process(Out, &mut pkt, &mut ameta);
+        // Hack to deal appease borrow checker, since `res` isn't `Copy`.
+        let res_dbg = format!("{:?}", &res);
```
rzezeski, line 2554: The problem was that when you tried to format it directly in `panic!` it was borrowing the `res` from this lexical scope (don't ask me why I don't know). If you explicitly match in the panic arm with `(res, _)` you don't need to do this.

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2218,204 +2217,356 @@ fn gateway_router_advert_reply() {
...
+                // Dropped the packet, as we expected
+                continue;
+            },
+            (Ok(Hairpin(hp)), Some(na)) => {
+                assert_port!(g1);
+                println!("\n{}\n{}", d.ns, na);
+                validate_hairpin_advert(&mut pcap, hp, na);
+            }
+            _ => panic!(
+                "Generated unexpected packet from NS: {}\nResult: {}\nExpected: {:?}",
```
rzezeski, line 2567: Did you mean to print the NA as debug?

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2218,204 +2217,356 @@ fn gateway_router_advert_reply() {
...
+    let g1_cfg = g1_cfg();
+    let v2p = Arc::new(Virt2Phys::new());
+    let mut ameta = ActionMeta::new();
+    let mut g1 = oxide_net_setup("g1_port", &g1_cfg, v2p.clone());
+    g1.port.start();
+    set_state!(g1, PortState::Running);
+    let mut pcap = PcapBuilder::new("gateway_neighbor_advert_reply.pcap");
+
+    let data = generate_solicit_test_data(&g1_cfg);
+    for d in data.into_iter().skip(1).take(1) {
```
rzezeski, line 2549: This looks like leftover from when you were trying to just test one case.

oxide-vpc/tests/integration_tests.rs
```diff
@@ -2218,204 +2217,356 @@ fn gateway_router_advert_reply() {
...
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+struct SolicitTestData {
+    ns: SolicitInfo,
+    na: Option<AdvertInfo>,
+}
+
+// Generate the set of Neighbor Solicitations and their corresponding Neighbor
```
rzezeski, line 2321: Love all the description here about how this is all supposed to work.

rzezeski (changes requested):
This latest update is fantastic! Really love the cleanup and extra descriptions in the test. The only real issues I found were a handful of things in the test. Take care of that and I think we are good to go!

oxide-vpc/tests/integration_tests.rs
bnaecker: Nice catch, thanks. Fixed.

oxide-vpc/tests/integration_tests.rs
bnaecker: Oh, now _that's_ a good catch, thank you!

oxide-vpc/tests/integration_tests.rs
bnaecker: I believe I tried that, without success. IIRC, then we can't pass the `hp` variable we match in the second arm to the core validation function `validate_hairpin_advert`. If we pass `hp` by reference there, then we can't call `hp.parse()` inside the function, since that takes `mut self`. But if we pass `hp` by value, then we have a partial move out of `res`, meaning we can't put it in the `panic!` message. There are other ways around this, but it's all kinda hacky and fine for the tests.

oxide-vpc/tests/integration_tests.rs
bnaecker: I think that was me being sloppy. I implemented `Display` for these types so I didn't have to try to interpret `[u8]`s for the addresses, and then forgot to change that one. Thanks, I'll move it over now.

rzezeski (approved):

bnaecker:
Rebased on the latest master branch to resolve conflicts. Once this goes green, I'll merge it in.

Issue #261, opened 2022-09-26
Use CString/CStr from alloc/core


rzezeski:
Use `core::ffi::CStr` and `alloc::ffi::CString` now that they are stabilized.

https://github.com/rust-lang/rust/pull/99277

rzezeski:
Addressed in 64d24afb107e34811b594138715dbf78212f9eb5.

PR #262, opened 2022-09-26
PR: Use CString/CStr from alloc/core (#261)

rzezeski:


opte/src/engine/mod.rs
```diff
@@ -123,7 +124,7 @@ cfg_if! {
 
             unsafe {
                 if opte_debug != 0 {
-                    let cstr = cstr_core::CString::new(msg).unwrap();
+                    let cstr = CString::new(msg).unwrap();
```
bnaecker, line 127: Not really part of this PR, but we might want to truncate the message or something, rather than unwrapping.

opte/src/lib.rs
```diff
@@ -242,12 +242,8 @@ impl LogProvider for KernelLog {
             LogLevel::Error => ddi::CE_WARN,
         };
 
-        unsafe {
-            ddi::cmn_err(
-                cmn_level,
-                cstr_core::CString::new(msg.to_string()).unwrap().as_ptr(),
-            )
-        }
+        let msg_arg = alloc::ffi::CString::new(msg.to_string()).unwrap();
```
bnaecker, line 245: The extra `.to_string()` conversion here seems unnecessary, I think. `CString::new` takes anything convertible to a `Vec<u8>`, which includes `&str`.

bnaecker (approved):
Nice, pretty straightforward. Couple of small things, but looks good to me.

opte/src/engine/mod.rs
rzezeski: IIRC, the unwrap is only in case this is invalid UTF-8, and I see no way for that to ever be the case given that all these should be strings coming from opte code or from network implementations using opte code. I.e., I can't ever see this string coming from bytes passed down by the user. But in any event, truncation does make sense in terms of line length. We really never should be making long debug/failure messages. And so we could probably replace this with something like a static 1024 byte buffer like we've talked about for the SDT probes.

opte/src/lib.rs
rzezeski: Done.

bnaecker (approved):

PR #273, opened 2022-10-05
Add some docs to checksum module

rzezeski:
- Simplified the process of finalizing a rolling checksum and converting it into a header checksum.

- Removed `pub` access from `csum_sub()` as it's meant to be a private helper function.


bnaecker (approved):
Thanks, these details are all really helpful.

Issue #208, opened 2022-08-08
Fix cargo doc warnings


rzezeski:
Fix the current cargo doc warnings and add a check to the buildomat job to fail on future warnings.

```
rpz@sector5 [~/src/oxidecomputer/opte-xde-devs/opte]                                                         
$ cargo doc --features=api,std,engine,vpc                                                                    
 Documenting opte v0.1.0 (/Users/rpz/src/oxidecomputer/opte-xde-devs/opte)                                   
warning: unresolved link to `Packet`                                                                         
  --> src/engine/headers.rs:54:62                                                                            
   |                                                                                                         
54 |     /// Read a zerocopy version of the raw header from the [`Packet`]                                   
   |                                                              ^^^^^^ no item named `Packet` in scope     
   |                                                                                                         
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default                                          
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`                           
                                                                                                             
warning: unresolved link to `Rule`                                                                           
   --> src/engine/icmp.rs:123:63                                                                             
    |                                                                                                        
123 | /// serde implementation; allowing this value to be used in [`Rule`]                                   
    |                                                               ^^^^ no item named `Rule` in scope       
    |                                                                                                        
    = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`                          
                                                                                                             
warning: unresolved link to `Layer`                                                                          
  --> src/engine/ioctl.rs:36:31                                                                              
   |
36 | /// *name*: The name of the [`Layer`] to dump.
   |                               ^^^^^ no item named `Layer` in scope
   |
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`

warning: unresolved link to `EndOfPacket`
    --> src/engine/packet.rs:1949:11
     |
1949 |     /// [`EndOfPacket`] is returned.
     |           ^^^^^^^^^^^ no item named `EndOfPacket` in scope
     |
     = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`

warning: unresolved link to `crate::rule::Action`
    --> src/engine/port.rs:1753:10
     |
1753 | /// [a]: crate::rule::Action
     |          ^^^^^^^^^^^^^^^^^^^ no item named `rule` in module `opte`

warning: unresolved link to `Resoruce`
    --> src/engine/port.rs:1868:26
     |
1868 |         /// Remove the [`Resoruce`].
[...]

PR #274, opened 2022-10-05
Fix cargo doc warnings (#208)

rzezeski:
I left oxide-vpc broken for this first push to make sure the job catches failure.


bnaecker (approved):
Nice cleanup, thanks.

Issue #281, opened 2022-10-17
Add VNI filter to oxide-vpc firewall


plotnick:
For firewall rules with target or host filter type "VPC", we'd like to filter directly on the VNI associated with that VPC rather than filtering on each subnet in the VPC.

From the discussion on [Omicron#1756]( https://github.com/oxidecomputer/omicron/issues/1756#issuecomment-1281229748):
> We could add first class support for this type of firewall rule in oxide-vpc itself, but it would require a little work. On decap we would extract the VNI from Geneve and store it as action metadata to pass through the port's processing pipeline. The firewall layer could then add support for filtering on this value. This would be the most efficient in terms of control plane churn and oxide-vpc implementation. I kind of like this idea but not sure there are any edge cases I'm not thinking of.

See also [RFD 21 table 14](https://rfd.shared.oxide.computer/rfd/0021#_vpc_subnet_apis).

rzezeski:
Addressed in 486ffd8fa0de96e6a542fcc46d866e3ce6d23d1e.

PR #284, opened 2022-10-18
Opte 281 vpc firewall

rzezeski:
Still a little more testing to do. But this should solve #281.


plotnick:
Looks good so far, I'll work up a draft Omicron PR that uses the new `Address::Vni` variant instead of VPC subnets. Should be straightforward.

PR #296, opened 2022-11-01
Use opte-api version as pkg minor

rzezeski:


rzezeski:
This seems to have created the package correctly.

```
rpz@kalm:~$ pkgrepo list -s opte.p5p 
PUBLISHER     NAME                                          O VERSION
helios-netdev driver/network/opte                             0.18.139:20221101T212758Z
```

PR #306, opened 2022-12-01
opteadm: use anyhow instead of unwrapping.

luqmana:
Ran opteadm without pfexec and noticed it just panics for for `list-ports`. Looks like that was the only command using `unwrap()` instead of `unwrap_or_die()`. Just added anyhow which lets us throw in `?` instead everywhere.


rzezeski (approved):
This simplifies the code, thank you.

Issue #74, opened 2022-03-22
Xde + Propolis/Viona Integration Tracking Issue


rcgoodfellow:
This issue is to track integration with viona.

## Background

The first pass at xde/viona integration did not "just work". While xde provides a mac interface and viona consumes a mac interface, the situation is not that simple. When viona sets up a mac client for the underlying network device it's using, the [mac_client_open call](https://github.com/oxidecomputer/illumos-gate/blob/master/usr/src/uts/i86pc/io/viona/viona_main.c#L727) as a zero value for flags. This means that either a subsequent `mac_unicast_add` is needed, or the device is a vnic which results in a tx data path implicitly. Alternatively, if the `MAC_OPEN_FLAGS_NO_UNICAST_ADDR` were used, like [xde does](https://github.com/oxidecomputer/opte/blob/master/xde/src/xde.rs#L802) on its underlying devices, a basic tx data path would be set up.

## Interim Solution

In the interim, for viona integration, we will just layer vnics atop xde devices and have viona use vnics. This should work.

## Solution

There are a few avenues to take here. The following list is a starting point, feel free to add others.

- **_Conditionally_** use `MAC_OPEN_FLAGS_NO_UNICAST_ADDR` in viona.
- **_Conditionally_** use `mac_unicast_add` in viona.
- Figure out what would need to be added to xde to allow it to work in the same way vnic does where this implicit tx path is created for mac clients.

## Related

Propolis currently assumes the underlying network device for a port is going to be a vnic [here](https://github.com/oxidecomputer/propolis/blob/master/propolis/src/hw/virtio/viona.rs#L62). This would need to be generalized. Perhaps [netadm-sys](https://oxidecomputer.github.io/netadm-sys/libnet/index.html) could be helpful here.

luqmana:
Looking at the possible solutions, is using `MAC_OPEN_FLAGS_NO_UNICAST_ADDR` in viona enough? 

In [mac_client_open](https://github.com/oxidecomputer/illumos-gate/blob/06b650b02f3e47978b5928d2fe9cc3e5fae49585/usr/src/uts/common/io/mac/mac_client.c#L1461-L1463):
```c
	/*
	 * We will do mimimal datapath setup to allow a MAC client to
	 * transmit or receive non-unicast packets without waiting
	 * for mac_unicast_add.
	 */
	if (mcip->mci_state_flags & MCIS_NO_UNICAST_ADDR) {
[...]

Issue #70, opened 2022-03-16
wire up promisc support for xde


rzezeski:
We should be able to snoop an xde device. Although this brings up a bit of a tantalizing question: should the promisc path give us the top side (guest), bottom side (physical/underlay aka the part after OPTE transforms the packet), or both? I know "both" kind of sounds crazy, as it means opening up a pcap in Wireshark would probably throw up all sorts of warnings, but in some ways it also makes sense to show both. In kind of makes me wish there was a way for snoop/promisc to know it is dealing with a filtering device and have an option to specify which side you are interested in. But for now I guess it makes sense to just pass the top-side (guest) traffic, as we can still get at the physical traffic at the underlying device, though it means snooping 2 devices and it means adding a filter for the specific guest traffic you are interested in. Although that means we need to be able to filter based on inner frame. Is that possible? I imagine the snoop language supports checking byte sequences at arbitrary offsets -- which isn't nearly as clean as proper language support but it would work.

EDIT: Actually, maybe capturing both sides wouldn't cause any problems with wireshark, as it's just going to view it as two separate flows I suppose. So perhaps that's the best place to start. I'm sure we'll figure out more when we implement this.

rzezeski:
@bnaecker and I are going through all the issues and I wanted to make a note that we don't **need** this until we get rid of the VNIC that sits atop the xde device.

luqmana:
After removing the VNIC and using the xde device directly with propolis I realized snoop wasn't working:

```console
$ pfexec snoop -d vm_test0 -rV
snoop: cannot open "vm_test0": No such device or address
```

("No such device or address" is the error string for `ENXIO`).

Dug into it a bit and found the culprit: https://github.com/oxidecomputer/opte/blob/adc8cdceb3ddd6576704d18569a9eca8033b5a49/xde/src/xde.rs#L1114

When `snoop` goes to open the device it ends up calling `nodev()` which always returns `ENXIO (6)`:

```
  0  25378                     nodev:return = 6                                                                                                                                                         
              genunix`dev_to_instance+0x4f                                                                                                                                                                  
              genunix`e_ddi_hold_devi_by_dev+0x47                                                                                                                                                           
              specfs`spec_open+0x2f5                                                                                                                                                                        
              genunix`fop_open+0x96                                                                                                                                                                         
              genunix`vn_openat+0x20c                                                                                                                                                                       
              genunix`copen+0x421                                                                                                                                                                           
[...]

PR #308, opened 2022-12-17
Remove VNIC workaround

luqmana:
The OPTE side changes for https://github.com/oxidecomputer/opte/issues/74 enabling `viona` to use an xde device directly without an intermediate VNIC. The [viona side](https://code.illumos.org/c/illumos-gate/+/2558) changes are not yet merged but wanted to put this up for folks to see.

~~Also includes a fix for https://github.com/oxidecomputer/opte/issues/70 to enable `snoop` on xde devices.~~

~~And a slight change to the arp rules for the external ip hack; seems like after the gateway layering changes incoming ARP packets get dropped because they hit the firewall layer first. The rules added to explicitly allow them are in the gateway layer which would only be hit after getting past firewall. Given all this is a hack anyways, I just modified the rules to get added to the firewall layer for now.~~


oxide-vpc/src/engine/gateway/arp.rs
```diff
@@ -82,7 +85,8 @@ pub fn setup(
                 cfg.guest_mac,
             )));
             let rule = Rule::new(1, action);
-            layer.add_rule(Direction::In, rule.finalize());
+            // Incoming ARP requests will hit the "firewall" layer first so add the rule there.
+            pb.add_rule("firewall", Direction::In, rule.finalize())?;
```
rzezeski, line 89: These changes should not be here. The firewall layer should only allow what control plane tells it to allow. Instead, for testing, you can use the `opteadm add-fw-rule` command to add an inbound rule to allow all traffic, which includes ARP.

```
$ opteadm add-fw-rule --action allow --dir in --hosts any --ports any --priority 10 --protocol any -p <port name>
```

It's also worth noting that the "external IP hack" is going away ASAP.

luqmana: Ah, I wanted to test other firewall rules worked as expected (e.g. SSH/ICMP-in) and I couldn't see a way to add more fine-grained rules to match just ARP. Since it is just a hack I'll remove that commit

opte-api/src/lib.rs
```diff
@@ -56,7 +56,7 @@ pub use ulp::*;
 ///
 /// We rely on CI and the check-api-version.sh script to verify that
 /// this number is incremented anytime the oxide-api code changes.
-pub const API_VERSION: u64 = 19;
```
rzezeski, line 59: Apologies for the delay here. This change looks good, though I don't see why the API version needs to be bumped.

luqmana: yup, don't need this anymore, removed it

rzezeski (approved):
This looks good. Thanks again.

xde/src/xde.rs
```diff
@@ -1308,7 +1302,8 @@ static mut xde_cb_ops: cb_ops = cb_ops {
 static mut xde_devops: dev_ops = dev_ops {
     devo_rev: DEVO_REV,
     devo_refcnt: 0,
-    devo_getinfo: xde_getinfo,
+    // If this is None (represented NULL), it will be set to `dld_getinfo` via `mac_init_ops`.
```
luqmana, line 1305: I'd just drop this comment now as we've long since added our own `getinfo(9e)`.

crates/illumos-sys-hdrs/src/kernel.rs
```diff
@@ -211,12 +211,14 @@ unsafe impl Sync for cb_ops {}
 pub struct dev_ops {
     pub devo_rev: c_int,
     pub devo_refcnt: c_int,
-    pub devo_getinfo: unsafe extern "C" fn(
-        dip: *mut dev_info,
-        infocmd: ddi_info_cmd_t,
-        arg: *mut c_void,
-        result: *mut *mut c_void,
-    ) -> c_int,
+    pub devo_getinfo: Option<
```
luqmana, line 214: I know this is technically a more accurate representation, but we don't actually set it to `None`/`NULL` anymore and realistically this crate is only used by xde. So I'm a bit inclined to revert it?

xde/src/xde.rs
```diff
@@ -765,15 +767,7 @@ fn create_xde(req: &CreateXdeReq) -> Result<NoResp, OpteError> {
...
-    // TODO Total hack to allow a VNIC atop of xde to have the guest's
-    // MAC address. The VNIC **NEEDS** to have the guest's MAC address
-    // or else none of the ethernet rule predicates will match.
-    //
-    // The real answer is to stop putting VNICs atop xde. The xde
-    // device needs to sit in the place where a VNIC would usually go.
-    mreg.m_src_addr = EtherAddr::from([0xA8, 0x40, 0x25, 0x77, 0x77, 0x77])
-        .to_bytes()
-        .as_mut_ptr();
+    mreg.m_src_addr = guest_addr.as_mut_ptr();
```
luqmana, line 770: Let's bump the API version as this is not really a transparent change

luqmana (commented):
Thanks for picking this up Kyle, a few small nits but LGTM!

(I can't approve since this was originally my PR tho 😅)

crates/illumos-sys-hdrs/src/kernel.rs
FelixMcFelix: I wasn't sure if we wanted to keep it around in case `mac` registration were to make its way into illumos-rs or another crate of that nature -- although that work would probably be undertaken with a close eye wrt the manpages.

PR #310, opened 2022-12-22
Update for stable toolchain

luqmana:
With `asm_sym` stabilized, everything but xde can be built with stable.

### DIFF
--- a/.github/buildomat/jobs/opte-api.sh
+++ b/.github/buildomat/jobs/opte-api.sh
@@ -24,7 +24,7 @@ header "check API_VERSION"
 ./check-api-version.sh
 
 header "check style"
-ptime -m cargo fmt -- --check
+ptime -m cargo +nightly fmt -- --check
 
 header "analyze std"
 ptime -m cargo check
--- a/opte/Cargo.toml
+++ b/opte/Cargo.toml
@@ -27,8 +27,8 @@ illumos-sys-hdrs = { path = "../illumos-sys-hdrs" }
 kstat-macro = { path = "../kstat-macro" }
 opte-api = { path = "../opte-api", default-features = false }
 postcard = { version = "0.7.0", features = ["alloc"], default-features = false }
-zerocopy = "0.6.1"
 usdt = { version = "0.3.5", optional = true }
+zerocopy = "0.6.1"
 
 [dependencies.serde]
 version = "1.0"
--- a/opte/Cargo.toml
+++ b/opte/Cargo.toml
@@ -17,6 +17,7 @@ std = ["opte-api/std"]
 # Used for declaring methods which are useful for integration testing.
 #
 test-help = []
+usdt = ["std", "dep:usdt"]
 
 [dependencies]
 cfg-if = "0.1"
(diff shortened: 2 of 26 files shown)


opte/build.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright 2022 Oxide Computer Company
```
rzezeski, line 1: I haven't bothered to add any copyright/license to these basic build-related files. I have no strong opinion in that regard, it just didn't seem necessary to me. That said, as per RFD 224, I put OPTE under the MPL.

opte/src/lib.rs
```diff
@@ -4,31 +4,26 @@
...
-// This allows one to run the tests on macOS with the usdt probes
-// enabled.
-//
-// ```
-// cargo test --features=usdt
-// ```
-#![cfg_attr(target_os = "macos", feature(asm_sym))]
-#![feature(extern_types)]
-#![feature(vec_into_raw_parts)]
+#![cfg_attr(not(feature = "std"), no_std)]
```
rzezeski, line 7: Ah, that's a nice trick!

.github/buildomat/jobs/opte-api.sh
```diff
@@ -24,7 +24,7 @@ header "check API_VERSION"
 ./check-api-version.sh
 
 header "check style"
-ptime -m cargo fmt -- --check
+ptime -m cargo +nightly fmt -- --check
```
rzezeski, line 27: I feel like I'm asking something obvious, but: if we got most of the crates off nightly, why is it required here? Same goes for other places where `check` is called without the kernel feature.

opte/Cargo.toml
```diff
@@ -51,5 +43,10 @@ default-features = false
 #
 features = ["alloc", "medium-ethernet", "proto-ipv4", "proto-ipv6", "proto-dhcpv4", "socket", "socket-raw"]
 
+[build-dependencies]
+version_check = "0.9.4"
+
 [dev-dependencies]
```
rzezeski, line 49: So it looks like usdt is always enabled now for test? That's good, I should have done that to begin with, as I can see from you other modifications I let some of the probe sites rot.

rzezeski (commented):
Thanks for this improvement. Since opte has grown organically from a single crate to multiple I kind of got lazy and everything stayed on nightly when it didn't need to. This is a nice change to limit the scope of nightly to only where it's required. I also like how you cleaned up usdt handling and have it always enabled, preventing future rot from creeping in. Now that I'm looking at that with fresh eyes, it might make sense to just get rid of the `usdt` feature have have those probes built anytime we are in `std`? I haven't thought that through, and it should be its own commit/PR if we do that, but I thought I'd throw that idea out there.

I just had a few small comments on the code.

opte/build.rs
luqmana: ah, nice catch! I had copied the script from one of our other repos. I'll just remove this for now. We can do a bulk add as follow up later.

.github/buildomat/jobs/opte-api.sh
luqmana: Just need it for `fmt` specifically because the `rustfmt.toml` we have uses some unstable options that need a nightly. The `check` ones that follow don't use nightly otherwise.

opte/Cargo.toml
luqmana: Yup, tested on both illumos and macOS. Seems to work:

```
$ sudo dtrace -Z -f "*opte*" | rustfilt
dtrace: system integrity protection is on, some features will not be available

dtrace: description '*opte*' matched 0 probes
CPU     ID                    FUNCTION:NAME
  6  35374 opte::engine::port::Port<N>::port_process_entry_probe:port-process-entry 
  6  35372 opte::engine::layer::Layer::layer_process_entry_probe:layer-process-entry 
  6  35377 opte::engine::layer::RuleTable::rule_match_probe:rule-match 
  4  35374 opte::engine::port::Port<N>::port_process_entry_probe:port-process-entry 
[...snip...]
```

rzezeski (approved):
Looks good. Thanks!

Issue #236, opened 2022-08-19
Remove the "external IP hack"


rzezeski:
Once we have the ability to simulate Boundary Services we'll want to remove all code related to the "external IP hack". Below is a list of things to check for code to remove. It's not simply a matter of reverting the commits listed below, as some legit fixes are also intertwined. However, deleting anything commented with `XXX-EXT-IP` or gated behind `xde_ext_ip_hack` should allow the compiler to make sure you clean everything up.

- [ ] Grep for all instances of `XXX-EXT-IP` in the code.
- [ ] Commit `52176b99cc578f5b7b90f9bae8935013f961086c`
- [ ] Commit `23884d35aa7908e23accaa77f125a370ddf5c606`
- [ ] Commit `3ea62a9b77f58dbbbf7dd772362b0d008c548e87`
- [ ] Commit `1f8c0a1ed65b296da3876017ea0040bc1b4c6601`
- [ ] Commit `92a3ae21ffd9405a6897b5136410e2b11ee74005`
- [ ] Commit `14410f9420184702fa4b0ee134a0aa37216398eb`

luqmana:
Remaining OPTE bits removed in #311

PR #311, opened 2022-12-23
Remove the "external IP hack" (#236)

rzezeski:


bnaecker:
I just wanted to check about timing here. There's a lot of integration with Dendrite needed before we can really provision instances that don't use the hack. Is the plan to keep this up and in-waiting until we do that? That's fine by me, just checking.

rzezeski:
@bnaecker I just wanted to make sure I had a PR floated before I take leave. Though, to be fair, it could always be merged and you just have to just stay on the previous commit until you are ready to transition Omicron to the new world order (speaking of which, I believe we are already several PRs behind in Omicron that would update opte).

Anyways, I'll leave it to you when this should land.

bnaecker:
Oh duh, I forgot we pin a commit in Omicron. Thanks!

luqmana:
Ideally I'd want to get the no-more-vnic landed first and be able to pin omicron at a version with that. That's just currently blocked on landing the upstream viona changes.

luqmana:
With SoftNPU having landed a couple of weeks ago, omicron has since transitioned from requiring the external IP hack. It's finally time to get rid of it altogether!

PR #316, opened 2023-01-07
Allow filtering ARP packets in fw rule.

luqmana:
I know the external IP hack is on the way out but in the meantime you need to add a firewall rule to allow incoming ARP packets past the firewall layer.

It's currently possible to do this via adding an overly permissive rule:

```console
$ opteadm add-fw-rule -p vm_test0 --action allow --dir in --hosts any --ports any --priority 65534 --protocol any
$ opteadm dump-layer -p vm_test0 firewall
[...snip...]
Inbound Rules
ID     PRI    HITS   PREDICATES                             ACTION            
0      65534  0      *                                      "Stateful Allow"
```

With this PR you can make that a bit more fine grained:

```console
$ opteadm add-fw-rule -p vm_test0 --action allow --dir in --hosts any --ports any --priority 65534 --protocol arp
$ opteadm dump-layer -p vm_test0 firewall
[...snip...]
Inbound Rules
ID     PRI    HITS   PREDICATES                             ACTION            
0      65534  0      inner.ether.ether_type=ARP             "Stateful Allow"
```

### DIFF
--- a/oxide-vpc/src/api.rs
+++ b/oxide-vpc/src/api.rs
@@ -721,6 +721,7 @@ impl Display for Address {
 #[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub enum ProtoFilter {
     Any,
+    Arp,
     Proto(Protocol),
 }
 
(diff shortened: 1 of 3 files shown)


oxide-vpc/src/api.rs
```diff
@@ -730,6 +731,7 @@ impl FromStr for ProtoFilter {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s.to_ascii_lowercase().as_str() {
             "any" => Ok(ProtoFilter::Any),
+            "arp" => Ok(ProtoFilter::Arp),
```
rcgoodfellow, line 734: Should this be

```rust
ProtoFilter::Proto(Protocol::ARP)
```

to be consistent with the others?

rcgoodfellow: Ahh I see. Yeah seems like `ProtoFilter::IPProto` would be better there.

rcgoodfellow (approved):
LGTM!

PR #334, opened 2023-02-13
xde: underlay handling cleanup

luqmana:
Taking a look at `xde_rx` and had a couple fixes and some refactoring around handling different mac handles.

### DIFF
--- a/xde/src/mac.rs
+++ b/xde/src/mac.rs
@@ -177,17 +177,18 @@ impl MacClientHandle {
     ) -> Result<MacPromiscHandle, c_int> {
         let mut mph = 0 as *mut mac_promisc_handle;
 
-        // This extra ref is cleaned up in `MacPromiscHandle::drop()`
-        let arg = Arc::into_raw(self.clone()) as *mut c_void;
+        // `MacPromiscHandle` keeps a reference to this `MacClientHandle`
+        // until it is removed and so we can safely access it from the
+        // callback via the `arg` pointer.
+        let mch = self.clone();
+        let arg = Arc::as_ptr(&mch) as *mut c_void;
         let ret = unsafe {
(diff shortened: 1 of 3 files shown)


xde/src/mac.rs
```diff
@@ -110,30 +168,30 @@ impl MacClient {
...
         ptype: mac_client_promisc_type_t,
         promisc_fn: mac_rx_fn,
-        arg: *mut c_void,
         flags: u16,
-    ) -> Result<*mut mac_promisc_handle, c_int> {
+    ) -> Result<MacPromiscHandle, c_int> {
         let mut mph = 0 as *mut mac_promisc_handle;
 
+        // This extra ref is cleaned up in `MacPromiscHandle::drop()`
+        let arg = Arc::into_raw(self.clone()) as *mut c_void;
```
rcgoodfellow, line 181: Can we avoid the extra ref by just doing `let arg = Arc::as_ptr(self) as *mut c_void` here?

xde/src/xde.rs
```diff
@@ -927,144 +936,78 @@ unsafe extern "C" fn xde_attach(
...
-                errno: EFAULT,
-                msg: format!("mac_promisc_add failed for {}: {}", u2, e),
-            });
-        }
-    };
+    let u1_mph = u1_mch
+        .add_promisc(
+            mac::mac_client_promisc_type_t::MAC_CLIENT_PROMISC_ALL,
+            xde_rx,
+            mac::MAC_PROMISC_FLAGS_NO_TX_LOOP,
```
rcgoodfellow, line 983: Could you add a comment describing what prompted adding this flag?

rcgoodfellow (approved):
LGTM!

Issue #335, opened 2023-02-17
opte kernel panic


jmpesp:
Saw the following panic:

```
james@kibblesnbits:~$ fmdump -Vp -u ecd5a6f0-6c1d-e74d-eeee-d5a29265d851
TIME                           UUID                                 SUNW-MSG-ID
Feb 17 2023 20:59:12.711862000 ecd5a6f0-6c1d-e74d-eeee-d5a29265d851 SUNOS-8000-KL

  TIME                 CLASS                                 ENA
  Feb 17 20:59:12.6936 ireport.os.sunos.panic.dump_available 0x0000000000000000
  Feb 17 20:58:57.6792 ireport.os.sunos.panic.dump_pending_on_device 0x0000000000000000

nvlist version: 0
        version = 0x0
        class = list.suspect
        uuid = ecd5a6f0-6c1d-e74d-eeee-d5a29265d851
        code = SUNOS-8000-KL
        diag-time = 1676667552 695222
        de = fmd:///module/software-diagnosis
        fault-list-sz = 0x1
        fault-list = (array of embedded nvlists)
        (start fault-list[0])
        nvlist version: 0
                version = 0x0
                class = defect.sunos.kernel.panic
                certainty = 0x64
                asru = sw:///:path=/var/crash/kibblesnbits/.ecd5a6f0-6c1d-e74d-eeee-d5a29265d851
                resource = sw:///:path=/var/crash/kibblesnbits/.ecd5a6f0-6c1d-e74d-eeee-d5a29265d851
                savecore-succcess = 1
                dump-dir = /var/crash/kibblesnbits
                dump-files = vmdump.1
                os-instance-uuid = ecd5a6f0-6c1d-e74d-eeee-d5a29265d851
                panicstr = panicked at 'called `Result::unwrap()` on an `Err` value: PacketSeg { mp: 0xfffffeb1ddc0ec20, dblk: 0xfffffeb1ddd38d00, len: 68, avail: 208 }', /work/oxidecomputer/opte/opte/src/engine/packet.rs:801:
[...]

PR #336, opened 2023-02-18
Handle longer chains of mblk segments

bnaecker:
We previously used a statically-sized `heapless::Vec` inside the `Packet` type. That was used to represent the segments of a larger `mblk_t`, those linked by the `b_cont` field. This static array had enough space for 4 elements, and any attempt to push more segments panicked. We've now hit that.

This uses a straight-up Rust `Vec` instead of the `heapless` variant, to handle arbitrarily sized linked-lists of `mblk_t`s. This will address the immediate issue of panics. I've included a regression test for this.

However, we pay the cost of an allocation now. I've taken care that it's only one, by walking the `mblk_t` chain once to compute the size, allocating the `Vec`, and then walking the chain again to call `PacketSeg::wrap_mblk` on each. We may also pay an allocation cost when rewriting headers in some cases. Both of these should be measured when we get to performance issues.

### DIFF
--- a/opte/Cargo.lock
+++ b/opte/Cargo.lock
@@ -3,21 +3,6 @@
 version = 3
 
 [[package]]
-name = "atomic-polyfill"
-version = "0.1.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e3ff7eb3f316534d83a8a2c3d1674ace8a5a71198eba31e2e2b597833f699b28"
-dependencies = [
- "critical-section",
-]
-
-[[package]]
-name = "autocfg"
-version = "1.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
-
-[[package]]
 name = "bitflags"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -58,12 +43,6 @@ checksum = "28d997bd5e24a5928dd43e46dc529867e207907fe0b239c3477d924f7f2ca320"
 dependencies = [
  "libc",
 ]
-
-[[package]]
-name = "critical-section"
-version = "1.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6548a0ad5d2549e111e1f6a11a6c2e2d00ce6a3dafe22948d67c2b443f775e52"
 
 [[package]]
 name = "crypto-common"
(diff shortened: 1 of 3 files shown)


bnaecker:
This addresses #335. We would also like to audit the allocations get a sense of their cost. I'm not yet sure how to do that.

rmustacc:
> This addresses #335. We would also like to audit the allocations get a sense of their cost. I'm not yet sure how to do that.

To audit the cost of the allocations and figure out their relative priority I think the thing we'll want to do is set up streams going from VM to VM across two different sleds and within the same sled and then start to diagnose and understand the performance and where that's coming from.

The reason I suggest we get this up and understand it first as while allocations can inherently be bad, we should first measure and see if that is our top contributor. Some ways to do that is to start with setting this up and then setting up a flame graph and using guest tools to understand packet drops and related. It can be a little tricky because things will often enter a steady state.

One way that might make this easier to understand just the OPTE cost and to leverage debugging tools would be to set this up manually in the GZ. This would help eliminate the cost of forwarding from the zone and allow us to more readily compare with other pieces here. There is probably going to be a lot of perf work like making sure we're requesting or performing certain hardware checksumming and TSO features and related, so it may be that there are a lot of things to do before we worry about the memory allocation.

luqmana:
The change here seems pretty straight forward and does address the immediate issue.

But thinking on how we could avoid that extra allocation: is there any reason we can't use the linked-list of mblk's as-is? While convenient, I don't think we particularly need `O(1)` random access of segments. It would mean a fair bit of refactoring so I'm with rm on just getting things up running first without panics and then measuring.

jmpesp:
I can confirm that this fixes the panic seen in the issue.

bnaecker:
@luqmana I don't think there's any fundamental reason we can't use the `mblk_t` list already. The `Packet` and `PacketSeg` types provide a safer, idiomatic-Rust interface to them. @rzezeski may also have had other reasons for putting this in place that I'm not aware of. But in this case, if it comes down to a safe, array-based interface that requires allocations; and a minimal wrapper around the _existing_ linked-list of `mblk_t`s that does not, I would say we should pick the latter. But you're right that it'll require some work, and that we should merge this and perform that work incrementally.

Thanks @jmpesp! Do you happen to know if you still see `mblk_t` chains with more than 4 segments? I don't know that it's crucial, but it'd be nice to confirm that we're actually handling the previously-problematic packets.

jmpesp:
@bnaecker Using the following D script:

```d
fbt:xde:xde_rx:entry {
    p = (mblk_t *)arg2;
    if (p->b_cont != 0) {
        p = (mblk_t *)p->b_cont;
        if (p->b_cont != 0) {
            p = (mblk_t *)p->b_cont;
            if (p->b_cont != 0) {
                p = (mblk_t *)p->b_cont;
                if (p->b_cont != 0) {
                    print("larger than 4!");
                }
            }
        }
    }
}
```

I'm seeing the print fire:

```
james@dinnerbone:~$ pfexec ./trace_mblks.d 
dtrace: script './trace_mblks.d' matched 11 probes
CPU     ID                    FUNCTION:NAME
 27  10579                     xde_rx:entry string "larger than 4!"

 27  10579                     xde_rx:entry string "larger than 4!"

 27  10579                     xde_rx:entry string "larger than 4!"
 ```

luqmana (approved):

PR #338, opened 2023-02-21
Add some more conversions for IP types.

luqmana:
We already had some `From` impls to go between the standard rust IP types and our own. Added the other direction as well as conversions for the types from the `ipnetwork` crate which omicron makes use of.

### DIFF
--- a/opte-api/src/ip.rs
+++ b/opte-api/src/ip.rs
@@ -332,6 +332,26 @@ impl From<Ipv6Addr> for IpAddr {
     }
 }
 
+#[cfg(any(feature = "std", test))]
+impl From<std::net::IpAddr> for IpAddr {
+    fn from(ip: std::net::IpAddr) -> Self {
+        match ip {
+            std::net::IpAddr::V4(ipv4) => Self::Ip4(ipv4.into()),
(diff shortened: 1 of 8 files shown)


luqmana:
Don't really need to bump the API version here as the user-kernel API hasn't changed.

opte-api/src/ip.rs
```diff
@@ -814,6 +855,26 @@ impl FromStr for IpCidr {
     }
 }
 
+#[cfg(feature = "ipnetwork")]
```
bnaecker, line 858: I don't see an `"ipnetwork"` feature in `opte-api/Cargo.lock`. Did you mean to add one, or make this `feature = "std"`? This shows up a few other places too.

bnaecker (commented):
Small question about feature naming, but the actual `From` implementations look good to me!

opte-api/src/ip.rs
luqmana: Optional deps in `Cargo.toml` implicitly define a corresponding feature.

bnaecker: I did not know that! Cool!

bnaecker (approved):

Issue #337, opened 2023-02-21
opte seems to be adding 40 bytes to payload size


jmpesp:
After adc8cdc, opte seems to be adding 40 bytes to the payload size of outgoing GENEVE packets. Here's an example from a ping to an instance (10.1.0.101):

```
16:20:59.953213 IP6 (hlim 63, next-header UDP (17) payload length: 114) fd03:1122:3344:101::15.47783 > fd00:1122:3344:101::1.6081: [udp sum ok] Geneve, Flags [none], vni 0xba043f
        IP (tos 0x0, ttl 64, id 49097, offset 0, flags [DF], proto ICMP (1), length 84)
    10.0.0.1 > 10.1.0.101: ICMP echo request, id 21890, seq 2, length 64

16:20:59.953341 IP6 truncated-ip6 - 40 bytes missing!(hlim 128, next-header UDP (17) payload length: 154) fd00:1122:3344:101::1.7777 > fd03:1122:3344:101::15.6081: [bad udp cksum 0x0000 -> 0x6f76!] Geneve, Flags [none], vni 0x63
        IP (tos 0x0, ttl 64, id 41518, offset 0, flags [DF], proto ICMP (1), length 84)
    10.1.0.101 > 10.0.0.1: ICMP echo reply, id 21890, seq 2, length 64
```

Notice that tcpdump says `40 bytes missing!`

The ICMP reply packet arrives at the sidecar but is dropped due to the length check:

```
$ pfexec dtrace -n 'ip_drop_input:entry,ip_drop_output:entry { trace(stringof(arg0)); stack(); }'
dtrace: description 'ip_drop_input:entry,ip_drop_output:entry ' matched 2 probes
CPU     ID                    FUNCTION:NAME
 12  65401              ip_drop_input:entry   ipIfStatsInTruncatedPkts         
              ip`ip_check_length+0x131
              ip`ill_input_short_v6+0xca
              ip`ip_input_common_v6+0x283
              ip`ip_input_v6+0x1f
              ip`ip_rput_v6+0x71
              unix`putnext+0x233
              dld`dld_str_rx_fastpath+0x37
              dls`i_dls_link_rx+0x311
              mac`mac_rx_deliver+0x2e
              mac`mac_rx_soft_ring_process+0x1be
              mac`mac_rx_srs_proto_fanout+0x29a
              mac`mac_rx_srs_drain+0x2e5
              mac`mac_rx_srs_process+0x123
              mac`mac_rx_common+0x213
[...]

PR #340, opened 2023-02-21
Fix setting IPv6 total length with extension headers

bnaecker:


opte/src/engine/ip6.rs
```diff
@@ -435,7 +447,7 @@ impl<'a> Ipv6Hdr<'a> {
     /// Return the total length of the packet, including the base header, any
     /// extension headers, and the payload itself.
     pub fn total_len(&self) -> usize {
-        usize::from(self.base.payload_len()) + Self::BASE_SIZE
+        self.pay_len() + Self::BASE_SIZE
```
luqmana, line 450: Should this also be updated to `self.hdr_len() + self.pay_len()`?

bnaecker: I don't think so. `Ipv6Hdr::hdr_len()` returns the length _including the extension headers_. To be clear, I am not sure that it _should_ return that, but that's what it currently does and what the documentation claims that it does.

luqmana: Ah ok, so that's a bit different than the methods on `Ipv6Meta` which is a bit confusing. Would it be too much to update one of the other so we're consistent

bnaecker: I agree that it's all extremely confusing. That said, I see `Ipv6Meta::hdr_len` as returning `BASE_SIZE + self.ext_len`, which I think is consistent with `Ipv6Hdr`, no?

bnaecker: I see part of the issue. The `Ipv6Meta::total_len()` method appears incorrect as well. That returns `self.hdr_len() + self.pay_len`. But `Ipv6Meta::hdr_len()` should return the `BASE_SIZE + self.ext_len`. We are currently using the `pay_len` field to set the actual Payload Length field of the IPv6 base header, which implies that it includes the length of the extension headers. Those two things can't both be true: either `pay_len` includes the extension headers, and the `Ipv6Meta::total_len()` method is wrong; or `pay_len` does _not_ include the extension headers, and we're setting the IPv6 Payload Length field incorrectly.

bnaecker: I should say, part of the confusion comes from the fact that we have _similar_ methods for IPv4 and IPv6. I say similar, because they're just named the same thing, they are not trait methods or anything statically constrained to be the same by the compiler.

That's a problem, because the protocols are so different that they don't really have the same concepts. It's not clear what you mean when you say "header length" or "payload length" in the case of IPv6. We have methods that return all of the different pieces (base header length, base + extensions, just extensions, IPv6 Payload Length == extensions + ULP, and ULP length), but it's not clear to me they're being used correctly.

bnaecker: Alright, I've found the issue. I'm going to push that onto this PR, since it's pretty small and really all related. @luqmana Let me know if you think there is more cleanup or documentation to be added around the various IPv6 methods. It's a bit confusing, but I believe it's now consistent and I've added some more documentation.

bnaecker:
This should now resolve #337

opte/src/engine/packet.rs
```diff
@@ -1396,7 +1396,19 @@ impl Packet<Parsed> {
             }
 
             Some(IpMeta::Ip6(ip6)) => {
-                ip6.pay_len = (new_pkt_len - pkt_offset) as u16;
+                // IPv6 Payload Length field is defined in RFC 2640 section 3
+                // as:
+                //
+                // > Length of the IPv6 payloaed, i.e., the rest of the packet
```
luqmana, line 1402:
```suggestion
                // > Length of the IPv6 payload, i.e., the rest of the packet
```

luqmana (approved):

opte/src/engine/packet.rs
bnaecker: Nice catch thanks. I wrote it while reading it, to make sure I _understood the damn thing_.

bnaecker:
Sorry about tagging you for review again, didn't see you'd already approved modulo the typo. Thanks @luqmana!

luqmana:
Np! Thanks for the fix!

Issue #346, opened 2023-03-03
panicked at 'called `Result::unwrap()` on an `Err` value: MaxCapacity(8096)'


jmpesp:
I was trying to reproduce the hang seen in https://github.com/oxidecomputer/propolis/issues/322 today on my setup, but managed to panic the host machine instead!

```
WARNING: panicked at 'called `Result::unwrap()` on an `Err` value: MaxCapacity(8096)', /work/oxidecomputer/opte/opte/src/engine/port.rs:1890:17

panic[cpu26]/thread=fffffeb1da1177a0: panicked at 'called `Result::unwrap()` on an `Err` value: MaxCapacity(8096)', /work/oxidecomputer/opte/opte/src/engine/port.rs:1890:17

fffffe00f732ee40 xde:rust_begin_unwind+146 ()
fffffe00f732ee80 xde:_ZN4core9panicking9panic_fmt17hbd42276bcaa2dfd8E+2d ()
fffffe00f732ef00 xde:_ZN4core6result13unwrap_failed17h5b67be868f6dc91bE+6c ()
fffffe00f732f310 xde:_ZN4opte6engine4port13Port$LT$N$GT$7process17h9bca109443cf9deaE+2bdf ()
fffffe00f732fc40 xde:xde_mc_tx+32f ()
fffffe00f732fc70 mac:mac_ring_tx+43 ()
fffffe00f732fcc0 mac:mac_provider_tx+85 ()
fffffe00f732fd80 mac:mac_tx+295 ()
fffffe00f732fe40 viona:viona_tx+48c ()
fffffe00f732fea0 viona:viona_worker_tx+73 ()
fffffe00f732fef0 viona:viona_worker+126 ()
fffffe00f732ff00 unix:thread_start+b ()

dumping to /dev/zvol/dsk/rpool/dump, offset 65536, content: kernel
```

I did the following:

- use qemu to boot `alpine-virt-3.17.2-x86_64.iso` and install it to an nvme drive:

```
truncate -s 2G install.raw

sudo qemu-system-x86_64 \
	  -enable-kvm \
	  -m 2g \
	  -bios /usr/share/OVMF/OVMF_CODE.fd \
	  -cdrom /home/jwm/Downloads/alpine-virt-3.17.2-x86_64.iso \
	  -drive file="install.raw",if=none,format=raw,id=nvme1 \
	  -device nvme,drive=nvme1,serial=nvme-1,addr=0x4
```

[...]

Issue #315, opened 2022-12-31
Need to expire TIME_WAIT connections


rzezeski:
Currently we perform no expiration of `TIME_WAIT` entries in the TCP flow table. For example, here are connections sitting in `TIME_WAIT` that are well past their time to be reclaimed. I don't think there's any reason to overcomplicate this, as we ultimately leave it up to the guest's stack to determine if a connection is allowed or not. So as a first strike I think it's good enough to just expire these after some hard-coded amount of time as part of the `expire_periodic()` in `xde`.

```
rpz@kalm:~/oxidecomputer/opte/opteadm$ pfexec opteadm dump-tcp-flows -p opte1
FLOW                                             STATE        HITS     SEGS IN  SEGS OUT BYTES IN   BYTES OUT 
TCP:172.30.0.6:22:10.0.0.10:53134                ESTABLISHED  176      87       90       8671       15117     
TCP:172.30.0.6:35480:172.30.0.5:5201             TIME_WAIT    36       18       19       1533       1721      
TCP:172.30.0.6:46190:172.30.0.5:5201             TIME_WAIT    35       17       19       1467       1722      
TCP:172.30.0.6:46534:172.30.0.5:5201             TIME_WAIT    35       17       19       1465       1721      
TCP:172.30.0.6:52788:172.30.0.5:5201             TIME_WAIT    35       17       19       1464       1725      
```

FelixMcFelix:
Closed by #444.

PR #347, opened 2023-03-03
Do not unwrap on flow table limits

bnaecker:
- Add error type for handling TCP flow errors
- Return new error variants rather than `String` from methods processing TCP flows.
- Handle TCP flow errors, now just unexpected segments, in the port processing code. This is done by returning a new `ProcessError` variant.
- Add `ProcessError::FlowTableFull`, which wraps attempts to add a new flow entry which fail because the table is full. This includes the table "kind", which for now is just a static string, and the current limit.
- Add an integration test verifying that TCP packets are dropped when the flow table is full.

### DIFF
--- a/opte/src/engine/port.rs
+++ b/opte/src/engine/port.rs
@@ -34,6 +34,7 @@ use super::rule::HdrTransformError;
 use super::rule::Rule;
 use super::tcp::TcpState;
 use super::tcp_state::TcpFlowState;
+use super::tcp_state::TcpFlowStateError;
 use super::HdlPktAction;
 use super::NetworkImpl;
 use crate::ddi::kstat;
@@ -84,6 +85,9 @@ pub enum ProcessError {
     HandlePkt(&'static str),
     HdrTransform(HdrTransformError),
     WriteError(super::packet::WriteError),
+    MissingFlow(InnerFlowId),
+    TcpFlow(TcpFlowStateError),
+    FlowTableFull { kind: &'static str, limit: u64 },
 }
 
 impl From<super::HdlPktError> for ProcessError {
@@ -1219,6 +1223,7 @@ impl<N: NetworkImpl> Port<N> {
     }
 }
 
+#[derive(Debug)]
 enum TcpMaybeClosed {
     Closed { ufid_inbound: Option<InnerFlowId> },
     NewState(TcpState),
@@ -1410,7 +1415,7 @@ impl<N: NetworkImpl> Port<N> {
         data: &mut PortData,
         pmeta: &PacketMeta,
         pkt_len: u64,
-    ) -> result::Result<TcpState, String> {
+    ) -> result::Result<TcpState, ProcessError> {
         use Direction::In;
 
         // All TCP flows are keyed with respect to the outbound Flow
@@ -1450,11 +1455,11 @@ impl<N: NetworkImpl> Port<N> {
                         Ok(tcp_state)
                     }
 
-                    Err(e) => Err(e),
+                    Err(e) => Err(ProcessError::TcpFlow(e)),
                 }
             }
 
-            None => Err(format!("TCP flow missing: {}", ufid_out)),
+            None => Err(ProcessError::MissingFlow(ufid_out)),
         }
     }
 
@@ -1466,7 +1471,7 @@ impl<N: NetworkImpl> Port<N> {
         ufid_in: &InnerFlowId,
         pmeta: &PacketMeta,
         pkt_len: u64,
-    ) -> result::Result<TcpState, String> {
+    ) -> result::Result<TcpState, ProcessError> {
         use Direction::In;
 
         // All TCP flows are keyed with respect to the outbound Flow
@@ -1509,7 +1514,7 @@ impl<N: NetworkImpl> Port<N> {
                         Ok(tcp_state)
                     }
 
-                    Err(e) => Err(e),
+                    Err(e) => Err(ProcessError::TcpFlow(e)),
                 };
 
                 // We need to store the UFID of the inbound packet
@@ -1532,7 +1537,7 @@ impl<N: NetworkImpl> Port<N> {
                 let tcp_state = match res {
                     Ok(TcpState::Closed) => return Ok(TcpState::Closed),
                     Ok(tcp_state) => tcp_state,
-                    Err(e) => return Err(e),
+                    Err(e) => return Err(ProcessError::TcpFlow(e)),
                 };
 
                 let tfes = TcpFlowEntryState::new_inbound(
@@ -1540,9 +1545,15 @@ impl<N: NetworkImpl> Port<N> {
                     tfs,
                     pkt_len,
                 );
-                // TODO kill unwrap
-                tcp_flows.add(ufid_out.clone(), tfes).unwrap();
-                Ok(tcp_state)
+                match tcp_flows.add(ufid_out.clone(), tfes) {
+                    Ok(_) => Ok(tcp_state),
+                    Err(OpteError::MaxCapacity(limit)) => {
+                        Err(ProcessError::FlowTableFull { kind: "TCP", limit })
+                    }
+                    Err(_) => unreachable!(
+                        "Cannot return other errors from FlowTable::add"
+                    ),
+                }
             }
         }
     }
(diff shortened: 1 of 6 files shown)


bnaecker:
Fixes #346.

This PR removes some of the lingering `unwrap()` calls around TCP behavior. When establishing a new connection, we try to create a flow-table entry for it. Those have thus far been statically-sized in the `xde` driver, to 8096 active connections, for example. It seems this limit has been hit in #346.

There's not really a great static limit we can apply here. This PR uses the `VpcCfg` provided for a guest, specifically the number of available ports, to size the flow tables. If one provisions a guest with 1024 ports, that's the limit of the UFT and TCP flow tables. If one has the full port range for an IPv4 and IPv6 address, then the guest gets `2 * u16::MAX` entries.

I've verified that this panic is fixed with a new integration test. That creates a tiny flow-table of 1 entry, and then tries to send two outbound packets through the OPTE port. Here's the behavior of that integration test on `master`:

```
bnaecker@feldspar : ~/opte/oxide-vpc $ cargo t -- flow_table_full
   Compiling oxide-vpc v0.1.0 (/home/bnaecker/opte/oxide-vpc)
    Finished test [unoptimized + debuginfo] target(s) in 4.14s
     Running unittests src/lib.rs (target/debug/deps/oxide_vpc-fa7540b92438ab30)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 6 filtered out; finished in 0.00s

     Running tests/firewall_tests.rs (target/debug/deps/firewall_tests-a3f255194b2ae2a7)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 3 filtered out; finished in 0.00s

     Running tests/integration_tests.rs (target/debug/deps/integration_tests-fc02c98ac1f45784)

running 1 test
test no_panic_on_flow_table_full ... FAILED

failures:

---- no_panic_on_flow_table_full stdout ----
thread 'no_panic_on_flow_table_full' panicked at 'called `Result::unwrap()` on an `Err` value: MaxCapacity(1)', /home/bnaecker/opte/opte/src/engine/port.rs:1820:55
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    no_panic_on_flow_table_full

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 23 filtered out; finished in 0.00s

error: test failed, to rerun pass `--test integration_tests
```

On this PR branch, that test passes fine.

One thing I'm not sure about is really whether this is the right solution. There are two aspects to this. First, we're sizing both the TCP and UFT flow tables to be the number of ports. But the UFT flow table will include non-TCP connections, suggesting we might want that to be larger. It seems overkill to make the table large enough to hold all possible protocols, though, so I don't know the right answer.

The other aspect is the failure mode when a guest tries to open a connection and the table is exhausted. We just drop the packet. I don't think there's anything else we can reasonably do, but it still feels odd. For example, a guest might have 16k entries in their table. But from the perspective of the guest OS, there's no real reason that one couldn't open a socket on all 65K ports. With this fix, they'd just see some of those time out. Like I said, I'm not sure what else one can do, since we can't really inject an RST or something back into the guest without confusing the crap out of them. @rmustacc and @rcgoodfellow I'd love to hear if you have any thoughts on that.

But the TL;DR is that this should address the panic itself.

bnaecker:
Ah, I also just realized this may be alleviated by a solution to #315. It's possible there are not 8096 _active_ connections, but there have been more than 8096 connections made at _some_ point. We should certainly get rid of these unwraps, but we may want to revisit the choices of size made here (based on the VPC config).

jmpesp (approved):
I tested this, and the host no longer panics in the same situation. :shipit:

luqmana:
Thanks for addressing the panic! Was out Friday so just taking a look today.

Reading through the issue a bit now, I'd propose a different approach: adding an entry to the flow tables should be infallible. We already expect flows to expire after some time (or just be removed explicitly in the case of TCP when the state machine gets to close/reset). If the table's full, we should just retire the earliest flow (*). Of course that still requires tweaking table size so we're not constantly ping-ponging flows in and out.

(*) have to check but not sure if the current structure of the code allows easily determining earliest flows, in which case I'd say just clearing the table all together is not the worst short-term option. All flows are already currently hard-coded with a TTL of 60s today AFAIK.

bnaecker:
Thanks @luqmana. That approach seems plausible. I don't see a quick way to look up the flows in time order now, so we'd need to add a "secondary index" to `FlowTable` that points to `InnerFlowId`s or `FlowEntry`s ordered by time. I didn't realize flows were expired on a period now, or at all. That seems to be done by the time of last hit, which is good.

In any case, I'm not what to make of this design. Talking out loud a bit, let's suppose we picked an arbitrary flow table limit, as we currently do, of N. All connections up to N would behave the same way, in that we'd set up a UFT and then keep hitting it as long as those are active. For connections made beyond N however, flows would start to behave less predictably. First, we'd have some initial setup time to find and remove the oldest flow and insert the new one. Second, there'd be some complicated interactions, where adding new connections would start to impact others. The impacted ones are "old," so that might be fine. It seems particularly hard to reason about in a guest that has a lot of active connections, such as a database server or load balancer. That might be an edge case though, I really don't know.

luqmana:
I agree but there is already some amount of unpredictability given the slow-path (traverse layers), fast-path (lookup in UFT) design. But I think that's better than just dropping the packets. It's definitely something we can tweak with how long before we expire flows and how big of a table we use. But that's already where we're at today.

Another not fully fleshed out idea is maybe just skip creating a table entry if it's full but otherwise deliver the packet as normal. Downside is subsequent packets will end up hitting slowpath but by then perhaps previous flows would've expired?

bnaecker:
That's fair about the existing design. I would imagine it's slow once, and then fast after that (assuming enough consistent activity to not expire the flow). You're definitely right though, some amount of this is already there.

It's a good point that we'd degrade more gracefully if we drop old flows as new ones above the static limit come in. As this is currently implemented, those TCP connections would just be slow to establish until / unless other flows were expired based on time. It might be better to have it driven by time _and_ guest activity.

By "skip creating the table entry," you mean the UFT table? So that would mean the flow would always be layer-processed? That's an interesting idea. It's similar to the first proposal, except newer flows can't bump older ones out of the UFT. Those later flows would be slower but more predictable, IIUC. That sounds better that the FIFO-eque behavior of the UFT, given how sensitive TCP can be to small latency variations.

I think I'm on board with one or both of your proposals, with a slight preference for the second (not adding a UFT entry if it's full, and using layer-processing instead).

luqmana:
> By "skip creating the table entry," you mean the UFT table? So that would mean the flow would always be layer-processed?

Right, at the point we'd be running into this we're already in the slow-path for that flow. Not returning an error/dropping the packet then would mean a subsequent packet would hit the slow-path again but hopefully by then we have space. 

In the slightly longer term we should probably start thinking of different workloads & usecases to start collecting perf data. We can add some probes/stats in this case to see how often we run into this situation and use that to better size the table.

opte/src/engine/port.rs
```diff
@@ -1410,7 +1415,7 @@ impl<N: NetworkImpl> Port<N> {
         data: &mut PortData,
         pmeta: &PacketMeta,
         pkt_len: u64,
-    ) -> result::Result<TcpState, String> {
+    ) -> result::Result<TcpState, ProcessError> {
```
luqmana, line 1418: I would use a separate error here so you don't need all the `unreachable!` cases for errors that don't apply. You can then add that as a variant to `ProcessError` if you want.

bnaecker:
> I agree but there is already some amount of unpredictability given the slow-path (traverse layers), fast-path (lookup in UFT) design. But I think that's better than just dropping the packets.

I've been thinking more about this. A lot of this is notes for myself, but it hopefully helps others.

I was having some trouble squaring this with the ideas in RFD 21 and the current implementation in Nexus. In particular, we limit SNAT IP addresses to 16K ports. Let's suppose we follow this proposal, of "falling back" to layer-processing when a guest hits the size limit for a UFT table. I couldn't quite grok how our 16K limit is actually _limiting_ anything though.

As a concrete example, if we allow 16K + 1 connections, what port do we rewrite the outgoing packets with? We can't pick anything outside the range that OPTE's `SNat{4,6}Cfg` has been provided, because those don't belong to us. Those are ports on an IP on the customer's network, and another VM may already be using them.

So it _seemed_ like we'd need to drop _something_, say this new connection or the oldest. Both of those suck as a customer experience. There's nothing inherent in the VM that would appear to be limiting. The `connect(2)` call doesn't give some errno indicating a lack of resources. All they'd see is a black hole for data sent on that socket.

The think I'm doing here is conflating two distinct objects: the unified flow table, and the NAT pool. The former is really just to store the "compiled" set of rules applying to a particular flow, as an optimization avoiding iterating over each layer and action for every packet. The latter is the pool of IPs (always one right now) _and ports_ that are available to the guest for use. There are two flavors of this: `opte::engine::nat::Nat`, which is a 1-1 NAT that allows using the whole port range, and `NatPool`, which is used to support a _limited_ number of outbound connections.

These are different things. The WIP I have here makes them the same _size_, so that every possible connection a guest can make is equally snappy. But they are still different things, and different limits and policies may apply to them.

One thing that's confusing here is that, the current `NatPool` will definitely just start dropping packets if the guest exceeds the number of ports the control plane has allocated. We have issues tracking how we can learn if this happens, and whether we need to bump the number of ports for each guest. But today, if you are in such a guest, connection 16K + 1 will just appear to fly off into space.

All of this is to say that (1) the proposal of less efficient processing is fine, (2) limiting the size of the pool to the number of connections the guest can make is also fine (I think?), but unnecessary. If the guest _is in fact_ limited, because they've an SNAT IP with 16K ports only, then the packets will already be dropped because the `oxide-vpc` NAT layer will just fail to generate an `ActionDesc` as the SNAT resource (ports, in this case) has been exhausted.

rcgoodfellow:
@bnaecker can we merge this? It fixes a crash we are encountering in system-level testing.

bnaecker:
@luqmana It looks like the fix here is blocking some other work, so I'd like to merge this. I'll file a new issue to track moving towards what we discussed: new flows are either always layer-processed; displace the oldest flow; or result in entirely clearing the UFT when it's full.

PR #350, opened 2023-03-18
Update ICMPv6 checksum, set correct IPv6 payload len and misc

luqmana:
Fixes:
1. We weren't setting the right hop limit (255) for the Neighbour Advertisement & Router Advertisement packets sent to the guest. 
2. Filled in some missing validity checks for Neighbour Solicitation & Router Solicitation packets received from the guest.
3. We were erroneously including the IPv6 base header size in the payload length when updating modified inner headers.
4. ICMPv6 checksums are calculated including an IP psuedo-header which meant the SNAT transformations would render them invalid. We now treat ICMPv6 as a psuedo ULP and handle checksum updates like TCP/UDP.

Misc:
1. Allow `opteadm` to create a port using IPv6.
2. Accept `icmp6` as a protocol filter when adding a firewall rule via `opteadm`.

Future:
It would be nice to be able to filter specific ICMP message types in the firewall (i.e. allow incoming echo request/reply but drop upstream NS)

### DIFF
--- a/opte/src/engine/icmpv6.rs
+++ b/opte/src/engine/icmpv6.rs
@@ -373,12 +373,7 @@ impl HairpinAction for RouterAdvertisement {
 
         // Collect the src / dst IP addresses, which are needed to emit the
         // resulting ICMPv6 packet using `smoltcp`.
-        let (src_ip, dst_ip) = if let Some(metadata) = meta.inner_ip6() {
-            (
-                IpAddress::Ipv6(Ipv6Address(metadata.src.bytes())),
-                IpAddress::Ipv6(Ipv6Address(metadata.dst.bytes())),
-            )
-        } else {
+        let Some(ip6) = meta.inner_ip6() else {
             // Getting here implies the predicate matched, but that the
             // extracted metadata indicates this isn't an IPv6 packet. That
             // should be impossible, but we avoid panicking given the kernel
@@ -388,10 +383,14 @@ impl HairpinAction for RouterAdvertisement {
                 meta
             )));
         };
+        let src_ip = IpAddress::Ipv6(Ipv6Address(ip6.src.bytes()));
+        let dst_ip = IpAddress::Ipv6(Ipv6Address(ip6.dst.bytes()));
+
         let body = rdr.copy_remaining();
         let src_pkt = Icmpv6Packet::new_checked(&body)?;
-        let src_ndisc =
-            Icmpv6Repr::parse(&src_ip, &dst_ip, &src_pkt, &Csum::ignored())?;
+        let mut csum = Csum::ignored();
+        csum.icmpv6 = Checksum::Rx;
+        let src_ndisc = Icmpv6Repr::parse(&src_ip, &dst_ip, &src_pkt, &csum)?;
 
         if !matches!(
             src_ndisc,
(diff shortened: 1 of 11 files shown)


opte/src/engine/icmpv6.rs
```diff
@@ -349,6 +349,8 @@ impl HairpinAction for RouterAdvertisement {
             dst: meta.inner_ip6().unwrap().src,
             proto: Protocol::ICMPv6,
             next_hdr: IpProtocol::Icmpv6,
+            // RFC 4861 7.1.2 requires that the hop limit be 255 in an RA.
+            hop_limit: 255,
```
bnaecker, line 474: Nit, but 7.1.2 is about validating NAs. The section specifying this for RAs is 6.1.2.

luqmana: Thanks, good catch. Looks like the guest I was using (alpine + dhcpd) doesn't check but we should be setting the same for NAs too.

luqmana:
Ok, I updated the PR description and added some tests. Should be ready for review now.

opte/src/engine/packet.rs
```diff
@@ -1498,7 +1548,19 @@ impl Packet<Parsed> {
...
+                // > Length of the IPv6 payload, i.e., the rest of the packet
+                // > following this IPv6 header, in octets. (Note that any
+                // > extension headers [section 4] present are considered part
+                // > of the payload, i.e., included in the length count.)
+                //
+                // So we need to remove the size of the fixed header (40
+                // octets), which is included in the total new packet length,
+                // when setting the payload length.
+                ip6.pay_len =
+                    (new_pkt_len - pkt_offset - Ipv6Hdr::BASE_SIZE) as u16;
```
bnaecker, line 1563: I thought this looked familiar! I'm confused, I see this exact diff [here](https://github.com/oxidecomputer/opte/pull/340/commits/ff9873b4fba809c7f8817ff336938ece89c35f03) from #340.

oxide-vpc/tests/common/mod.rs
```diff
@@ -120,7 +120,7 @@ pub fn g1_cfg() -> VpcCfg {
             gateway_ip: "fd00::1".parse().unwrap(),
             snat: Some(SNat6Cfg {
                 external_ip: "2001:db8::1".parse().unwrap(),
-                ports: 1025..=4096,
+                ports: 4097..=8192,
```
bnaecker, line 123: Just curious, why this change?

bnaecker (approved):
Looks great, thanks for the careful checks on lengths! Merge whenever you're ready.

opte/src/engine/packet.rs
luqmana: Right, that PR fixed the calculation on the outer header (`Packet::emit_outer_headers`) but not the inner header (`Packet::emit_inner_headers`).

oxide-vpc/tests/common/mod.rs
luqmana: While adding the `guest_to_internet_ipv6` test I had initially left one assert referring to the v4 SNAT config. To hopefully reduce those kinds of mix ups I made this change to use different ranges for IPv4 vs IPv6.

Also, previously `g1_cfg` and `g2_cfg` were using the same port range and IP for the v6 SNAT config.

PR #361, opened 2023-05-02
prevent zero-sized continuation blocks

rcgoodfellow:
When emitting headers, the structure of the message block for the packet being modified is somtimes altered. If the block holding the current headers is too small for the new headers, we shrink the old block by the size of the old headers, allocate a new block for the new headers and link the new block to the previous block through its continuation pointer. If the block being shurnk also contains ULP data this is fine. However, if it's purely a header block, then we wind up shrinking the old block to zero, and our new block has a zero sized continuation. This is not a generally expected situation and has caused NIC hardware to stop working. This commit prevents this from happening.

I placed a panic in `PacketSeg::link` that fires if we try to make a zero-sized continuation (only during tests) to see if any tests currently trigger this behavior. It turns out several of the integration tests in `oxide-vpc` do.

### DIFF
--- a/opte/src/engine/packet.rs
+++ b/opte/src/engine/packet.rs
@@ -1409,12 +1409,24 @@ impl Packet<Parsed> {
                     assert_eq!(body.seg_offset - old_hdr_len, 0);
                     body.seg_offset = 0;
                 }
-                seg.link(&segs[0]);
-
-                // TODO-performance: This may necessitate another allocation. We
-                // will want to measure how often we hit this branch, and the
-                // impact of the allocation.
-                segs.insert(0, seg);
+                if segs[0].len() > 0 {
+                    seg.link(&segs[0]);
+                    // TODO-performance: This may necessitate another allocation. We
+                    // will want to measure how often we hit this branch, and the
+                    // impact of the allocation.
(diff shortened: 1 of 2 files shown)


rcgoodfellow:
Potentially fixes 
- https://github.com/oxidecomputer/stlouis/issues/392

opte/src/engine/packet.rs
```diff
@@ -1865,7 +1887,23 @@ impl PacketSeg {
     }
 
     fn link(&mut self, seg: &PacketSeg) {
-        unsafe { (*self.mp).b_cont = seg.mp };
+        unsafe {
+            // We should not be creating message block continuations to zero
+            // sized blocks. This is not a generally expected thing and has
+            // caused NIC hardware to stop working. Stopping short of a
+            // production panic, but this should fail any tests.
+            #[cfg(any(feature = "std", test))]
```
luqmana, line 1895: Maybe change this to a `debug_assert_ne!`?

rcgoodfellow: Yup, that's definitely nicer. Done.

luqmana (approved):
Wooo! Thanks for staying on this Ry!

PR #363, opened 2023-05-04
xde: release ires after use

### LINKED ISSUE #362: Leaking ire objects
The `ire` objects in the following code needs corresponding `ire_refrele` calls.

https://github.com/oxidecomputer/opte/blob/3f92917953d4e2d2ee2792ce530cace2bd055127/xde/src/xde.rs#L1680-L1685
https://github.com/oxidecomputer/opte/blob/3f92917953d4e2d2ee2792ce530cace2bd055127/xde/src/xde.rs#L1736

We appear to be hitting this here.

- https://github.com/oxidecomputer/stlouis/issues/398

rcgoodfellow:
- Fixes #362 
- Fixes https://github.com/oxidecomputer/stlouis/issues/398

### DIFF
--- a/xde/src/xde.rs
+++ b/xde/src/xde.rs
@@ -1704,6 +1704,7 @@ fn next_hop<'a>(
         }
         let ill = (*ire).ire_ill;
         if ill.is_null() {
+            ip::ire_refrele(ire);
             opte::engine::dbg(format!(
                 "destination ILL is NULL for {:?}",
                 ip6_dst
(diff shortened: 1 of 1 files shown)


rcgoodfellow:
I just tested this on a debug build of stlouis (`stlouis-0-g58898f314e`) and am not seeing the assertion failures reported in stlouis [#398](https://github.com/oxidecomputer/stlouis/issues/398).

citrus-it (approved):

xde/src/xde.rs
```diff
@@ -1758,6 +1760,7 @@ fn next_hop<'a>(
             );
             return (EtherAddr::zero(), EtherAddr::zero(), underlay_port);
         }
+        ip::ire_refrele(gw_ire);
```
luqmana, line 1763: Looking back at this, it doesn't seem like we actually ever use `gw_ire`? Was that lookup just to make sure such an IRE exists?

rcgoodfellow: Yeah, this is just a check to make sure the link-local address of the next hop has an IRE, and we bail early if that is not the case.

PR #365, opened 2023-05-06
fix nce leak, better recount handling, revive mac unicast

### LINKED ISSUE #364: xde: assertion failed: client != NULL in mac_client.c
I came across this kernel assertion failure when testing #363.  This happened when I forgot to add an IPv6 link local address to both underlay devices. When addresses are present on both devices, this does not happen.

Here is a summary of the core dump.

```
> ::status
debugging crash dump vmcore.0 (64-bit) from vandor1
operating system: 5.11 stlouis-0-g58898f314e (i86pc)
build version: heads/stlouis-0-g58898f314e

image uuid: 09e92464-430a-6f01-e0cc-9433978af215
panic message: assertion failed: client != NULL, file: ../../common/io/mac/mac_client.c, line: 2301
dump content: kernel pages only
```

```
> $C
fffffe000fcef230 vpanic()
fffffe000fcef280 0xfffffffffbdcb375()
fffffe000fcef2b0 mac_update_single_active_client+0x82(fffffe0bc8d57fc0)
fffffe000fcef360 mac_client_datapath_setup+0x280(fffffe0bc67c0d18, 0, 0, 0, 1, 0)
fffffe000fcef3f0 mac_client_open+0x3cf(fffffe0bc8d57fc0, fffffe000fcef488, fffffe0bc78b0640, 40)
fffffe000fcef510 _ZN3xde3xde20create_underlay_port17hb19ece629e0868d5E+0x2b2()
fffffe000fcef5d0 init_underlay_ingress_handlers+0x140()
fffffe000fcef6a0 set_xde_underlay+0xe9()
fffffe000fcefc70 xde_ioc_opte_cmd+0x4a4e()
fffffe000fcefcb0 xde_ioctl+0x8b()
fffffe000fcefcf0 cdev_ioctl+0x2b(11f00020000, de777701, fffffc7fffdf0e80, 202003, fffffe0bcf1c5aa8, fffffe000fcefe08)
fffffe000fcefd40 spec_ioctl+0x45(fffffe0bc9431f00, de777701, fffffc7fffdf0e80, 202003, fffffe0bcf1c5aa8, fffffe000fcefe08, 0)
fffffe000fcefdd0 fop_ioctl+0x5b(fffffe0bc9431f00, de777701, fffffc7fffdf0e80, 202003, fffffe0bcf1c5aa8, fffffe000fcefe08, 0)
fffffe000fcefef0 ioctl+0x153(3, de777701, fffffc7fffdf0e80)
fffffe000fceff00 sys_syscall+0x283()
> 
```

The full core dump is available at

```
catacomb:/data/staff/ry/xde/xde-set-underlay-vmdump.0
```

rcgoodfellow:
- Fixes `nce` leaks.
- Fixes `ipstack` leaks.
- Fixes #364

For the last one, a unicast address is re-introduced for the mac clients of the underlay devices. This was removed in the initial `xde` implementation in favor of `MAC_OPEN_FLAGS_NO_UNICAST_ADDR`. However as noted in [this comment](https://github.com/oxidecomputer/opte/issues/364#issuecomment-1536794468) we appear to be violating illumos' expectations for active mac clients by doing this.


rcgoodfellow:
I've taken this for a spin in a local Omicron development environment and things are working as expected. In a Linux instance I Installed a pile of things from `apt`, cloned some repos, curled some things, played some doom.

rcgoodfellow:
Hey @luqmana @bnaecker I'd like to try to get this one in before the next Dogfood test.

tests/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "opte-tests"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+zone = { git = "https://github.com/oxidecomputer/zone" }
+ztest = { git = "https://github.com/oxidecomputer/falcon", branch = "ztest" }
```
bnaecker, line 8: Could we merge this branch and point this to `main` before merging?

tests/tests/loopback.rs
```diff
@@ -0,0 +1,306 @@
...
+        Ok(Self { _zfs: zfs, zone })
+    }
+
+    /// Wait for the network to come up, then set up the overlay network.
+    fn setup(&self, devname: &str, addr: String) -> Result<()> {
+        self.zone.wait_for_network()?;
+        self.zone
+            .zexec(&format!("ipadm create-addr -t -T dhcp {}/test", devname))?;
+        self.zone.zexec(&format!("route add -iface 10.0.0.254 {}", addr))?;
+        self.zone.zexec("route add 10.0.0.0/24 10.0.0.254")?;
```
bnaecker, line 41: Where do these come from? Should we pull them to `const`s?

tests/tests/loopback.rs
```diff
@@ -0,0 +1,306 @@
...
+        })?;
+        Ok(())
+    }
+}
+impl Drop for Xde {
+    /// When this object is dropped, remove the xde kernel module from the
+    /// underlying system.
+    fn drop(&mut self) {
+        let mut cmd = Command::new("pfexec");
+        cmd.args(["rem_drv", "xde"]);
```
bnaecker, line 189: Why are we removing the driver? It just seems a bit heavy-handed, especially if someone is running this test locally.

tests/tests/loopback.rs
```diff
@@ -0,0 +1,306 @@
...
+///     | *--------* |    *-------*      +--| sim1 |--+
+///     | | vopte1 |------| opte1 |         *------*
+///     | *--------* |    *-------*
+///     #============#
+///         zone b
+///
+/// The following system of overlay/underlay routes is set up
+///
+/// 10.0.0.1 -> fd47::/74 via sim1
+/// 10.0.0.2 -> fd74::/74 via sim0
```
bnaecker, line 218: This is kind of a nit, but these addresses are pretty tricky for folks with dyslexia. It'd make it much harder to mess up with something like:

```
10.0.0.6 -> fd66::/66 via sim1
10.0.0.7 -> fd77::/77 via sim0
```

Those are arbitrary, but not "anagrams".

tests/tests/loopback.rs
```diff
@@ -0,0 +1,306 @@
...
+    let sim = SimnetLink::new("sim0", "sim1")?;
+    let ll0 = LinkLocal::new(&sim.end_a, "ll")?;
+    let ll1 = LinkLocal::new(&sim.end_b, "ll")?;
+
+    Xde::set_xde_underlay(&sim.end_a, &sim.end_b)?;
+    // TODO this is a sort of force unset underlay until we have an unset
+    // underlay command. When this object drops it will remove the xde driver.
+    // If we do not do this, xde will hold references to the simnet devices
+    // preventing us from cleaning them up after this test.
+    let _xde = Xde {};
```
bnaecker, line 245: I think this answers my question above about `rem_drv`. Is `modunload` insufficient?

xde/src/mac.rs
```diff
@@ -176,6 +178,28 @@ impl MacClientHandle {
         unsafe { mac_rx_set(self.mch, rx_fn, arg) };
     }
 
+    pub fn add_unicast(
```
bnaecker, line 181: A quick docstring would be helpful, even if it only points to `mac_unicast_add`.

bnaecker (approved):
Seems ok to me, and it's certainly nice to have more tests! A few nits, nothing blocking.

tests/tests/loopback.rs
rcgoodfellow: The logic here is that this test is running with a driver that was specifically added to run this test.

https://github.com/oxidecomputer/opte/blob/72bb20c8f6c740dccd072a21f09aee333eca65e1/.github/buildomat/jobs/test.sh#L34-L37

And we're doing the opposite on the way out. Admittedly, this would be better if the whole cycle was in the rust-based test harness instead of split across bash and rust.

.github/buildomat/jobs/test.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+#:
+#: name = "test"
+#: variety = "basic"
+#: target = "lab-opte-0.22"
```
luqmana, line 5: We're just unloading the existing driver so switch to regular `lab` target? Or even `helios-latest` since it doesn't seem like we need to explicitly running on bare metal. That would free up the lab machines to run other CI.

.github/buildomat/jobs/xde.sh
```diff
@@ -80,4 +84,11 @@ cp $REL_SRC/xde $REL_TGT/
...
 
 cp $REL_LINK_SRC/libxde_link.so $REL_TGT/xde_link.so
-sha256sum $REL_TGT/xde_link.so > $REL_TGT/xde_link.so.sha256
\ No newline at end of file
+sha256sum $REL_TGT/xde_link.so > $REL_TGT/xde_link.so.sha256
+
+popd
+pushd tests
+cargo build --test loopback
+loopback_test=`find target/debug/deps -regex '.*loopback-[0-9a-f]*'`
```
luqmana, line 96: Probably not an issue in CI given it runs in a clean env, but if anyone runs this locally it might pick up an older artifact. Maybe something like this instead:

`cargo build --test loopback --message-format=json | jq -r "select(.profile.test == true) | .filenames[]"`

tests/.cargo/config.toml
```diff
@@ -0,0 +1,3 @@
+# https://github.com/rust-lang/cargo/issues/3946#issuecomment-973132993
+[env]
+CARGO_WORKSPACE_DIR = { value = "..", relative = true }
```
luqmana, line 3: We really should just convert the repo to a workspace at some point.

tests/rustfmt.toml
```diff
@@ -0,0 +1 @@
+../rustfmt.toml
```
luqmana, line 1: Huh, shouldn't it already look in the parent directory by default?

xde/src/xde.rs
```diff
@@ -1720,10 +1793,9 @@ fn next_hop<'a>(
         }
 
         // Step (2): Lookup the IRE for the gateway's link-local
-        // address. This is going to return one of the `fe80::/10`
+        // address. This is geing to return one of the `fe80::/10`
```
luqmana, line 1794: Typo

xde/src/xde.rs
```diff
@@ -1535,6 +1551,60 @@ unsafe extern "C" fn xde_mc_tx(
...
+    unsafe { ip::ire_refrele(ire) }
+}
+
+fn nce_refrele(ire: *mut ip::nce_t) {
+    unsafe { ip::nce_refrele(ire) }
+}
+
+fn netstack_rele(ns: *mut ip::netstack_t) {
+    unsafe { ip::netstack_rele(ns) }
+}
```
luqmana, line 1604: Nice, that's a useful abstraction. Perhaps it should live in a different mod to reuse it in other spots. But that can be a follow up.

tests/Cargo.toml
rcgoodfellow: Done

tests/tests/loopback.rs
rcgoodfellow: See later comment.

tests/tests/loopback.rs
rcgoodfellow: Done

xde/src/mac.rs
```diff
@@ -176,6 +178,28 @@ impl MacClientHandle {
...
+        self: &Arc<Self>,
+        mac: EtherAddr,
+    ) -> Result<MacUnicastHandle, c_int> {
+        let mut diag = mac_diag::MAC_DIAG_NONE;
+        let mut ether = mac.to_bytes();
+        let mut muh = 0 as *mut mac_unicast_handle;
```
luqmana, line 188: nit: `ptr::null_mut()` instead of casting from 0

xde/src/mac.rs
rcgoodfellow: Done

xde/src/xde.rs
```diff
@@ -1097,8 +1112,9 @@ unsafe extern "C" fn xde_detach(
                 // We explicitly drop them in order here to ensure there are no
```
luqmana, line 1110: Update comment to also indicate `MacUnicastHandle` has a ref on `MacClientHandle` as well?

tests/tests/loopback.rs
```diff
@@ -0,0 +1,306 @@
...
+struct OpteZone<'a> {
+    _zfs: &'a Zfs,
+    zone: Zone,
+}
+
+impl<'a> OpteZone<'a> {
+    /// Create a new zone with the given name, underlying zfs instance and set
+    /// of interfaces. In illumos parlance, the interfaces are data link
+    /// devices.
+    fn new(name: &str, zfs: &'a Zfs, ifx: &[&'a str]) -> Result<Self> {
```
luqmana, line 40: Don't think you need the lifetime on the inner `str`.

```suggestion
    fn new(name: &str, zfs: &'a Zfs, ifx: &[&str]) -> Result<Self> {
```

tests/tests/loopback.rs
```diff
@@ -0,0 +1,306 @@
...
+        // If we're running in CI, use the omicron1 zone, because that's whats
+        // there. If this is running locally, use a sparse zone which is much
+        // easier to set up on a regular Helios dev box.
```
luqmana, line 43: Is there explicit setup needed for the `omicron1` brand beyond just installing the `brand/omicron1/tools` pkg?

tests/tests/loopback.rs
luqmana: I believe it's based on the `OptePort`'s `VpcCfg` below. The fact we have to add these routes here manually is also just a bug in the illumos DHCP client (https://github.com/oxidecomputer/stlouis/issues/326 / https://www.illumos.org/issues/11990).

tests/tests/loopback.rs
```diff
@@ -0,0 +1,306 @@
...
+impl OptePort {
+    /// Create a new OPTE port with the provided `name`. The `private_ip` and
+    /// `guest_mac` parameters represent the ip/mac combo in a VM instance. The
+    /// `phys_ip` is the underlay address this port will use for sourcing
+    /// encapsulated traffic.
+    fn new(
+        name: &str,
+        private_ip: &str,
+        guest_mac: &str,
+        phys_ip: &str,
```
luqmana, line 82: Why not take appropriate types instead of strings?

tests/tests/loopback.rs
luqmana: Perhaps an ENV variable to allow overriding this?

tests/tests/loopback.rs
rcgoodfellow: They are the overlay network subnet/gw used in tests. Added constants and comments.

tests/tests/loopback.rs
```diff
@@ -0,0 +1,306 @@
...
+/// that we're not hitting things like debug asserts in the OS.
+#[test]
+fn test_xde_loopback() -> Result<()> {
+    // Create the "underlay loopback". With simnet device pairs, any packet that
+    // goes in one is forwarded to the other. In the topology depicted above,
+    // this means that anything vopte0 sends, will be encapsulated onto the
+    // underlay by opte0, sent to sim0 (more on the routing that makes that
+    // happen later), is forwarded to sim1, is decapsulated by opte1 and then
+    // sent to vopte1.
+    let sim = SimnetLink::new("sim0", "sim1")?;
```
luqmana, line 255: It might be nice to use more unique names so as to not clash with any existing links.

.github/buildomat/jobs/test.sh
```diff
@@ -0,0 +1,56 @@
...
+#: [dependencies.xde]
+#: job = "opte-xde"
+#:
+
+#### >>>>>>>>>>>>>>>>>>>>>>>>>>>> Local Usage >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+####
+#### If you are running this locally, you must run the xde.sh job first to have
+#### the artifacts at the expected spot.
+####
+#### <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
```
luqmana, line 21: Maybe echo this or something short in the `BUILDOMAT_JOB_ID` cond below?

luqmana (commented):
Thanks for fixing those leaks ry and the new test! Looks good overall, just left a couple comments.

.github/buildomat/jobs/test.sh
rcgoodfellow: We can do that once there is a sufficiently up OS image that runs on non-metal.

tests/tests/loopback.rs
rcgoodfellow: Yes and it's involved/complex enough that I threw in the towel and opted to just have support for both `sparse` for local testing and `omicron1` for CI testing.

tests/tests/loopback.rs
rcgoodfellow: This is to capture all the `parse().unwrap()` boilerplate in one function.

.github/buildomat/jobs/test.sh
luqmana: Is `helios-latest` not up to date enough? We can ask josh to push newer bits if needed

rcgoodfellow:
GitHub will not let me respond to this, so here it is out of band.

> Perhaps an ENV variable to allow overriding this?

Without this, the test will fail to clean up. I realize it's pretty drastic. But i think we need more ioctl functionality to properly address this and would like to defer that.

xde/src/xde.rs
rcgoodfellow: Fixed

xde/src/mac.rs
rcgoodfellow: Done

xde/src/xde.rs
rcgoodfellow: Done

tests/tests/loopback.rs
rcgoodfellow: Fixed.

.github/buildomat/jobs/test.sh
rcgoodfellow: Done.

Issue #359, opened 2023-04-28
Kernel panic if ip config is empty


rcgoodfellow:
During a manual setup of OPTE I hit this. It looks like either an external IP or SNAT setup is required now, and my old scripts did not have that. We should probably kick back an error here instead of panicking. This is not urgent as there is an easy workaround, and we likely won't hit this case when Omicron is running the show.

https://github.com/oxidecomputer/opte/blob/2da9dca2131869a462fc22d6e738ec257cf41de0/xde/src/xde.rs#L2016-L2043

```
panic[cpu94]/thread=fffffcfa2f76a3c0: panicked at 'called `Option::unwrap()` on a `None` value', src/xde.rs:2043:17

fffff78815c7ab50 xde:rust_begin_unwind+146 ()
fffff78815c7ab90 xde:_ZN4core9panicking9panic_fmt17hbd42276bcaa2dfd8E+2d ()
fffff78815c7abe0 xde:_ZN4core9panicking5panic17h57af998a49b567d5E+44 ()
fffff78815c7b3e0 xde:new_port+8b5 ()
fffff78815c7b740 xde:create_xde+5ee ()
fffff78815c7bc80 xde:xde_ioc_opte_cmd+5ec1 ()
fffff78815c7bcc0 xde:xde_ioctl+8b ()
fffff78815c7bd00 genunix:cdev_ioctl+3f ()
fffff78815c7bd50 specfs:spec_ioctl+55 ()
fffff78815c7bde0 genunix:fop_ioctl+40 ()
fffff78815c7bf00 genunix:ioctl+144 ()
fffff78815c7bf10 unix:brand_sys_syscall+1fe ()
```

rcgoodfellow:
@leftwo reported an incident on a live rack being run by today where this may have occurred.

```
[8]> $C
fffffffffbe8fe40 kmdb_enter+0xb()
fffffffffbe8fe70 debug_enter+0x75(fffffffffbd5fff8)
fffffffffbe8ff60 panicsys+0x56e(fffffd00f9d6e470, fffff7881fc20b38, fffffffffbe8ff70, 1)
fffff7881fc20b20 vpanic+0x15c()
fffff7881fc20b90 0xfffffffffbca6571()
fffff7881fc20c20 xde`rust_begin_unwind+0x146()
fffff7881fc20c60 xde`_ZN4core9panicking9panic_fmt17hbd42276bcaa2dfd8E+0x2d()
[...]

PR #370, opened 2023-05-17
Avoid panic if there is no external address configured.

bnaecker:
We previously panicked if no external IP address was provided at all. This is because we need a non-zero u32 for creating flow tables, and unwrap when that size is zero. This commit addresses that by returning an error instead. We now allow no external IP configuration, in which case we use a minimum size of 1 for the NAT flow table, though we still create an empty layer. However, if an IP address is provided, allows for zero ports (an odd configuration), we return an error instead of unwrapping.


bnaecker:
Resolves #359

opte-api/src/lib.rs
```diff
@@ -58,7 +58,7 @@ pub use ulp::*;
 ///
 /// We rely on CI and the check-api-version.sh script to verify that
 /// this number is incremented anytime the oxide-api code changes.
-pub const API_VERSION: u64 = 22;
+pub const API_VERSION: u64 = 23;
```
rcgoodfellow, line 61: Is this PR actually changing the API as perceived by ioctl consumers?

xde/src/xde.rs
```diff
@@ -98,7 +97,6 @@ use oxide_vpc::engine::VpcParser;
 //
 // Unwrap: We know all of these are safe to unwrap().
 const FW_FT_LIMIT: Option<NonZeroU32> = NonZeroU32::new(8096);
```
rcgoodfellow, line 99: Instead of having these options that we need to unwrap later, we can just use `NonZeroU32::new_unchecked`. While the name says unchecked, if the value is a constant, it _is_ actually checked at compile time.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9a0bcc48ceff704e8d08476235325842

opte-api/src/lib.rs
bnaecker: It is, yeah. There's a new error variant, `OpteError::InvalidIpCfg`, which is returned if there is an external IP address, but it has zero ports.

xde/src/xde.rs
bnaecker: Nice find, thanks! I'll use that.

rcgoodfellow (approved):

PR #375, opened 2023-06-08
Switch to using a workspace.

luqmana:


luqmana:
The `opte-api` job failure is expected as the check is too coarse-grained. The files were moved but the actual contents weren't changed and thus we don't need an `API_VERSION` bump.

rcgoodfellow (approved):
Thanks @luqmana. This will make things quite a bit easier to work with and maintain.

The only small thing I might suggest is putting a readme in `lib/` and `crate/`, as that distinction was confusing to me until I found in the top level `Cargo.toml` that public API crates are in `lib` and internal crates are in `crates`.

PR #376, opened 2023-06-13
Get to clippy clean and add it to CI

luqmana:


bin/opteadm/Cargo.toml
```diff
@@ -1,8 +1,10 @@
 [package]
 name = "opteadm"
 version = "0.2.0"
-authors = ["Ryan Zezeski <ryan@oxide.computer>"]
-edition = "2021"
+
+edition.workspace = true
```
rcgoodfellow, line 5: Neat. I did not know you could do this with non-dependency things.

rcgoodfellow (approved):
Thanks for getting us to passing clippy @luqmana.

PR #378, opened 2023-06-16
Compute packet body length using packet headers.

### LINKED ISSUE #367: OPTE incorrectly changing payload length on padded short packets
On attempting to access Nexus on the dogfood rack, some folks on Windows noticed they weren't able to and were instead getting HTTP 400 responses or SSL errors. Also managed to reproduce it with a local omicron + SoftNPU deployment. After digging in with @rcgoodfellow we found some weird behaviour in the initial `SYN/SYN-ACK/ACK` sequence that was tripping things up.

Turns out on Windows, TCP timestamps are disabled by default which ends up meaning it may send the final `ACK` in the handshake as a zero-length packet with no TCP options. The IPv4 total length field would then be `20 (IPv4 header) + 20 (TCP header) = 40 bytes`. Yet `snoop` inside the Nexus zone on its OPTE-layered `vNIC` showed a length of 46 bytes and indeed there were an extra 6 bytes at the end! Turns out 6 bytes of zero does not a valid HTTP(S) request make, hence the 400/SSL errors we were seeing.

Ok, so 2 questions now:
1. Why are there an extra 6 zeros?
2. Why is the length 46 bytes instead of 40 bytes?

The answer to (2) is because OPTE changed it. During packet processing we apply whatever transformations (removing outer Geneve encapsulation, rewriting addresses/ports for NAT, etc) and write out a new set of headers:

https://github.com/oxidecomputer/opte/blob/d367866fe4ae47df27f8881dc8d93b86d14f15e0/opte/src/engine/packet.rs#L1458-L1471

But in writing out the IPv4 total length, we calculate it based on the total payload we've received and hence include those 6 extra bytes:

https://github.com/oxidecomputer/opte/blob/d367866fe4ae47df27f8881dc8d93b86d14f15e0/opte/src/engine/packet.rs#L1628

Now in this case, as far as OPTE is concerned, it should respect the original stated length here and that is something we should fix. But coming back to (1): why are there an extra 6 zeros? And where are they getting added? We took some packet captures along the way to try to answer that.

```
Nexus:      192.168.1.21  (External IP) / 172.30.1.5 (OPTE Private IP)
Windows:    192.168.2.110
```

First, as seen from Windows' perspective:
```
19:53:13.113798 IP (tos 0x0, ttl 128, id 34606, offset 0, flags [DF], proto TCP (6), length 40)
    192.168.2.110.50050 > 192.168.1.21.80: Flags [.], cksum 0x84ee (incorrect -> 0xb6b6), ack 1825894618, win 6147, length 0
        0x0000:  4500 0028 872e 4000 8006 eecd c0a8 026e  E..(..@........n
        0x0010:  c0a8 0115 c382 0050 3404 06c1 6cd4 f0da  .......P4...l...
        0x0020:  5010 1803 84ee 0000                      P.......
```

As mentioned before, we're sending out the final `ACK` in the handshake as a zero-length packet with no TCP options. The IPv4 total length is 40 bytes and there are no extra zeros, as expected. Since Windows is running in a VM here, we can also take a look at the packet leaving the Linux host on its physical interface:
```
19:53:13.113850 IP (tos 0x0, ttl 127, id 34606, offset 0, flags [DF], proto TCP (6), length 40)
    192.168.2.110.50050 > 192.168.1.21.80: Flags [.], cksum 0x84ee (incorrect -> 0xb6b6), ack 1825894618, win 6147, length 0
        0x0000:  4500 0028 872e 4000 7f06 efcd c0a8 026e  E..(..@........n
        0x0010:  c0a8 0115 c382 0050 3404 06c1 6cd4 f0da  .......P4...l...
        0x0020:  5010 1803 84ee 0000                      P.......
```
The TTL has gone down (`128 -> 127`) and the IP check sum is different, but everything else is the same as expected. So far so good then. Next step we can take a look at the packet coming in on the physical interface (`e1000g0`) in Helios (before SoftNPU):
```
________________________________
192.168.2.110 -> 192.168.1.21 ETHER Type=0800 (IP), size=60 bytes
192.168.2.110 -> 192.168.1.21 IP  D=192.168.1.21 S=192.168.2.110 LEN=40, ID=34606, TOS=0x0, TTL=127
192.168.2.110 -> 192.168.1.21 TCP D=80 S=50050 Ack=1825894618 Seq=872679105 Len=0 Win=6147

           0: a8e1 de01 701d e0d5 5e2b 9c25 0800 4500    ....p...^+.%..E.
          16: 0028 872e 4000 7f06 efcd c0a8 026e c0a8    .(..@........n..
          32: 0115 c382 0050 3404 06c1 6cd4 f0da 5010    .....P4...l...P.
          48: 1803 b6b6 0000 0000 0000 0000              ............
```
Well well would you look at that: the mysterious zeros have shown up, picked up somewhere along the way from Linux through the local router/switch and onto Helios.
At least the IPv4 length and derived TCP length are correct, 40 and 0, respectively. Which makes sense as we've yet to hit the OPTE codepath mentioned above. We can confirm that at least by also taking a look at the packet inside the Nexus zone:
```
________________________________
192.168.2.110 -> 172.30.1.5   ETHER Type=0800 (IP), size=60 bytes
192.168.2.110 -> 172.30.1.5   IP  D=172.30.1.5 S=192.168.2.110 LEN=46, ID=34606, TOS=0x0, TTL=127
192.168.2.110 -> 172.30.1.5   TCP D=80 S=50050 Ack=1825894618 Seq=872679105 Len=6 Win=6147
192.168.2.110 -> 172.30.1.5   HTTP (body)

           0: a840 25ff a4ed a840 25ff 7777 0800 4500    .@%....@%.ww..E.
          16: 002e 872e 4000 7f06 0462 c0a8 026e ac1e    ....@....b...n..
          32: 0105 c382 0050 3404 06c1 6cd4 f0da 5010    .....P4...l...P.
          48: 1803 cb4a 0000 0000 0000 0000              ...J........
```
OPTE has definitely touched this packet as we can see the 1-1 NAT for the external IP was applied (`192.168.1.21 -> 172.30.1.5`). The IPv4 length also got updated to 46 bytes which means what was originally a zero-length TCP packet with some extra bytes at the end is now treated as a TCP packet with a 6 byte payload.

So, we've confirmed the problematic OPTE behaviour in modifying inner IP length. As for the extra zeros, Ethernet frames have a minimum size of 64 bytes (which includes a 4 byte Frame Check Sequence (FCS) after the payload). The Linux `igb` driver sets the Pad Short Packets flag on the Transmit Control Register for the NIC (`TCTL.PSP`), for which the datasheet says:

```
Pad Short Packets
0b = Do not pad.
1b = Pad short packets.
Padding makes the packet 64 bytes long
```

Doing the math: `64 - (14 [Ethernet Header] + 20 [IPv4 Header] + 20 [TCP Header] + 4 [FCS]) = 6 bytes` of padding and so mystery solved.

luqmana:
Fixes #367.

### DIFF
--- a/lib/opte/src/engine/packet.rs
+++ b/lib/opte/src/engine/packet.rs
@@ -740,12 +740,22 @@ impl Packet<Initialized> {
             pkt_offset,
         );
 
-        let mut body = BodyInfo {
-            pkt_offset,
-            seg_index,
-            seg_offset,
-            len: self.state.len - pkt_offset,
+        let ulp_hdr_len = info.meta.inner.ulp.map(|u| u.hdr_len()).unwrap_or(0);
+        let body_len = match info.meta.inner.ip {
+            // If we have IP and ULP metadata, we can use those to compute
+            // the payload length.
+            // If there's no ULP, just return the L3 payload length.
+            Some(IpMeta::Ip4(ip4)) => {
+                usize::from(ip4.total_len) - ip4.hdr_len() - ulp_hdr_len
+            }
+            Some(IpMeta::Ip6(ip6)) => usize::from(ip6.pay_len) - ulp_hdr_len,
+
+            // If there's no IP metadata, we fallback to considering any
+            // remaining bytes in the packet buffer to be the body.
+            None => self.state.len - pkt_offset,
         };
+        let mut body =
+            BodyInfo { pkt_offset, seg_index, seg_offset, len: body_len };
         let flow = InnerFlowId::from(&info.meta);
 
         // Packet processing logic requires all headers to be in the leading
@@ -3447,7 +3457,6 @@ mod test {
         );
 
         // The computed body length also shouldn't include the padding
-        // XXX: This is not true right now
         assert_eq!(pkt.state.body.len, body.len());
 
         // Pretend some processing happened...
@@ -3463,7 +3472,6 @@ mod test {
         let tcp_hdr = TcpHdr::parse(&mut rdr).unwrap();
 
         // And make sure they don't include the padding bytes
-        // XXX: This is not true right now
         assert_eq!(
             usize::from(ip4_hdr.total_len()),
             usize::from(ip4_hdr.hdr_len()) + tcp_hdr.hdr_len() + body.len()
--- a/lib/opte/src/engine/packet.rs
+++ b/lib/opte/src/engine/packet.rs
@@ -869,7 +869,10 @@ impl Packet<Initialized> {
         self.segs[i].get_writer()
     }
 
-    pub fn add_seg(&mut self, size: usize) -> Result<usize, SegAdjustError> {
+    pub fn add_seg(
+        &mut self,
+        size: usize,
+    ) -> Result<PacketSegWriter, SegAdjustError> {
         let mut seg = PacketSeg::alloc(size);
         seg.expand_end(size)?;
         let len = self.segs.len();
@@ -879,7 +882,8 @@ impl Packet<Initialized> {
         }
         self.segs.push(seg);
         self.state.len += size;
-        Ok(len)
+
+        Ok(self.seg_wtr(len))
     }
 
     /// Wrap the `mblk_t` packet in a [`Packet`], taking ownership of
--- a/lib/opte/src/engine/packet.rs
+++ b/lib/opte/src/engine/packet.rs
@@ -3404,4 +3404,69 @@ mod test {
             assert_eq!(seg.mp, mblk);
         }
     }
+
+    #[test]
+    fn small_packet_with_padding() {
+        const MINIMUM_ETH_FRAME_SZ: usize = 64;
+        const FRAME_CHECK_SEQ_SZ: usize = 4;
+
+        // Start with a test packet that's smaller than the minimum
+        // ethernet frame size (64).
+        let body = [];
+        let mut pkt = tcp_pkt(&body);
+        assert!(pkt.len() < MINIMUM_ETH_FRAME_SZ);
+
+        // Many (most?) NICs will pad out any such frames so that
+        // the total size is 64.
+        let padding_len = MINIMUM_ETH_FRAME_SZ
+            - pkt.len()
+            // Discount the 4 bytes for the Frame Check Sequence (FCS)
+            // which is usually not visible to upstack software.
+            - FRAME_CHECK_SEQ_SZ;
+
+        // Tack on a new segment filled with zero to pad the packet so that
+        // it meets the minimum frame size.
+        // Note that we do NOT update any of the packet headers themselves
+        // as this padding process should be transparent to the upper
+        // layers.
+        let mut padding_seg_wtr = pkt.add_seg(padding_len).unwrap();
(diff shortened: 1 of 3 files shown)


luqmana:
Tested with a local omicron deploy and accessing Nexus from Windows now works:
```
PS C:\Users\Luqman> curl pluto.oxide-preview.com/v1/me
curl : { "request_id": "aed376a2-018b-4880-897a-dabb024b822c", "error_code": "Unauthorized", "message": "credentials
missing or invalid" }
```

lib/opte/src/engine/packet.rs
```diff
@@ -740,12 +740,22 @@ impl Packet<Initialized> {
...
+            // the payload length.
+            // If there's no ULP, just return the L3 payload length.
+            Some(IpMeta::Ip4(ip4)) => {
+                usize::from(ip4.total_len) - ip4.hdr_len() - ulp_hdr_len
+            }
+            Some(IpMeta::Ip6(ip6)) => usize::from(ip6.pay_len) - ulp_hdr_len,
+
+            // If there's no IP metadata, we fallback to considering any
+            // remaining bytes in the packet buffer to be the body.
+            None => self.state.len - pkt_offset,
```
rcgoodfellow, line 755: perhaps this is worth a log entry?

lib/opte/src/engine/packet.rs
```diff
@@ -2867,14 +2878,15 @@ mod test {
...
         let ip4 = Ipv4Meta {
             src: SRC_IP4,
             dst: DST_IP4,
             proto: Protocol::TCP,
             ttl: 64,
             ident: 99,
-            hdr_len: 20,
-            total_len: 40,
+            hdr_len: Ipv4Hdr::BASE_SIZE.try_into().unwrap(),
+            total_len: ip4_total_len.try_into().unwrap(),
```
rcgoodfellow, line 2889: Were we not using this in a substantive way before? Seems crazy that a hardcoded `40` would work at all when the intended semantic of the `Ipv4Meta::hdr_len` field is to be the same as the actual IPv4 header.

rcgoodfellow (commented):
Thanks @luqmana just a few questions/comments.

lib/opte/src/engine/packet.rs
luqmana: This function previously always returned an empty TCP packet, so an IPv4 total length of 40 (20 IPv4 Header + 20 TCP Header) was correct. I changed it to also accept a payload now hence removing the hardcoded values.

lib/opte/src/engine/packet.rs
rcgoodfellow: Of course, if we expect this will be a normal condition this will be a really hot path and logging would be bad here.

luqmana: I didn't add one here specifically because the `NetworkParser` impls will fail in the `parse_{inbound,outbound}` calls above if there's no inner IP. The only exception being for ARP packets.

rcgoodfellow (approved):

PR #379, opened 2023-06-16
Handle IPv4 options

luqmana:
An IPv4 header, although not common, may include some options that increase it beyond the base 20 bytes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,6 @@ default-members = [
   "bin/*",
   "crates/*",
   "lib/*",
-  "xde-tests",
 ]
 resolver = "2"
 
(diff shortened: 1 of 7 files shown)


luqmana:
Pulled this out into it's own draft PR because it seems to break something when testing with omicron. A basic ping works but for TCP, I see the initial SYN packet coming into the nexus zone but no corresponding SYN-ACK. Needs some more investigation.

luqmana:
All but the last commit have been merged in https://github.com/oxidecomputer/opte/pull/378.

The surrounding code has since changed a bit and I don't recall the exact issue mentioned earlier. Gonna close this for now and track as https://github.com/oxidecomputer/opte/issues/571

PR #383, opened 2023-08-21
Filter and detect external traffic using Geneve option tags

### LINKED ISSUE #380: Firewall rules using VPC as target should allow/deny traffic based on private IP, not external IP
While investigating oxidecomputer/omicron#3373, I've dug in a bit more on just the default firewall rule behavior (w/o making any changes to the VPC fw rules). This is the opte firewall setup for a brand new project and default VPC (and no change to the fw rules):
```
BRM44220011 # /opt/oxide/opte/bin/opteadm list-layers -p opte6
NAME         RULES IN   RULES OUT  DEF IN   DEF OUT  FLOWS     
gateway      1          5          deny     deny     0         
firewall     3          0          deny     stateful allow 0         
router       0          2          allow    deny     0         
nat          1          2          allow    allow    0         
overlay      1          1          deny     deny     0  

BRM44220011 # /opt/oxide/opte/bin/opteadm dump-layer firewall -p opte6
Layer firewall
Inbound Flows
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                
TCP    172.20.17.42     64834  172.30.0.5       5201   0        no-op                 

Outbound Flows
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                
TCP    172.30.0.5       5201   172.20.17.42     64834  1        no-op                 

Inbound Rules
ID     PRI    HITS   PREDICATES                             ACTION            
2      65534  1      inner.ip.proto=TCP                     "Stateful Allow"
                     inner.ulp.dst=22                      

1      65534  1      meta: vni=7524168                      "Stateful Allow"
0      65534  0      inner.ip.proto=ICMP                    "Stateful Allow"
DEF    --     0      --                                     "deny"

Outbound Rules
ID     PRI    HITS   PREDICATES                             ACTION            
DEF    --     47     --                                     "stateful allow
```

I created an instance in it and ran an iperf3 server on it.
```
ubuntu@default-fwrules:~$ iperf3 -s -D
ubuntu@default-fwrules:~$ exit
logout
Connection to 172.20.26.45 closed.
pisces-2:docs angela$ nc -vz 172.20.26.45 5201
Connection to 172.20.26.45 port 5201 [tcp/targus-getdata1] succeeded!
```
Based on the default rules, the only port accessible on tcp should be 22. So it'd look like the deny-all default is not taking effect.

Next, I tried disabling the `allow-ssh` rule and saw that the change was reflected in the opte port: 
```
BRM44220011 # /opt/oxide/opte/bin/opteadm dump-layer firewall -p opte6
Layer firewall
Inbound Flows
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                
TCP    172.20.17.42     64897  172.30.0.5       22     0        no-op                 

Outbound Flows
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                
TCP    172.30.0.5       22     172.20.17.42     64897  1        no-op                 

Inbound Rules
ID     PRI    HITS   PREDICATES                             ACTION            
4      65534  1      meta: vni=7524168                      "Stateful Allow"
3      65534  0      inner.ip.proto=ICMP                    "Stateful Allow"
DEF    --     0      --                                     "deny"

Outbound Rules
ID     PRI    HITS   PREDICATES                             ACTION            
DEF    --     50     --                                     "stateful allow"
```

Afterwards, I was still able to SSH to the vm. This further confirms that there is no firewall at all.

FelixMcFelix:
This PR uses the presence of a min-sized Geneve TLV to detect packets on a given VNI which are originally external to the rack. To correctly filter such traffic at the firewall, the semantics of the `Address::Vni` host filter have changed to mean '*internal traffic* on this Vni'. Accordingly, I've added simple option parsing/emission.

This will require changes to Dendrite to correctly emit the new option on any SNAT transformations on ingress to the rack, as well as an OANA registration for option type 0.

Aims to fix #380 (in tandem with oxidecomputer/dendrite#606).

### DIFF
--- a/lib/opte/src/engine/geneve.rs
+++ b/lib/opte/src/engine/geneve.rs
@@ -27,13 +27,22 @@ pub const GENEVE_VSN: u8 = 0;
 pub const GENEVE_VER_MASK: u8 = 0xC0;
 pub const GENEVE_VER_SHIFT: u8 = 6;
 pub const GENEVE_OPT_LEN_MASK: u8 = 0x3F;
+pub const GENEVE_OPT_LEN_SCALE_SHIFT: u8 = 2;
 pub const GENEVE_PORT: u16 = 6081;
+
+pub const GENEVE_OPT_CRIT_SHIFT: u8 = 7;
+pub const GENEVE_OPT_TYPE_MASK: u8 = (1 << GENEVE_OPT_CRIT_SHIFT) - 1;
+pub const GENEVE_OPT_RESERVED_SHIFT: u8 = 5;
+pub const GENEVE_OPT_RESERVED_MASK: u8 = (1 << GENEVE_OPT_RESERVED_SHIFT) - 1;
+pub const GENEVE_OPT_CLASS_OXIDE: u16 = 0x0129;
 
 #[derive(Clone, Copy, Debug, Default, Eq, Ord, PartialEq, PartialOrd)]
 pub struct GeneveMeta {
     pub entropy: u16,
     pub vni: Vni,
     pub len: u16,
+
+    pub oxide_external_pkt: bool,
 }
 
 #[derive(
@@ -92,16 +101,30 @@ impl GeneveMeta {
 
 impl<'a> From<&GeneveHdr<'a>> for GeneveMeta {
     fn from(geneve: &GeneveHdr<'a>) -> Self {
-        Self {
+        let mut out = Self {
             vni: geneve.vni(),
             entropy: geneve.entropy(),
             len: geneve.len() as u16,
+            ..Default::default()
+        };
+
+        if let Some(ref opts) = geneve.opts {
+            // XXX: Prevent duplication by making Meta generation fallible
+            //      in same way as Parsing?
+            // Unwrap safety: Invalid options will have been caught in
+            // GeneveHdr::parse.
+            GeneveOption::parse_all(&opts, Some(&mut out)).unwrap();
         }
+
+        out
     }
 }
 
 pub struct GeneveHdr<'a> {
+    /// Main body of the Geneve Header.
     bytes: LayoutVerified<&'a mut [u8], GeneveHdrRaw>,
+    /// Byte slice occupied by Geneve options.
+    opts: Option<&'a mut [u8]>,
 }
 
 impl<'a> GeneveHdr<'a> {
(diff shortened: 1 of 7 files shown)


lib/oxide-vpc/src/engine/overlay.rs
```diff
@@ -379,10 +379,16 @@ impl StaticAction for DecapAction {
...
         match &pkt_meta.outer.encap {
             Some(EncapMeta::Geneve(geneve)) => {
-                action_meta.insert(
-                    ACTION_META_VNI.to_string(),
-                    geneve.vni.to_string(),
-                );
+                // XXX: This is now slightly overloaded to mean "packet
+                //      originated from the VPC which has this VNI".
+                //      Unsure if we want to have sled agent add a
+                //      VNI && (not external) predicate.
```
luqmana, line 385: I think that's fine, the user facing firewall API doesn't even deal in VNI's directly, but rather VPC originated traffic and sled-agent translates it to the corresponding VNI.

If we did want to separate out things a bit, perhaps instead we could add a new meta tag to indicate internal/external and then update the [predicate generated for `Address::Vni`](https://github.com/oxidecomputer/opte/blob/98d33125413f01722947e322f82caf9d22209434/lib/oxide-vpc/src/engine/firewall.rs#L211-L214):

```rust
            (_, Address::Vni(vni)) => Some(Predicate::And(
                Predicate::Meta(ACTION_META_EXTERNAL.to_string(), "false".to_string()),
                Predicate::Meta(ACTION_META_VNI.to_string(), vni.to_string()),
            )),
```
(`Predicate::And` doesn't exist but i imagine would be straight forward)

Otherwise, I would at the very least add a comment here explaining why we only conditionally add `ACTION_META_VNI`.

FelixMcFelix: Thanks, Luqman -- I'll explain this via comment for now, though I may have one or two tweaks after testing with sidecar-lite!

lib/opte/src/engine/geneve.rs
```diff
@@ -245,17 +303,195 @@ impl From<&GeneveMeta> for GeneveHdrRaw {
...
+}
+
+/// Geneve options defined by Oxide, [`GENEVE_OPT_CLASS_OXIDE`].
+#[non_exhaustive]
+pub enum OxideOption {
+    /// A tag indicating that this packet originated from outside the VPC.
+    ///
+    /// Option Type `0`. Currently includes no body.
+    External,
+}
```
luqmana, line 404: Need to update https://github.com/oxidecomputer/oana#geneve-options with the new option type.

luqmana (approved):
Thanks Kyle. Pending testing end-to-end, looks good.

lib/opte/src/engine/geneve.rs
FelixMcFelix: Okay, I have a [PR opened](https://github.com/oxidecomputer/oana/pull/20).

Issue #317, opened 2023-01-10
Need to track packets dropped by firewall layer


bnaecker:
We currently track a bunch of stats about ports, including packet drops either in- or outbound. (See [here](https://github.com/oxidecomputer/opte/blob/f501445f5a6c275c79f08a876fff6a861df31d46/opte/src/engine/port.rs#L523) for inbound drops caused by a layer rule.) We need to distinguish at least the packets dropped due to a firewall rule. It's not clear how to best to that, whether to split that out alone as a separate stat or to split out drops by layer name.

PR #384, opened 2023-09-01
Track packets dropped/denied per layer via `LayerStats`

FelixMcFelix:
This PR adds `{in, out}_deny` fields to the `LayerStats` kstat provider, which allow us to track, e.g., firewall drops via "opte:opte0_firewall". These increment on any rule-based or default `Deny`.

This provides bare-minimum stats needed for #317.


bnaecker (approved):
Looks great, thanks for taking this!

PR #387, opened 2023-09-04
Migrate opteadm to Clap, emit package version on -V

### LINKED ISSUE #256: Move from `structopt` to `clap`
With version 3 of `clap`, the `structopt` module we're using in `opteadm` is deprecated. All its functionality is subsumed by `clap`. We should move `opteadm` to it.

### LINKED ISSUE #345: Better version output
Often times it's useful to figure out which (API) version of the driver/tools is installed. Especially when the usual packaging tools aren't available so you can't do `pkg info -l driver/network/opte` (like on the gimlet ramdisks). `opteadm` is included and has a version flag but it's output is less than useful:
```
gimlet-sn21 # /opt/oxide/opte/bin/opteadm  -V
opteadm 0.2.0
```

It's been on `0.2.0` for quite a while 😅 At the very least, the packaged version of `opteadm` should match the package version (e.g. `0.21.160`).

FelixMcFelix:
As indicated in #256, modern Clap now supports all features of structopt
and is actively maintained. This PR implements this (simple) move. Additionally,
opteadm should now return a version string which matches the intended
package version if the OPTE/XDE it is built alongside, using the git
commit count and API version.

Clap v4 has an MSRV of 1.70.0, so I've bumped the central toolchain to
1.72 to accommodate this with room. None of the other deps in the
lockfile have been `cargo update`d.

Should close #256 and close #345.


bin/opteadm/build.rs
```diff
@@ -0,0 +1,31 @@
+use std::error::Error;
+use std::process::Command;
+
+fn main() -> Result<(), Box<dyn Error>> {
```
bnaecker, line 4: No real need to change this, but I find `anyhow` excellent for these kinds of one-off, mostly string-based errors in small applications.

bin/opteadm/build.rs
```diff
@@ -0,0 +1,31 @@
...
+    println!("cargo:rerun-if-changed=../../.git");
+
+    let commit_count_out =
+        Command::new("git").args(["rev-list", "--count", "HEAD"]).output()?;
```
bnaecker, line 8: Is there any issue with pointing to HEAD rather than, say `main`? I think we always build CI artifacts from `main`, but just checking.

bin/opteadm/src/bin/opteadm.rs
```diff
@@ -7,7 +7,10 @@
 use std::io;
 use std::str::FromStr;
 
-use structopt::StructOpt;
+// use structopt::StructOpt;
```
bnaecker, line 10: Dead code?

bin/opteadm/build.rs
```diff
@@ -0,0 +1,31 @@
+use std::error::Error;
+use std::process::Command;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    println!("cargo:rerun-if-changed=../../.git");
+
+    let commit_count_out =
```
bnaecker, line 7: Perhaps we could consider also adding the commit SHA from `HEAD` itself? I've found that helpful in many situations. The `vergen` and `built` crates provide easy access to a number of pieces of build information, such as timestamps, which may also be helpful.

bin/opteadm/src/bin/opteadm.rs
```diff
@@ -240,6 +237,26 @@ impl From<Filters> for FirewallFilters {
...
+    #[arg(long)]
+    snat_start: Option<u16>,
+
+    /// The ending L4 port used for source NAT for the guest.
+    #[arg(long)]
+    snat_end: Option<u16>,
+}
+
+fn opte_pkg_version() -> String {
+    format!("0.{API_VERSION}.{COMMIT_COUNT}")
```
bnaecker, line 257: Maybe pull the 0 into a `pub const MAJOR_VERSION: u64 = 0;`?

bnaecker (approved):
Looks good to me, thanks for doing this! I've a few suggestions, all at your discretion. Thanks!

bin/opteadm/src/bin/opteadm.rs
FelixMcFelix: Good catch, thanks.

bin/opteadm/build.rs
FelixMcFelix: `master`/`main` would probably be more correct here -- at least for preventing odd version numbers in development branches -- but HEAD is what is used by `pkg/build.sh`. We could change it, we'd just need to do so in both spots.

bin/opteadm/src/bin/opteadm.rs
```diff
@@ -123,74 +129,65 @@ enum Command {
...
-        /// The ending L4 port used for source NAT for the guest.
-        #[structopt(long)]
-        snat_end: Option<u16>,
+        #[command(flatten)]
+        snat: SnatConfig,
 
         /// A list of domain names provided to the guest, used when resolving
         /// hostnames.
-        #[structopt(long, parse(try_from_str))]
+        #[arg(long)]
```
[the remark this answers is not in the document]
FelixMcFelix: I'm maybe leaning towards `value_delimiter=','`, then: a bit less verbose than default, and no need to disambiguate from positional args with a `--`.

bin/opteadm/build.rs
```diff
@@ -0,0 +1,31 @@
...
+        Command::new("git").args(["rev-list", "--count", "HEAD"]).output()?;
+
+    if commit_count_out.status.success() {
+        let commit_count: u64 =
+            std::str::from_utf8(&commit_count_out.stdout)?.trim().parse()?;
+
+        std::fs::write(
+            "src/gen.rs",
```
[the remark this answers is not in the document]
FelixMcFelix: Thanks, I was unaware of this!

PR #393, opened 2023-09-08
Filter all uncaught {In, Out} NDP traffic

### LINKED ISSUE #356: ICMPv6 layer should drop more NDP traffic
See [this comment](https://github.com/oxidecomputer/omicron/pull/2857#issuecomment-1517029866) and the surrounding thread for context. We're working on getting OPTE underneath Oxide services like Nexus, and providing those services external IP addresses. During development @jmpesp noted that NDP NAs make it from the OPTE port all the way out past boundary services. The fact that BS decaps the packet is sort of fine -- that's it's job. But OPTE should really be dropping that right away.

Our model for OPTE-based networking (either a true guest VM or for Oxide services) is that each port is on a segment with exactly one peer: OPTE itself. There should be no other neighbors, and so both ARP and NDP need to be tightly controlled. We're currently hairpinning NDP RSs back to the guest, advertising OPTE itself as the router. We're also hairpinning NSs for the OPTE address as well, which makes sense. But we need to drop all other NDP traffic. No NSs or NAs should make it out past OPTE, and OPTE should drop all inbound NDP of any kind. The OPTE port is an island.

FelixMcFelix:
Adds an additional gateway rule in each direction when v6-configured that will filter unexpected NDP traffic. I was unable to locally replicate the exact behaviour reported in omicron, but putting together some test cases similar enough to the original report confirmed the packets were uncaught (and should now be).

Also adds a `TypeMatch` so that we can express matches on collections with a little more ease. Predicates could also take another look, but that's out of scope for this bugfix PR.

Closes #356.

### DIFF
--- a/lib/opte/src/engine/dhcpv6/protocol.rs
+++ b/lib/opte/src/engine/dhcpv6/protocol.rs
@@ -128,6 +128,12 @@ impl From<MessageType> for u8 {
     }
 }
 
+impl PartialOrd for MessageType {
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        u8::from(*self).partial_cmp(&u8::from(*other))
+    }
+}
+
 /// A DHCPv6 message.
 ///
 /// All DHCPv6 transactions occur through this type. Clients send messages,
--- a/lib/opte/src/engine/predicate.rs
+++ b/lib/opte/src/engine/predicate.rs
@@ -677,3 +677,39 @@ impl DataPredicate {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use smoltcp::wire::DhcpMessageType as SmolDhcpType;
+    use smoltcp::wire::Icmpv4Message;
+    use smoltcp::wire::Icmpv6Message;
+
+    // Some 'enum with unknown' ways of encoding message types
+    // can have some unexpected behaviour with PartialOrd -- we
+    // need to sort on the underlying representation for range
+    // matches to be sensible.
+    #[test]
+    fn data_predicate_ranges_handle_unknown() {
+        let dhcp_range: Match<DhcpMessageType> = (SmolDhcpType::Discover.into()
+            ..=SmolDhcpType::Decline.into())
+            .into();
+        let icmp_range: Match<IcmpMessageType> =
+            (Icmpv4Message::EchoReply.into()..=Icmpv4Message::Redirect.into())
+                .into();
+        let dhcp6_range: Match<Dhcpv6MessageType> =
+            (Dhcpv6MessageType::Renew..=Dhcpv6MessageType::Reply).into();
+
+        let icmp6_range: Match<Icmpv6MessageType> =
+            (Icmpv6Message::RouterSolicit.into()
+                ..=Icmpv6Message::Redirect.into())
+                .into();
+
+        // The `Unknown` cases here are artificial (i.e., the opcode is understood)
+        // in case the underlying repr adds support for a test opcode.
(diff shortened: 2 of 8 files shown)


lib/opte/src/engine/predicate.rs
```diff
@@ -505,12 +506,53 @@ impl Predicate {
...
+        }
+    }
+}
+
+impl<T: Display> Display for Match<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Exact(target) => write!(f, "={target}"),
+            Self::Range(range) => {
+                write!(f, "∈({}..={})", range.start(), range.end())
```
luqmana, line 532: Debug formatter on `RangeInclusive` will also give you the `{}..={}` output.

lib/oxide-vpc/src/engine/gateway/icmpv6.rs
```diff
@@ -89,17 +86,32 @@ pub fn setup(
...
-    for i in 0..n_rule_actions {
+    hairpins.into_iter().enumerate().for_each(|(i, action)| {
         let priority = u16::try_from(i + 1).unwrap();
-        let rule = Rule::new(priority, rule_actions.remove(0));
+        let rule = Rule::new(priority, action);
         layer.add_rule(Direction::Out, rule.finalize());
-    }
+    });
+
+    // Filter any uncaught in/out-bound NDP traffic.
```
luqmana, line 101: I suppose it's not a bad idea to have the rules, but isn't the gateway layer deny by default?

lib/opte/src/engine/dhcpv6/protocol.rs
```diff
@@ -128,6 +128,12 @@ impl From<MessageType> for u8 {
     }
 }
 
+impl PartialOrd for MessageType {
```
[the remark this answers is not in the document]
FelixMcFelix: The reason I didn't derive it here is that all the other DHCP/ICMP types are nested wrappers around a uint type, while Dhcpv6's MessageType isn't, so you'd end up with oddities like `MessageType::Other(6) > MessageType::Reply` if you were to just derive it.

lib/oxide-vpc/src/engine/gateway/icmpv6.rs
FelixMcFelix: The layer is deny by default, but due to the [last in/outbound rules applied here](https://github.com/oxidecomputer/opte/blob/master/lib/oxide-vpc/src/engine/gateway/mod.rs#L211-L234) this only denies IP-spoofed or misaddressed packets. The NDP packets in question have the correct unicast sender/recipient, so would not be denied.

lib/opte/src/engine/predicate.rs
FelixMcFelix: Would that not provide `{:?}..={:?}`? Although it would probably be less space consuming than having long display formats like "router solicitation" constantly appearing in the rule set.

lib/opte/src/engine/dhcpv6/protocol.rs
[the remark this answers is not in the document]
FelixMcFelix: Digging in a little deeper it seems like smoltcp's `enum_with_unknown!` macro ends up just deriving `PartialOrd`, leading to the same issue. I'll probably manually implement it for the other message types to get consistent range behaviour.

lib/opte/src/engine/predicate.rs
luqmana: ah right, this is `Display` impl not `Debug`.

luqmana (approved):
Thanks!

PR #394, opened 2023-09-13
Offer per-VpcCfg DNS servers & hostname via DHCP

### LINKED ISSUE #382: Allow configuring hostname as part of `VpcCfg` to return with DHCP response.
A DHCP server can send along an option[^1] indicating the host name for a client as part of the DHCP response.

Per the RFC, the host name doesn't need to be fully qualified but relatedly, we might want to also send a domain name[^2]?

[^1]: [RFC 2132 § 3.14 Host Name Option](https://www.rfc-editor.org/rfc/rfc2132.html#section-3.14)
[^2]: [RFC 2132 § 3.17 Domain Name](https://www.rfc-editor.org/rfc/rfc2132.html#section-3.17)

### LINKED ISSUE #390: Hard coded DHCP DNS options
This line is causing problems and needs to be parameterized with the rack configured dns addresses.

https://github.com/oxidecomputer/opte/blob/e5423d80dff36e516cefba95ae35865173902bf4/lib/oxide-vpc/src/engine/gateway/dhcp.rs#L84

FelixMcFelix:
Replumbs DHCP-related state out of `VpcCfg` and into a separate map in `XdeState`, allowing us to dynamically update DNS servers, hostnames, domain search lists etc. on an open port via ioctl. Currently this defaults to the old {8.8.8.8, v6 equivalents}, and is then overridden -- we can include this in the `CreateXdeReq` if needed.

This also adds support for carrying host names and the local domain, and so OPTE will respond to hostname/domain/FQDN option requests in v4/v6. Most VMs appear to have the hostname preset during boot; it's up to the guest's configuration as to whether it will *use* these fields.

Closes #390 and closes #382.


lib/oxide-vpc/tests/integration_tests.rs
```diff
@@ -2248,6 +2247,17 @@ fn verify_dhcpv6_essentials<'a>(
         client_id,
         reply.find_option(dhcpv6::options::Code::ClientId).unwrap()
     );
+
+    // Assert FQDN is correctly constructed.
```
rcgoodfellow, line 2251: A high-level comment on integration test coverage. It looks like we have DHCP6 coverage but not DHCP4?

iliana:
> we can include this in the `CreateXdeReq` if needed

I think being able to pass a `DhcpCfg` into `CreateXdeReq` would be ideal. (It could be an optional parameter in `VpcCfg`?)

iliana:
Some curious results after implementing this in the control plane (oxidecomputer/omicron#4265):

- On Debian 12 (bookworm), /etc/resolv.conf gets a `search .` entry. This is somewhat unexpected because we are specifying no search domains, and my initial glance at [the relevant-seeming OPTE code](https://github.com/oxidecomputer/opte/blob/d338e68a9df1c0c29c9fc469c86ff3224a678269/lib/opte/src/engine/dhcp.rs#L559-L564) indicates that we aren't encoding these options at all if the list of search domains is empty.
- While trying to debug that issue, I found that `networkctl renew enp0s9` on a Debian 12 instance did not result in a DHCP response that I could capture! The DHCP request is being sent to 172.30.0.1 instead of a broadcast address, perhaps this is relevant?

iliana:
> On Debian 12 (bookworm), /etc/resolv.conf gets a search . entry.

Looks like systemd-resolved just Does This; not a weirdness with the DHCP response.

xde/src/xde.rs
```diff
@@ -2245,6 +2272,58 @@ fn dump_tcp_flows_hdlr(
...
+    let dhcp_cfg = {
+        let map_locked = state.dhcp_map.read();
+        if let Some(dhcp) = map_locked.get(&dev.vni) {
+            dhcp.clone()
+        } else {
+            return Err(OpteError::PortNotFound(req.port_name));
+        }
+    };
+    let mut dhcp_locked = dhcp_cfg.write();
+    *dhcp_locked = req.data;
```
luqmana, line 2295: I'm not sure `{Dump,Set}DhcpParams` work as expected if `dhcp_map` is keyed on VNI's:

1. Add port 1 with VNI 100. `dhcp_map[100]` will be the Google/Cloudflare `DhcpCfg` and that's also what port 1 gets configured with.
2. SetDhcpParams on port 1 with a new `DhcpCfg`. This grabs `dhcp_map[100]` and updates the `Arc<KwRwLock<DhcpCfg>>` also held by port 1. All good so far.

But consider if a new port gets added in between on the same VPC and thus same VNI. `dhcp_map[100]` gets overridden with a new base `DhcpCfg` that's passed to port 2. Now when the SetDhcpParams request happens, even though we wanted port 1, we're actually updating the state associated with port 2.

luqmana:
```console
# create port 1
$ pfexec ./target/debug/opteadm create-xde vm_foo0 [...]

$ pfexec ./target/debug/opteadm list-ports
LINK                             MAC ADDRESS              IPv4 ADDRESS     EXTERNAL IPv4    IPv6 ADDRESS                             EXTERNAL IPv6                            STATE
vm_foo0                          A8:40:25:FF:00:00        10.0.0.10        192.168.1.20     None                                     None                                     running

$ pfexec ./target/debug/opteadm dump-dhcp-params -p vm_foo0
DHCP Configuration
Hostname: <unset>
Domain: <unset>
Search Domains: <unset>
IPv4 DNS Servers:
- 8.8.8.8
- 1.1.1.1
IPv6 DNS Servers:
- 2606:4700:4700::1111
- 2606:4700:4700::1001
- 2001:4860:4860::8888
- 2001:4860:4860::8844

$ pfexec ./target/debug/opteadm create-xde vm_bar0 [...]

$ pfexec ./target/debug/opteadm list-ports
LINK                             MAC ADDRESS              IPv4 ADDRESS     EXTERNAL IPv4    IPv6 ADDRESS                             EXTERNAL IPv6                            STATE
vm_foo0                          A8:40:25:FF:00:00        10.0.0.10        192.168.1.20     None                                     None                                     running
vm_bar0                          A8:40:25:FF:00:01        10.0.0.11        192.168.1.21     None                                     None                                     running

$ pfexec ./target/debug/opteadm dump-dhcp-params -p vm_bar0
DHCP Configuration
Hostname: <unset>
Domain: <unset>
Search Domains: <unset>
IPv4 DNS Servers:
- 8.8.8.8
- 1.1.1.1
IPv6 DNS Servers:
- 2606:4700:4700::1111
- 2606:4700:4700::1001
- 2001:4860:4860::8888
- 2001:4860:4860::8844

# try updating dhcp config for `vm_foo0` port
$ pfexec ./target/debug/opteadm set-dhcp-params -p vm_foo0 --hostname foo

$ pfexec ./target/debug/opteadm dump-dhcp-params -p vm_foo0
DHCP Configuration
Hostname: foo
Domain: <unset>
Search Domains: <unset>
IPv4 DNS Servers: <unset>
IPv6 DNS Servers: <unset>

$ pfexec ./target/debug/opteadm dump-dhcp-params -p vm_bar0
DHCP Configuration
Hostname: foo
Domain: <unset>
Search Domains: <unset>
IPv4 DNS Servers: <unset>
IPv6 DNS Servers: <unset>
```

bin/opteadm/src/bin/opteadm.rs
```diff
@@ -283,6 +299,51 @@ impl TryFrom<SnatConfig> for SNat6Cfg {
...
+            .dns_servers
+            .into_iter()
+            .filter_map(|ip| match ip {
+                IpAddr::Ip4(ip) => {
+                    dns4_servers.push(ip);
+                    None
+                }
+                IpAddr::Ip6(ip) => Some(ip),
+            })
+            .collect();
```
luqmana, line 335:
```suggestion
        let (dns4_servers, dns6_servers) = value
            .dns_servers
            .into_iter()
            .partition(|ip| matches!(ip, IpAddr::Ip4(_)))
            .collect();
```

crates/opte-api/src/ip.rs
```diff
@@ -94,68 +94,71 @@ impl Display for IcmpEchoReply {
...
+    /// IPv4 external DNS servers provided to a guest.
+    pub dns4_servers: Vec<Ipv4Addr>,
+
+    /// IPv6 external DNS servers provided to a guest.
+    pub dns6_servers: Vec<Ipv6Addr>,
+}
+
+impl DhcpCfg {
+    /// Provide DNS servers which allow basic name resolution via CloudFlare/Google.
+    pub fn base_reachable() -> DhcpCfg {
```
luqmana, line 121: Mentioned in another comment, but I'd remove this from opte proper and leave it to consumers to decide on a default.

crates/opte-api/src/ip.rs
```diff
@@ -1394,4 +1397,35 @@ mod test {
...
+
+        let mut space = vec![];
+
+        no_host.push_fqdn(&mut space);
+        assert!(space.is_empty());
+
+        only_host.push_fqdn(&mut space);
+        assert_eq!(&space, "\x05mybox\x00".as_bytes());
+
+        space.truncate(0);
```
luqmana, line 1423: nit: clear instead of truncate

lib/opte/src/engine/dhcpv6/protocol.rs
```diff
@@ -363,10 +370,35 @@ fn generate_reply_options<'a>(
...
+    }
+
+    if msg.has_option(OptionCode::Fqdn)
+        || msg.has_option_request_with(OptionCode::Fqdn)
+    {
+        // XXX: We should verify customer flow here -- correct
+        //      for internal DNS, maybe not external?
+        // Flags: we are (O)verriding client preference, and (S)erver is
+        // responsible for installing DNS records.
+        //                   xxxx_xNOS
```
luqmana, line 393: Mention what the N flag is and why we're clearing it?

lib/opte/src/engine/ioctl.rs
```diff
@@ -191,3 +191,10 @@ pub fn dump_tcp_flows(
 ) -> Result<DumpTcpFlowsResp, OpteError> {
     port.dump_tcp_flows()
 }
+
+pub fn dump_dhcp_params(
+    port: &Port<impl crate::engine::NetworkImpl>,
+    _req: &DumpTcpFlowsReq,
+) -> Result<DumpTcpFlowsResp, OpteError> {
+    port.dump_tcp_flows()
+}
```
luqmana, line 200: Copy paste? Doesn't seem like this is called anywhere

lib/oxide-vpc/tests/integration_tests.rs
luqmana: AFAIK DHCPv4 doesn't have a single FQDN option like DHCPv6 but instead separate Hostname and Domain options. But yea some DHCPv4 coverage would be welcome.

luqmana (changes requested):
Ok, see review comments but I don't think we should merge this as-is. The SetDhcpParams/DumpDhcpParams commands do not function as expected for ports within the same VPC.

For the bits necessary for the omicron side of things, I'd say making we should make the Dhcp config part of the CreateXde request directly instead of as a separate step. That does mean doing without the ability to update the DHCP config for a running port but I don't think there's omicron support for that yet anyways, so I wouldn't say it's a blocker.

bin/opteadm/src/bin/opteadm.rs
iliana: This is not quite the same; this results in `(Vec<IpAddr>, Vec<IpAddr>)`, but we need `(Vec<Ipv4Addr>, Vec<Ipv6Addr>)`. (I realized this while trying to implement the same thing in Omicron.)

lib/opte/src/engine/dhcpv6/protocol.rs
iliana: [RFC 4704 § 4.1] covers the flags field:

```
   The format of the Flags field is:

        0 1 2 3 4 5 6 7
       +-+-+-+-+-+-+-+-+
       |  MBZ    |N|O|S|
       +-+-+-+-+-+-+-+-+


   The "S" bit indicates whether the server SHOULD or SHOULD NOT perform
   the AAAA RR (FQDN-to-address) DNS updates.  A client sets the bit to
   0 to indicate that the server SHOULD NOT perform the updates and 1 to
   indicate that the server SHOULD perform the updates.  The state of
   the bit in the reply from the server indicates the action to be taken
   by the server; if it is 1, the server has taken responsibility for
   AAAA RR updates for the FQDN.

   The "O" bit indicates whether the server has overridden the client's
   preference for the "S" bit.  A client MUST set this bit to 0.  A
   server MUST set this bit to 1 if the "S" bit in its reply to the
   client does not match the "S" bit received from the client.

   The "N" bit indicates whether the server SHOULD NOT perform any DNS
   updates.  A client sets this bit to 0 to request that the server
   SHOULD perform updates (the PTR RR and possibly the AAAA RR based on
   the "S" bit) or to 1 to request that the server SHOULD NOT perform
   any DNS updates.  A server sets the "N" bit to indicate whether the
   server SHALL (0) or SHALL NOT (1) perform DNS updates.  If the "N"
   bit is 1, the "S" bit MUST be 0.

   The remaining bits in the Flags field are reserved for future
   assignment.  DHCPv6 clients and servers that send the Client FQDN
   option MUST clear the MBZ bits, and they MUST ignore these bits.
```

So I guess in the response, `S` is "whether the server is installing AAAA resource records", and `N` is "whether the server is installing _any_ resource record changes (AAAA, PTR, or whatever shows up in future RFCs)". I'll add a note about what N is.

iliana:
Per a suggestion from @luqmana, in the interest of getting this landed sooner to resolve a customer pain point, 4b5bc48 removes live configuration of DHCP options by removing `XdeState::dhcp_map` and the `SetDhcpParams` and `DumpDhcpParams` commands. We will likely want these back at some point, but as part of figuring out the design for DHCP options / instance-facing DNS across the whole product we will want to evaluate the granularity of certain DHCP settings. `Vni` is definitely an incorrect key for fields like `hostname`, but that level of granularity is probably correct for the remaining fields.

bin/opteadm/src/bin/opteadm.rs
luqmana: ah, d'oh. right

lib/opte/src/engine/dhcpv6/protocol.rs
```diff
@@ -321,9 +323,8 @@ fn generate_reply_options<'a>(
 
     // If requested, provide the list of DNS servers.
     if msg.has_option_request_with(OptionCode::DnsServers) {
-        let opt = Dhcpv6Option::DnsServers(IpList::from(
-            action.dns_servers.as_slice(),
-        ));
+        let ip_list = IpList(Cow::Owned(action.dhcp_cfg.dns6_servers.clone()));
```
luqmana, line 326: Nit: can probably just use `Cow::Borrowed` here since the `dhcp_cfg` isn't behind a RwLock anymore.

lib/oxide-vpc/src/api.rs
```diff
@@ -570,6 +567,24 @@ pub struct RemFwRuleReq {
...
+pub struct DumpDhcpParamsReq {
+    pub port_name: String,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct DumpDhcpParamsResp {
+    pub data: DhcpCfg,
+}
+
+impl CmdOk for DumpDhcpParamsResp {}
```
luqmana, line 586: Remove these too

luqmana (approved):
2 small nits but looks good otherwise, thanks!

PR #400, opened 2023-10-24
Chores: Toolchains, Dependencies, Imports

### LINKED ISSUE #391: Simplify alloc/std imports
Most files throughout OPTE have some form of the following import block:

```rust
cfg_if! {
    if #[cfg(all(not(feature = "std"), not(test)))] {
        use alloc::string::String;
        use alloc::string::ToString;
        use alloc::vec::Vec;
    } else {
        use std::string::String;
        use std::string::ToString;
        use std::vec::Vec;
    }
}
```

This is redundant, as these are re-exported [identical types](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f70f42f1fb74eab88065acbf0c5b1f2f). This current construction stems from all our crates conditionally importing `alloc` **or** `std` – only `std` should be conditional. We should likely be doing the following:

<table>
<tr> <td>Before</td> <td>After</td> </tr>

<tr>
<td>

```rust
#![no_std]

#[cfg(any(feature = "std", test))]
#[macro_use]
extern crate std;

#[cfg(all(not(feature = "std"), not(test)))]
#[macro_use]
extern crate alloc;
```
</td>
<td>

```rust
#![no_std]

#[cfg(any(feature = "std", test))]
#[macro_use]
extern crate std;

#[macro_use]
extern crate alloc;
```

</td>
</tr>

<tr>
<td>

```rust
cfg_if! {
    if #[cfg(all(not(feature = "std"), not(test)))] {
        use alloc::string::String;
        use alloc::string::ToString;
        use alloc::vec::Vec;
    } else {
        use std::string::String;
        use std::string::ToString;
        use std::vec::Vec;
    }
}
```

</td>
<td>

```rust
use alloc::string::String;
use alloc::string::ToString;
use alloc::vec::Vec;
```

</td>
</tr>

</table>

FelixMcFelix:
This PR covers some chores and cleanup at the repository-wide level:
* Nightly toolchain (`cargo fmt` and XDE) upgraded to nightly-10-23,
* `cfg-if`'d alloc imports all simplified (#391),
* Move `MAJOR_VERSION` into `opte-api`,
* Update all dependencies to their latest versions. Of these, smoltcp has needed the most significant work, but it lets us clean up how we insert Dhcp options using the new `additional_options` field.

Most of this work is purely mechanical, with smoltcp impacting our IPv6 and DHCPv4 handling.

Closes #391.


lib/opte/src/engine/ip6.rs
```diff
@@ -632,21 +618,30 @@ pub(crate) mod test {
...
-                        Ipv6FragmentHeader::new_checked(&mut buf).unwrap();
-                    frag.emit(&mut packet);
-                    frag.buffer_len()
+                        Ipv6ExtHeader::new_checked(&mut buf).unwrap();
+                    packet.set_next_header(IpProtocol::Tcp);
+                    let mut frag_packet =
+                        Ipv6FragmentHeader::new_checked(packet.payload_mut())
+                            .unwrap();
+                    frag.emit(&mut frag_packet);
+                    2 + frag.buffer_len()
```
luqmana, line 632: that's an annoying change. I guess this is a test, but not a fan of the random magic numbers required now 😕️

lib/opte/src/engine/dhcp.rs
```diff
@@ -469,19 +492,55 @@ impl HairpinAction for DhcpAction {
...
-            );
-            Some(dns_space)
+            for server in self
+                .dhcp_cfg
+                .dns4_servers
+                .iter()
+                .take(DHCP_MAX_DNS_SERVER_COUNT)
+            {
+                _ = dns_vec.push(Ipv4Address::from(*server));
+            }
```
luqmana, line 510:
```suggestion
            dns_vec.extend(
                self.dhcp_cfg
                    .dns4_servers
                    .iter()
                    .copied()
                    .map(Ipv4Address::from)
                    .take(DHCP_MAX_DNS_SERVER_COUNT),
            );
```

luqmana (approved):
A couple comments but mostly good, thanks for cleaning this up!

crates/opte-api/src/lib.rs
```diff
@@ -58,7 +47,10 @@ pub use ulp::*;
 ///
 /// We rely on CI and the check-api-version.sh script to verify that
 /// this number is incremented anytime the oxide-api code changes.
-pub const API_VERSION: u64 = 25;
+pub const API_VERSION: u64 = 26;
```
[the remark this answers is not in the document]
FelixMcFelix: It's primarily due to the API version check in CI, which fails on any changes to the opte-api crate. It'd be nice to find a less false-positivey way to determine this.

FelixMcFelix:
Thanks @luqmana!

PR #408, opened 2023-10-30
Fix ICMP flow coalesce behind SNAT

### LINKED ISSUE #396: ICMP flows inadvertently coalesced together breaking SNAT
Using a port configured only for SNAT, you'll only be able to successfully ping the same host once until the flow entries expire (~1min).

The first ping will succeed and we can see an echo request & reply with matching IDs (1029):
```
________________________________                                                                                                                                                                                     
  172.30.3.5 -> 1.1.1.1      ETHER Type=0800 (IP), size=98 bytes                                                                                                                                                     
  172.30.3.5 -> 1.1.1.1      IP  D=1.1.1.1 S=172.30.3.5 LEN=84, ID=1299, TOS=0x0, TTL=255                                                                                                                            
  172.30.3.5 -> 1.1.1.1      ICMP Echo request (ID: 1029 Sequence number: 0)                                                                                                                                         
________________________________                                                                                                                                                                                     
     1.1.1.1 -> 172.30.3.5   ETHER Type=0800 (IP), size=98 bytes                                                                                                                                                     
     1.1.1.1 -> 172.30.3.5   IP  D=172.30.3.5 S=1.1.1.1 LEN=84, ID=30322, TOS=0x0, TTL=53                                                                                                                            
     1.1.1.1 -> 172.30.3.5   ICMP Echo reply (ID: 1029 Sequence number: 0)
```

But a subsequent ping to the same host right after will seemingly fail even though we do see the replies make it back. The difference? The reply's ID gets rewritten to the old ID (1029) instead of matching the new echo request ID (1153):
```
________________________________
  172.30.3.5 -> 1.1.1.1      ETHER Type=0800 (IP), size=98 bytes
  172.30.3.5 -> 1.1.1.1      IP  D=1.1.1.1 S=172.30.3.5 LEN=84, ID=1300, TOS=0x0, TTL=255
  172.30.3.5 -> 1.1.1.1      ICMP Echo request (ID: 1153 Sequence number: 0)
________________________________
     1.1.1.1 -> 172.30.3.5   ETHER Type=0800 (IP), size=98 bytes
     1.1.1.1 -> 172.30.3.5   IP  D=172.30.3.5 S=1.1.1.1 LEN=84, ID=40090, TOS=0x0, TTL=53
     1.1.1.1 -> 172.30.3.5   ICMP Echo reply (ID: 1029 Sequence number: 0)
```

Dumping the flow table entries, we can see the ICMP flow entries always have src/dst port 0 and so all get coalesced together for the same src/dst host pair.

```
 $ pfexec opteadm dump-uft -p opte0
UFT Inbound: 2/16384
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     XFORMS                
ICMP   1.1.1.1          0      172.17.16.15     0      0        hdr: decap,SNAT,, body: ICMP Echo Ident/SNAT 1029 <=> 16379
UDP    168.235.89.132   123    172.17.16.15     16347  0        hdr: decap,SNAT,, body: 

UFT Outbound: 2/16384
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     XFORMS                
ICMP   172.30.3.5       0      1.1.1.1          0      0        hdr: SNAT,encap, body: ICMP Echo Ident/SNAT 1029 <=> 16379
UDP    172.30.3.5       54505  168.235.89.132   123    0        hdr: SNAT,encap, body: 
```

FelixMcFelix:
We now solve this issue by extracting the ICMP echo ID during header parsing
and treating it as a pseudo src/dest port for UFT matching on ICMP(v6) flows.

This takes a bunch of extra internal plumbing: ICMPv4 as now treated as a
proper ULP like ICMPv6 is. Accordingly, ICMP ID rewriting on SNAT is now
a header transform rather than a body transform. This does have the
added benefit that we're now making use of the same incremental ULP csum
code as all other ULPs, but we do need to take special care around ICMPv4's
csum *not* including the pseudoheader.

This includes some amount of cleanup around shared types between ICMP v4
and v6. Use of HTs also makes it easy to add echo handling to `SNat6`, but that's
best left as another PR given the amount of noise that unifying `Snat`/`Snat6` would
add.

Closes #396.


lib/oxide-vpc/src/engine/mod.rs
```diff
@@ -160,28 +160,21 @@ impl NetworkParser for VpcParser {
...
 
+        let use_pseudo = ulp_hi.meta.is_pseudoheader_in_csum();
         meta.inner.ulp = Some(ulp_hi.meta);
         offsets.inner.ulp = Some(ulp_hi.offset);
 
         let body_csum = if let Some(mut csum) = ulp_hdr.csum_minus_hdr() {
-            csum -= pseudo_csum;
+            if use_pseudo {
+                csum -= pseudo_csum;
+            }
```
luqmana, line 177: Missing similar pseudoheader exclusion in `GenericUlp::parse_ulp`

lib/opte/src/engine/icmp/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    #[inline]
+    pub fn hdr_len(&self) -> usize {
+        IcmpHdr::SIZE
+    }
+
+    #[inline]
+    pub fn body_echo(&self) -> Ref<&[u8], IcmpEchoRaw> {
+        // Panic safety: Size *must* be 8B by construction.
```
luqmana, line 77: Should be 4 instead of 8? (below as well)

luqmana (approved):
A few nits but looks good overall. Thanks Kyle!

lib/oxide-vpc/src/engine/mod.rs
FelixMcFelix: Good catch!

PR #416, opened 2023-11-06
Enable ICMP Echo Request/Reply rewriting through IPv6 SNAT

### LINKED ISSUE #411: Add ICMPv6 Echo Request/Reply Handling to `SNat6`
Currently, ICMPv4 echo flows interact with source NAT by allocating themselves a port in the SNAT range, then mapping this port to/from the correct echoID. This has not been extended to ICMPv6 – this code was originally written as a body transform, and ICMPv4 was explicitly left unparsed to enable this. Now that ICMPv4 is treated as a full-fledged ULP (#408), it should be straightforward to extend the newer header transform to ICMPv6.

This should also present some opportunity to cleanup similarities and identical code between `SNat` and `SNat6`.

FelixMcFelix:
This PR takes advantage of the extra plumbing added in #408 to add ICMP Echo ID/SNAT port substitution to ICMPv6, allowing `ping` to function for IPv6 hosts/VMs without ephemeral IPs. This follows on from how ICMP v4 and v6 are both now treated as full ULPs, and makes use of the newer `icmp_id` header transform.

Additionally, there's some amount of cleanup done here. Now that we can handle these protocols identically, OPTE folds `SNat` and `SNat6` into a single `SNat<T: ConcreteIpAddr>` -- this lets us remove a lot of duplicated code between V4 and V6 impls of e.g. `ActionDesc`.

Closes #411.

### DIFF
--- a/lib/opte/src/engine/snat.rs
+++ b/lib/opte/src/engine/snat.rs
@@ -246,9 +246,8 @@ impl<T: ConcreteIpAddr + 'static> SNat<T> {
                 icmp6.echo_id()
             }
             _ => Err(GenDescError::Unexpected {
-                msg:
-                    "Mistakenly called gen_icmp_desc on non Protocol::ICMP(v6)."
-                        .to_string(),
+                msg: "Mistakenly called gen_icmp_desc on non ICMP(v6)."
+                    .to_string(),
             })?,
         }
         .ok_or(GenDescError::Unexpected {
@@ -403,7 +402,7 @@ pub struct SNatIcmpEchoDesc<T: ConcreteIpAddr> {
 pub const SNAT_ICMP_ECHO_NAME: &str = "SNAT_ICMP_ECHO";
 
 impl<T: ConcreteIpAddr> ActionDesc for SNatIcmpEchoDesc<T> {
-    // SNAT needs to generate a payload transform for ICMP traffic in
+    // SNAT needs to generate an additional transform for ICMP traffic in
     // order to treat the Echo Identifier as a psuedo ULP port.
     fn gen_ht(&self, dir: Direction) -> HdrTransform {
         match dir {
--- a/lib/opte/src/engine/snat.rs
+++ b/lib/opte/src/engine/snat.rs
@@ -86,9 +86,19 @@ impl<T: ConcreteIpAddr> Default for NatPool<T> {
 }
 
 mod private {
-    pub trait Ip: Into<super::IpAddr> {}
-    impl Ip for super::Ipv4Addr {}
-    impl Ip for super::Ipv6Addr {}
+    use opte_api::Protocol;
+
+    pub trait Ip: Into<super::IpAddr> {
+        const MESSAGE_PROTOCOL: Protocol;
+    }
+
+    impl Ip for super::Ipv4Addr {
+        const MESSAGE_PROTOCOL: Protocol = Protocol::ICMP;
+    }
+
+    impl Ip for super::Ipv6Addr {
+        const MESSAGE_PROTOCOL: Protocol = Protocol::ICMPv6;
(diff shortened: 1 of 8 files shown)


lib/opte/src/engine/snat.rs
```diff
@@ -177,129 +186,116 @@ impl<T: ConcreteIpAddr> FiniteResource for NatPool<T> {
...
-                        icmp.msg_type,
-                    ),
-                });
+        let echo_ident = match T::MESSAGE_PROTOCOL {
+            Protocol::ICMP => {
+                let icmp = meta
+                    .inner_icmp()
+                    .ok_or(GenIcmpErr::<Icmpv4Message>::MetaNotFound)?;
+                if icmp.msg_type != Icmpv4Message::EchoRequest.into() {
+                    Err(GenIcmpErr::NotRequest(icmp.msg_type))?;
```
luqmana, line 233: Nit: I would just use a simple return instead of `?` on an `Err` here (and below) or alternative also use `Ok(icmp.echo_id())` and apply `?` to the entire match.

lib/oxide-vpc/tests/integration_tests.rs
```diff
@@ -981,37 +1052,122 @@ fn unpack_and_verify_icmp4(
...
+        &icmp,
+        &smoltcp::phy::ChecksumCapabilities::ignored(),
+    )
+    .unwrap();
+
+    let mut emit_caps = smoltcp::phy::ChecksumCapabilities::ignored();
+    emit_caps.icmpv6 = smoltcp::phy::Checksum::Both;
+    let mut new_buf = vec![0; pay_len];
+    let mut icmp2 = Icmpv6Packet::new_unchecked(&mut new_buf[..]);
+    parsy.emit(&src_ip, &dst_ip, &mut icmp2, &emit_caps);
```
luqmana, line 1093: This bit seems a bit redundant and can just be removed?

lib/oxide-vpc/tests/integration_tests.rs
```diff
@@ -981,37 +1052,122 @@ fn unpack_and_verify_icmp4(
...
+    assert_eq!(icmp.echo_ident(), expected_ident);
+    assert_eq!(icmp.echo_seq_no(), seq_no);
+}
+
 // Verify that an ICMP Echo request has its identifier rewritten by
 // SNAT.
 #[test]
 fn snat_icmp4_echo_rewrite() {
+    let dst_ip: Ipv4Addr = "45.55.45.205".parse().unwrap();
+    snat_icmp_shared_echo_rewrite(dst_ip.into());
```
luqmana, line 1105: Don't need to detour through `Ipv4Addr`
```suggestion
    snat_icmp_shared_echo_rewrite("45.55.45.205".parse().unwrap());
```

lib/oxide-vpc/tests/integration_tests.rs
```diff
@@ -981,37 +1052,122 @@ fn unpack_and_verify_icmp4(
...
+}
+
+// Verify that an ICMPv6 Echo request has its identifier rewritten by
+// SNAT.
+#[test]
+fn snat_icmp6_echo_rewrite() {
+    let dst_ip: Ipv6Addr = "2001:4860:4860::8888".parse().unwrap();
+    snat_icmp_shared_echo_rewrite(dst_ip.into());
```
luqmana, line 1113:
```suggestion
    snat_icmp_shared_echo_rewrite("2001:4860:4860::8888".parse().unwrap());
```

luqmana (approved):
Love the simplification and cleanup, thanks! A couple small nits aside, lgtm

FelixMcFelix:
Thanks for taking a look! 😄

PR #418, opened 2023-11-10
Implement Per-ULP SNAT allocation

### LINKED ISSUE #412: Add per-ULP port allocations for `SNat`/`SNat6`
Talking with Luqman, he raised the point that we could probably place ICMP traffic into a separate `SNat` pool. Since dendrite maps SNAT'd traffic back to an internal address using only the dest port range, we realised that this can be extended to each ULP to offer more port allocations.

This does require extra space for each free-list, but should be a net benefit. I expect we would modify `SNat`/`SNat6` to each have a `NatPool` for TCP, UDP, and ICMP. I don't know if we also want to use this as an opportunity to move the `Arc<NatPool<_>>` handle into `NatPoolEntry` to make it easier to automatically return ports to the correct pool.

FelixMcFelix:
This PR provides TCP/UDP/ICMP-Echo traffic on a SNAT with separate pools of available public ports for use, and moves some elements of `NatEntry` structs into a shared handle to simplify correctly returning ports to the intended pool.

Closes #412.

### DIFF
--- a/lib/opte/src/engine/rule.rs
+++ b/lib/opte/src/engine/rule.rs
@@ -86,14 +86,59 @@ pub trait FiniteResource: Resource {
 
     /// Obtain a new entry given the key.
     ///
+    /// Callers are responsible for manually `release`ing this entry into
+    /// the correct parent pool.
+    ///
     /// # Errors
     ///
     /// Return an error if no entry can be mapped to this key or if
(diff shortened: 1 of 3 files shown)


luqmana (approved):
Thanks for improving this!

PR #420, opened 2023-11-14
(Multiple) Floating IP support

### LINKED ISSUE #196: Want support for multiple external IP addresses for a Port

FelixMcFelix:
This PR focusses on adding support for floating IP addresses to ports. Broadly speaking the intended behaviour is that:
* a host will listen to all inbound traffic on all its external IPs (floating and ephemeral),
* outbound traffic will choose an external IP i order of priority from `floating > ephemeral > SNat`.
* outbound flows using a floating IP will choose between several (if supplied) by hashing the inner flow ID.

This includes the necessary changes such that UFTs created by inbound flows will now store their original destination address in their `ActionDesc`, preserving flow affinity.

Will close #196.


lib/opte/src/engine/layer.rs
```diff
@@ -1402,12 +1503,33 @@ impl Layer {
...
     ) {
         self.ft.clear();
+        self.set_rules_core(in_rules, out_rules);
+    }
+
+    /// Set all rules at once without clearing the flow table.
+    ///
+    /// See [FlowTable::mark_dirty] for the performance and correctness
+    /// implications.
+    pub fn set_rules_soft(
```
FelixMcFelix, line 1519: We could, with the current design, just revalidate all actions at the time `set_rules` is called. The lazy approach was partly designed around possibly needing access to packet metadata and body for action validation down the line.

luqmana (commented):
Thanks for getting this together Kyle!

Still looking through things but one overall thing: The whole ephemeral vs floating distinction feels like it should live in just oxide-vpc/omicron vs opte proper.

FelixMcFelix:
> Thanks for getting this together Kyle!
> 
> Still looking through things but one overall thing: The whole ephemeral vs floating distinction feels like it should live in just oxide-vpc/omicron vs opte proper.

Thanks, Luqman. I agree -- I've generally tried to structure this so that OPTE itself still only knows about NAT rules wrt individual external IPs, plus mechanisms for `NetworkImpl` providers to re-validate rules and preserve affinity (`Inbound` vs `Outbound` NAT actions). oxide-vpc is responsible for knowing about and enforcing the (eph + float IPs)<->(rules + priorities) mapping.

lib/opte/src/engine/nat.rs
```diff
@@ -24,40 +24,146 @@ use alloc::string::ToString;
...
+        if self.external_ips.len() > 1 {
+            write!(f, "{{")?;
+        }
+
+        for (i, ip) in self.external_ips.iter().enumerate() {
+            write!(f, "{ip}")?;
+            if i != self.external_ips.len() - 1 {
+                write!(f, ",")?;
+            }
+        }
```
luqmana, line 80: We already have `itertools` as a dep right? Could use [`Itertools::format`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.format):
```suggestion
        write!(f, "{}", self.external_ips.iter().format(","));
```

lib/opte/src/dynamic.rs
```diff
@@ -0,0 +1,76 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Copyright 2023 Oxide Computer Company
+
+//! A KMutex-based wrapper for dynamically updateable resources (e.g., config),
+//! and the outputs *generated* from
```
luqmana, line 8: Nit: drop the "from"?

lib/opte/src/dynamic.rs
```diff
@@ -0,0 +1,76 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Copyright 2023 Oxide Computer Company
+
+//! A KMutex-based wrapper for dynamically updateable resources (e.g., config),
```
luqmana, line 7:
```suggestion
//! A KRwLock-based wrapper for dynamically updateable resources (e.g., config),
```

luqmana:
> > Thanks for getting this together Kyle!
> > Still looking through things but one overall thing: The whole ephemeral vs floating distinction feels like it should live in just oxide-vpc/omicron vs opte proper.
> 
> Thanks, Luqman. I agree -- I've generally tried to structure this so that OPTE itself still only knows about NAT rules wrt individual external IPs, plus mechanisms for `NetworkImpl` providers to re-validate rules and preserve affinity (`Inbound` vs `Outbound` NAT actions). oxide-vpc is responsible for knowing about and enforcing the (eph + float IPs)<->(rules + priorities) mapping.

Yup, looked through the rest of the PR some more now and looks good. The only confusing bit was `opteadm` referring to ephemeral v.s. floating, but it's already `oxide-vpc` specific (maybe we should rename it at some point)

luqmana (approved):
Thanks!

Issue #426, opened 2023-12-12
Long TCP sessions lose their UFT entries


FelixMcFelix:
@citrus-it found that [ZFS transfers from a DigitalOcean VM to Colo](https://gist.github.com/citrus-it/fef4602f4c0494efd188a990665fd4e7) are failing with some odd characteristics -- transfers failed around 3.98GB completion, while outbound traffic on the connection (e.g., ACKs) successfully transits Colo->DO.

```
╔═════════════╗                   ╔═════════════╗
║             ╠────────ACKs───────▶             ║
║    Colo     ║                   ║DigitalOcean ║
║     Vm      ║             TCP   ║     VM      ║
║             ◀ ─ ─ ─ ─ X◀──Segs──╣             ║
╚═════════════╝                   ╚═════════════╝
```

We spent some time digging into this, broadly finding:
* NAT mappings are a-okay (`swadm nat get -e <ip> -p 22`), although they were invisible via `swadm nat list`.
* Issue persists after sled migration.
* Issue does *not* present itself when using iperf3 to send traffic from DO->Colo, but is dead reproducible with a zfs copy.
* Failure is not explicitly tied to size, but seems to happen at or around 30s boundaries (3:00, 3:30, 4:00, ...) [i.e., up to around 9GB].

Focussing on OPTE, we found that the UFT-In entry for the above flow is suddenly *expired* according to DTrace, in spite of its hit count actively increasing. After this point, no segments reach the Colo VM in spite of retransmits (no replacement UFT entry built). Expiry in particular is **very weird**: our understanding is that the packet <1s before expiry should have updated the `FlowEntry`'s `last_hit`. Our gut feelings seemed to indict either the TCP state tracking and/or `Moment::delta_as_millis()` -- while `ddi::gethrtime()` is monotonic, it was suggested that there might be per-core variation.

citrus-it:
Here's an example of an active outbound UFT entry going away. This corresponds with the `flow-expired` USDT probe firing. One can see that the hit count is increasing right up to the point that the flow disappears, so the open question is around why it was deemed to have expired.

```
Tue Dec 12 15:14:41 UTC 2023

UFT Inbound: 3/65535
----------------------------------------------------------------------
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     XFORMS
TCP    164.90.148.125   22     45.154.216.124   38174  7365157  hdr: decap,NAT,, body:

UFT Outbound: 3/65535
[...]

PR #427, opened 2023-12-12
Add `last_hit` to uft hit/expire probes, prevent `Moment` time delta wrapping

FelixMcFelix:
Adds some extra probes to help us track down and verify #426, and also
slightly changes up `Moment::delta_as_millis()` to saturate if the order
if `now` and `earlier` are seemingly back to front (as is apparently
now standard behaviour in std).

### DIFF
--- a/dtrace/opte-flow-expire.d
+++ b/dtrace/opte-flow-expire.d
@@ -8,7 +8,8 @@
 #include "common.h"
 #include "protos.d"
 
-#define	HDR_FMT "%-24s %-18s %s %s %s\n"
+#define	HDR_FMT		"%-24s %-18s %s %s %s\n"
+#define	LINE_FMT	"%-24s %-18s %s %u %u\n"
 
 BEGIN {
 	printf(HDR_FMT, "PORT", "FT NAME", "FLOW", "LAST_HIT", "NOW");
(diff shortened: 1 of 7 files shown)


dtrace/opte-flow-expire.d
```diff
@@ -8,20 +8,22 @@
...
+	printf(HDR_FMT, "PORT", "FT NAME", "FLOW", "LAST_HIT", "NOW");
 	num = 0;
 }
 
 flow-expired {
 	this->port = stringof(arg0);
 	this->name = stringof(arg1);
 	this->flow = (flow_id_sdt_arg_t *)arg2;
+	this->last_hit = stringof(arg3);
+	this->now = stringof(arg4);
```
citrus-it, line 23: Are these really strings, and not u64s?

lib/opte/src/ddi/time.rs
```diff
@@ -75,6 +77,19 @@ impl Moment {
             }
         }
     }
+
+    /// Return the underlying timestamp for debugging purposes
+    /// if supported on the current platform.
+    #[allow(dead_code)]
```
rcgoodfellow, line 83: I presume this is because we're only calling this in a `no-std` context?

rcgoodfellow (approved):
LGTM. Just one non-blocking question.

lib/opte/src/ddi/time.rs
FelixMcFelix: It is, yes.

dtrace/opte-flow-expire.d
FelixMcFelix: Good catch, I'll convert this to match opte-uft-hit's stucture.

Issue #428, opened 2023-12-13
DHCP Hostname option is double encoded on the wire.


citrus-it:
The [Hostname option ](https://www.rfc-editor.org/rfc/rfc2132#section-3.14) in OPTE's DHCP responses appears to be double encoded, such that the hostname seen by the guest is prefixed with the length of the hostname itself.

For example:

`\x07us-west`
and
`\x04ekke`

On the latter VM, the `dhcpinfo` command shows this:

```
^Dekke# /sbin/dhcpinfo -c Hostname
0x04
0x65
0x6B
0x6B
0x65
0x00
```
(You can also see the ^D in the prompt).

There's also a trailing \x0 that should not be there.

I captured a packet trace to confirm what was on the wire, and decoded it with wireshark. That shows:

<img width="218" alt="image" src="https://github.com/oxidecomputer/opte/assets/29426693/31841229-e2e4-4745-84d4-a96541691ad1">

correponding to the byte sequence:

<img width="548" alt="image" src="https://github.com/oxidecomputer/opte/assets/29426693/f5a47455-344f-48bb-9969-107253ecc88f">

After the `0c` indicating the Hostname option, there is the length (6) and then the 6 bytes as shown by `dhcpinfo` above.

citrus-it:
Here, for comparison, is a DHCP response from ISC DHCP server, containing both a Hostname (12) and Domainname (15) response:

<img width="554" alt="image" src="https://github.com/oxidecomputer/opte/assets/29426693/f4db3730-7c87-494c-8738-7a6d6a5b717d">

citrus-it:
From code inspection, it looks like these are actually being encoded using the label-sequence format of RFC 1035 section 3.1, which is incorrect for DHCP.

luqmana:
Hmm, reading that part of the spec again it seems a bit ambiguous?

[RFC 1035 3.1. Name space definitions](https://datatracker.ietf.org/doc/html/rfc1035) is where the label length prefix + ending NUL byte encoding comes from:
>Domain names in messages are expressed in terms of a sequence of labels. Each label is represented as a one octet length field followed by that number of octets.  Since every domain name ends with the null label of the root, a domain name is terminated by a length byte of zero.

[RFC 2132 3.14. Host Name Option](https://datatracker.ietf.org/doc/html/rfc2132#section-3.14) doesn't explicitly call out the encoding but does say:
[...]

PR #429, opened 2023-12-13
DHCP DomainName options are being incorrectly RFC1035 encoded

citrus-it:
Fixes https://github.com/oxidecomputer/opte/issues/428


luqmana (approved):
Thanks for catching this! Looks good, I'd just add a test case or two with `test_option_emit`

FelixMcFelix (approved):
Looks good, thanks for fixing this Andy!

PR #444, opened 2024-01-08
TCP flow state refactor and fixes

### LINKED ISSUE #436: Old TCP flows are in `FLOW_WAIT`, blocking port reuse
@citrus-it observed that [outbound TCP flows will occasionally hang](https://gist.github.com/citrus-it/f5a597edadc01cc0c42900f4d2a9d8fe) while running a package sync operation from an OmniOS VM in Colo to both a DigitalOcean-hosted and Oxide-hosted mirror of the package repository. This workload is notable for spinning up a large number of short-lived TCP flows. Specifically the host will remain in `SYN-SENT`, none of the (repeated) SYN packets will be seen by the other host, and the OPTE state is caught in `TIME-WAIT` fo an unbounded time.

An initial reproduction hit this using `curl -I <site>` in a bash loop, but we found that this can be consistently triggered using `curl -I --local-port 56000 <site>`:
1. Invoke curl. HTTP HEAD request completes successfully.
  a. `opteadm dump-tcp-flows` shows flow is in state `TIME-WAIT`.
2. Invoke curl again.
  a. `snoop` shows SYNs being sent w/ exponential backoff. Reply hangs.
  b. `opteadm dump-tcp-flows` shows flow remains in state `TIME-WAIT`. Out byte/pkt count seems to increment.
  c. `kstat` for port in question shows `out_drop_tcp_err` increments in time with SYNs.
3. Ctrl-C.
  a. `snoop` shows RST sent.
  b. `opteadm dump-tcp-flows` shows flow is cleared.
4. GOTO 1.

Currently we're not doing any periodic pruning of TCP flow table entries, which makes it quite likely that one will encounter this on a long-running VM (or short-running if using `pkg`!).
* We could do periodic pruning, but this presents the issue that we don't know exactly what a guest's TCP_WAIT timer configuration is.
* The alternative is that, at least in the outbound case, a new SYN should always overwrite flow-state. I'd need to think more about whether it's sane to do this on inbound flows.

### LINKED ISSUE #442: TCP state machine race condition can leave connection in the wrong state
I was looking at the packet trace that @jgallagher took from a customer site as part of https://github.com/oxidecomputer/customer-support/issues/25 to try and understand what happened here.

image

This is the TCP termination sequence from an API ping request. What's particularly interesting is that we know that a subsequent request from source port 50030 some 45 seconds later failed with the same symptoms as #436. However, in this case the client ended up resetting the connection, bypassing the normal TIME_WAIT and CLOSE_WAIT states, and so was free to re-use the port.

Looking at the logic in the TCP state machine, a packet with the `RST` bit should immediately put the flow into `Closed`. However, any packet with the ACK bit set will punt a `Closed` flow back into `Established` and that is what I suspect has happened here. That would mean that opte processed the last two packets in the above trace in the reverse order, resulting in a connection stuck in `Established`. Unfortunately I can't confirm that with the available data.

FelixMcFelix:
This PR introduces two key changes to how we handle TCP flows, and refactors to give us some more consistency between how they're handled by {in, out}bound packets when we {do, don't} have an existing UFT entry.

### 1. Time-based flow expiry for TCP flows.
At present, TCP sessions which have been passive-closed by the guest are correctly moved to the CLOSED state and removed, while those which are active-closed become stuck in a TIME_WAIT state.

To remedy this, we add two fairly pessimistic cleanup timers: 120s for TIME-WAIT, and 2.5 hours for other states (the default 2 hour SO_KEEPALIVE time with a fudge factor). The CLOSED -> ESTABLISHED shortcut should allow an entry to be readded with shorter timers, however we might in future store more detailed flow state which would be lost with a shorter timeout.

### 2. Allow unexpected segments and reset flow stats when needed.
Existing TCP flows will no longer filter any packets which would correspond to unexpected state transitions, and instead fire a dtrace probe.

However, to keep stats consistent we have to do some more work to invalidate flows if we see an unexpected *SYN*-flagged packet. The main value of this is that if our flow state tracking is affected by out-of-order delivery, we can no longer be stuck in `Established` as in #442, but this also allows port reuse faster than every 120s (which is the other component of #436).

---

Should close #436 and close #442.

### DIFF
--- a/lib/opte/src/engine/port.rs
+++ b/lib/opte/src/engine/port.rs
@@ -828,7 +828,7 @@ impl<N: NetworkImpl> Port<N> {
     //
     fn tcp_err(
         &self,
-        data: &mut PortData,
+        data: &FlowTable<TcpFlowEntryState>,
         dir: Direction,
         msg: String,
         pkt: &mut Packet<Parsed>,
@@ -837,7 +837,7 @@ impl<N: NetworkImpl> Port<N> {
             super::err(format!("mblk: {}", pkt.mblk_ptr_str()));
             super::err(format!("flow: {}", pkt.flow()));
             super::err(format!("meta: {:?}", pkt.meta()));
-            super::err(format!("flows: {:?}", data.tcp_flows,));
+            super::err(format!("flows: {:?}", data));
             todo!("bad packet: {}", msg);
         } else {
             self.tcp_err_probe(dir, pkt, msg)
@@ -1506,24 +1506,24 @@ impl<N: NetworkImpl> Port<N> {
                 tfes.segs_in += 1;
                 tfes.bytes_in += pkt_len;
 
-                match tfes.tcp_state.process(
+                let next_state = tfes.tcp_state.process(
                     self.name_cstr.as_c_str(),
                     In,
                     &ufid_out,
                     tcp,
+                );
+
+                if matches!(
+                    next_state,
+                    Ok(TcpState::Closed)
+                        | Err(TcpFlowStateError::NewFlow { .. })
                 ) {
-                    Ok(tcp_state) => {
-                        if tcp_state == TcpState::Closed {
-                            let entry = tcp_flows.remove(&ufid_out).unwrap();
-                            let ufid_in = entry.state().inbound_ufid.as_ref();
-                            self.uft_tcp_closed(data, &ufid_out, ufid_in);
-                        }
-
-                        Ok(tcp_state)
-                    }
-
-                    Err(e) => Err(ProcessError::TcpFlow(e)),
+                    let entry = tcp_flows.remove(&ufid_out).unwrap();
+                    let ufid_in = entry.state().inbound_ufid.as_ref();
+                    self.uft_tcp_closed(data, &ufid_out, ufid_in);
                 }
+
+                next_state.map_err(ProcessError::TcpFlow)
             }
 
             None => Err(ProcessError::MissingFlow(ufid_out)),
@@ -1719,19 +1719,22 @@ impl<N: NetworkImpl> Port<N> {
                     }
                 }
 
+                // Unlike for existing flows, we don't allow through
+                // unexpected packets here for now -- the `TcpState` FSM
+                // already encodes a shortcut from `Closed` to `Established.
                 Err(ProcessError::TcpFlow(err)) => {
                     let e = format!("{err}");
-                    self.tcp_err(data, Direction::In, e, pkt);
+                    self.tcp_err(&data.tcp_flows, Direction::In, e, pkt);
                     Ok(ProcessResult::Drop { reason: DropReason::TcpErr })
                 }
                 Err(ProcessError::MissingFlow(flow_id)) => {
                     let e = format!("Missing TCP flow ID: {flow_id}");
-                    self.tcp_err(data, Direction::In, e, pkt);
+                    self.tcp_err(&data.tcp_flows, Direction::In, e, pkt);
                     Ok(ProcessResult::Drop { reason: DropReason::TcpErr })
                 }
                 Err(ProcessError::FlowTableFull { kind, limit }) => {
                     let e = format!("{kind} flow table full ({limit} entries)");
-                    self.tcp_err(data, Direction::In, e, pkt);
+                    self.tcp_err(&data.tcp_flows, Direction::In, e, pkt);
                     Ok(ProcessResult::Drop { reason: DropReason::TcpErr })
                 }
                 res => unreachable!(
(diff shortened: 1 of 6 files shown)


FelixMcFelix:
From local testing with an alpine-linux image, after an `apk add curl && curl google.com --local-port 40000`:

```bash
kyle@farme:~/gits/oxide.rs$ pfexec opteadm dump-tcp-flows -p opte6
FLOW                                             STATE        HITS     SEGS IN  SEGS OUT BYTES IN   BYTES OUT
TCP:172.30.0.5:36768:151.101.2.132:80            TIME_WAIT    908      467      442      723081     30087
TCP:172.30.0.5:37896:151.101.66.132:80           TIME_WAIT    2477     1583     895      2444208    60960
TCP:172.30.0.5:37906:151.101.66.132:80           TIME_WAIT    2512     1583     930      2445719    64034
TCP:172.30.0.5:40000:142.250.200.14:80           TIME_WAIT    9        4        6        1341       477
TCP:172.30.0.5:40454:151.101.194.132:80          TIME_WAIT    1227     704      524      1081416    35753
TCP:172.30.0.5:40678:172.217.169.46:80           TIME_WAIT    9        4        6        1341       477

kyle@farme:~/gits/oxide.rs$ sleep 120 && pfexec opteadm dump-tcp-flows -p opte6
FLOW                                             STATE        HITS     SEGS IN  SEGS OUT BYTES IN   BYTES OUT
```

lib/opte/src/engine/port.rs
```diff
@@ -1460,6 +1483,147 @@ impl<N: NetworkImpl> Port<N> {
...
+    ///
+    /// Unexpected TCP segments on existing connections will be allowed,
+    /// but will fire DTrace probes via `Self::tcp_err_probe`.
+    ///
+    /// # Errors
+    /// * `ProcessError::MissingFlow` if no flow currently exists.
+    /// * `ProcessError::TcpFlow(NewFlow { .. })` if this packet retired
+    ///   an existing TCP flow state entry.
+    ///
+    /// Callers which expect an existing UFT entry should resubmit a packet
```
rcgoodfellow, line 1557: Could you specify where in the packet processing pipeline such a resubmission should take place? At face value, I can't tell if this is referring to this function or somewhere else.

lib/opte/src/engine/tcp_state.rs
```diff
@@ -203,6 +221,13 @@ impl TcpFlowState {
                     return Some(CloseWait);
                 }
 
+                if tcp.has_flag(TcpFlags::SYN) {
```
rcgoodfellow, line 224: This is the remedy for the issue we saw in the colo?
- oxidecomputer/colo#61

rcgoodfellow (commented):
Thanks Kyle! Looks good on a first read. Just a few comments/questions. Going to take this through testing on `a4x2` to kick the tires a bit.

lib/opte/src/engine/port.rs
```diff
@@ -993,14 +1011,19 @@ impl<N: NetworkImpl> Port<N> {
...
+        let now = now.unwrap_or_else(Moment::now);
         check_state!(data.state, [PortState::Running])?;
 
         for l in &mut data.layers {
             l.expire_flows(now);
         }
         let _ = data.uft_in.expire_flows(now, |_| FLOW_ID_DEFAULT);
         let _ = data.uft_out.expire_flows(now, |_| FLOW_ID_DEFAULT);
+
+        // XXX: TCP state expiry currently runs on a longer time scale than
```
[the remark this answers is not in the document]
FelixMcFelix: This is worded poorly. I meant that any flows removed from the TCP table should be jointly removed from the UFT, but right now that's trivially true. We're given a Vec of expired flowkeys from every FlowTable::expire_flows, this comment says we might want to use that to feed back into UFT expiry if, e.g., we set a sub-60s expiry for the TCP table or a specific state.

lib/opte/src/engine/port.rs
```diff
@@ -1460,6 +1483,147 @@ impl<N: NetworkImpl> Port<N> {
...
+
+        let tcp_state = match tfs.process(
+            self.name_cstr.as_c_str(),
+            dir.dir(),
+            dir.local_flow(),
+            tcp,
+        ) {
+            Ok(tcp_state) => tcp_state,
+
+            // We're intentionally not allowing through unexpected segments or
```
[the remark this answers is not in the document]
FelixMcFelix: That is the case, although I think the only packets we're realistically blocking on flow create are bare FINs and RSTs. We could extend it to every packet and just not create the flow.

lib/opte/src/engine/tcp_state.rs
FelixMcFelix: No, this is more related to #442 and its interaction with the `NewFlow` handling -- this is accounting for the fact that Established flows would allow non-FIN/RST packets through without issue. The consequence was that a new flow would appear considerably older / include packets from an older incarnation in its counts, so this is just for tracking statistics more correctly in such a case.

The remedy for that bug and/or #436  is the remainder of the patch: TIME_WAITs should be actually cleaned up, they won't filter packets, and a new SYN should practically always supersede flowtable state.

rcgoodfellow:
Took this for a spin in `a4x2` and things look good.

rcgoodfellow (approved):
Thanks Kyle. This LGTM.

PR #449, opened 2024-01-26
Add Xtask for XDE build/install/package operations

### LINKED ISSUE #366: Include debug bits as variant in generated package.
Today we only include the release artifacts in the generated `driver/network/opte` package:

https://github.com/oxidecomputer/opte/blob/d367866fe4ae47df27f8881dc8d93b86d14f15e0/pkg/build.sh#L16-L19

We should also include the debug versions as a variant.

### LINKED ISSUE #413: Move XDE package building and installation to xtask
Currently, XDE is built using the `xde/build.sh` script. Locally I'm using a simple `reinstall.sh` which calls this, rebuilds opteadm, and then copies the binaries into the correct places – however, this only works after the first package installation (or manually setting up a link from `/dev/xde -> /devices/pseudo/xde@0:ctl`.

Ideally, I would like these to be a single xtask which rebuilds and installs XDE from the current repository in one command, with or without a package (re)build. This may also be a good chance to solve #366.

FelixMcFelix:
This PR adds some `cargo xtask`s to build and install XDE and opteadm on a given host -- `cargo xtask {build, install, package}`. These include flags for `pkg`-based installation as well as a brute-force method which simply copies our artifacts into the right place.

I've also bumped the nightly toolchain used by XDE since it's been some time since we've done so, and added `cargo xtask fmt` to automatically format using the correct nightly toolchain.

Closes #413 and closes #366.

### DIFF
--- a/pkg/build.sh
+++ b/pkg/build.sh
@@ -21,12 +21,9 @@ cp ../target/i686-unknown-illumos/release/libxde_link.so proto/usr/lib/devfsadm/
 if [ -z ${RELEASE_ONLY+x} ]; then
     cp ../target/debug/opteadm proto/opt/oxide/opte/bin/opteadm.dbg
     cp ../target/x86_64-unknown-unknown/debug/xde.dbg proto/kernel/drv/amd64/xde.dbg
+    INC_DEBUG=""
 else
-    # A bit of a hack for local release-only pkg builds.
-    # There are probably pkgmogrify transforms to do this correctly,
-    # based on RELEASE_ONLY!
-    cp proto/opt/oxide/opte/bin/opteadm proto/opt/oxide/opte/bin/opteadm.dbg
-    cp proto/kernel/drv/amd64/xde proto/kernel/drv/amd64/xde.dbg
+    INC_DEBUG="#"
 fi
 
 API_VSN=$(./print-api-version.sh)
@@ -35,7 +32,7 @@ API_VSN=$(./print-api-version.sh)
 sed -e "s/%PUBLISHER%/$PUBLISHER/g" \
     -e "s/%COMMIT_COUNT%/$COMMIT_COUNT/g" \
     -e "s/%API_VSN%/$API_VSN/g" \
-    opte.template.p5m | pkgmogrify -v -O opte.base.p5m
+    opte.template.p5m | pkgmogrify -v -D inc_debug="$INC_DEBUG" -O opte.base.p5m
 
 pkgdepend generate -d proto opte.base.p5m > opte.generate.p5m
 
--- a/pkg/build.sh
+++ b/pkg/build.sh
@@ -14,9 +14,20 @@ mkdir -p proto/kernel/drv/amd64
 mkdir -p proto/opt/oxide/opte/bin
 mkdir -p proto/usr/lib/devfsadm/linkmod
 cp ../target/release/opteadm proto/opt/oxide/opte/bin/
-cp ../target/x86_64-unknown-unknown/release/xde proto/kernel/drv/amd64/xde
+cp ../target/x86_64-unknown-unknown/release/xde proto/kernel/drv/amd64
 cp ../xde/xde.conf proto/kernel/drv/
 cp ../target/i686-unknown-illumos/release/libxde_link.so proto/usr/lib/devfsadm/linkmod/SUNW_xde_link.so
+
+if [ -z ${RELEASE_ONLY+x} ]; then
+    cp ../target/debug/opteadm proto/opt/oxide/opte/bin/opteadm.dbg
+    cp ../target/x86_64-unknown-unknown/debug/xde.dbg proto/kernel/drv/amd64/xde.dbg
+else
+    # A bit of a hack for local release-only pkg builds.
+    # There are probably pkgmogrify transforms to do this correctly,
+    # based on RELEASE_ONLY!
+    cp proto/opt/oxide/opte/bin/opteadm proto/opt/oxide/opte/bin/opteadm.dbg
+    cp proto/kernel/drv/amd64/xde proto/kernel/drv/amd64/xde.dbg
+fi
 
 API_VSN=$(./print-api-version.sh)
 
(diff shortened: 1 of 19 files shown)


pkg/build.sh
```diff
@@ -14,10 +14,21 @@ mkdir -p proto/kernel/drv/amd64
...
 cp ../xde/xde.conf proto/kernel/drv/
 cp ../target/i686-unknown-illumos/release/libxde_link.so proto/usr/lib/devfsadm/linkmod/SUNW_xde_link.so
 
+if [ -z ${RELEASE_ONLY+x} ]; then
+    cp ../target/debug/opteadm proto/opt/oxide/opte/bin/opteadm.dbg
+    cp ../target/x86_64-unknown-unknown/debug/xde.dbg proto/kernel/drv/amd64/xde.dbg
+else
+    # A bit of a hack for local release-only pkg builds.
+    # There are probably pkgmogrify transforms to do this correctly,
+    # based on RELEASE_ONLY!
```
luqmana, line 27: yea, seems like some `-D` flags to `pkgmogrify`. Definitely would be nice not to just add duplicates if not needed.

pkg/opte.template.p5m
```diff
@@ -6,11 +6,15 @@ set name=info.classification \
...
 set name=variant.arch value=i386
+set name=variant.debug.illumos value=false value=true
 dir path=opt owner=root group=sys mode=0755
 dir path=opt/oxide owner=root group=bin mode=0755
 dir path=opt/oxide/opte owner=root group=bin mode=0755
 dir path=opt/oxide/opte/bin owner=root group=bin mode=0755
-file path=opt/oxide/opte/bin/opteadm owner=root group=bin mode=0755
+file path=opt/oxide/opte/bin/opteadm owner=root group=bin mode=0755 \
+    variant.debug.illumos=false
+file opt/oxide/opte/bin/opteadm.dbg path=opt/oxide/opte/bin/opteadm owner=root group=bin mode=0755 \
```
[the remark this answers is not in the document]
luqmana: Barring any other issue, I'd say why not. The variant stuff already means you only get one or the other in a given boot environment.

pkg/opte.template.p5m
```diff
@@ -33,5 +37,12 @@ file path=usr/lib/devfsadm/linkmod/SUNW_xde_link.so owner=root group=sys mode=07
 #:
```
luqmana, line 37: (darn github too short context)

curious if we still need the bypass stuff mention in the XXX above

xtask/src/main.rs
```diff
@@ -0,0 +1,431 @@
...
+    Install {
+        /// Install from an illumos package file rather than by copying
+        /// the drivers into place.
+        #[arg(long)]
+        from_package: bool,
+
+        /// Override any package freeze held in place by omicron.
+        ///
+        /// No-op if `from_package` is not specified.
+        #[arg(long)]
```
luqmana, line 42:
```suggestion
        #[arg(long, requires = "from_package"]
```

pkg/build.sh
FelixMcFelix: Done!

xtask/src/main.rs
FelixMcFelix: Done.

pkg/opte.template.p5m
FelixMcFelix: Yeah, we still do:

```
+ pkgdepend resolve -d packages -s resolve.p5m opte.generate.p5m
/develop/gits/opte/pkg/opte.generate.p5m has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=dld \
        pkg.debug.depend.reason=kernel/drv/amd64/xde \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.path=kernel/misc/amd64 \
        pkg.debug.depend.path=usr/kernel/misc/amd64
' under the following combinations of variants:
variant.debug.illumos:false variant.opensolaris.zone:global
/develop/gits/opte/pkg/opte.generate.p5m has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=dld \
        pkg.debug.depend.reason=kernel/drv/amd64/xde \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.path=kernel/misc/amd64 \
        pkg.debug.depend.path=usr/kernel/misc/amd64
' under the following combinations of variants:
variant.debug.illumos:true variant.opensolaris.zone:global
/develop/gits/opte/pkg/opte.generate.p5m has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=mac \
        pkg.debug.depend.reason=kernel/drv/amd64/xde \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.path=kernel/drv/amd64 \
        pkg.debug.depend.path=usr/kernel/drv/amd64
' under the following combinations of variants:
variant.debug.illumos:false variant.opensolaris.zone:global
/develop/gits/opte/pkg/opte.generate.p5m has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=mac \
        pkg.debug.depend.reason=kernel/drv/amd64/xde \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.path=kernel/drv/amd64 \
        pkg.debug.depend.path=usr/kernel/drv/amd64
' under the following combinations of variants:
variant.debug.illumos:true variant.opensolaris.zone:global
```

I briefly tested different `pkgdepend -k <path>` and `pkg.depend.runpath=<path>:$PKGDEPEND_RUNPATH` to no luck.

luqmana: ah, unfortunate :/ i don't know enough about illumos packaging, so perhaps worth a question in chat. in any case, doesn't need to block this PR

luqmana (approved):
Thanks for adding this, definitely lowers the friction in testing changes!

pkg/opte.template.p5m
rmustacc: @citrus-it can likely help explain what's required and what's going on and may have some ideas here.

PR #452, opened 2024-02-07
Use TabWriter for `opteadm` output

FelixMcFelix:
This PR changes the formatting on opteadm to be constructed using TabWriter, which should provide a small improvement over the (many) hand-coded padding lengths in today's output strings.

For additional tweaks I've also added the current port name to the output of `opteadm dump-layer`, which should make it somewhat easier to track down the right layers when looking at integration test output on omicron. Multiple external IPs in `list-ports` are now also split over newlines, matching the style of multi-predicate layer rules.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -860,6 +860,7 @@ dependencies = [
  "anyhow",
  "cfg-if",
  "clap",
+ "itertools",
  "libc",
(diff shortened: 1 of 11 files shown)


rcgoodfellow (approved):
Thanks for doing this. Some of the column layouts were getting very wide with static spacing.

PR #456, opened 2024-02-12
Fix Geneve size in `integration_tests::common::encap`

### LINKED ISSUE #454: Inbound external packets contain uninitialised mblk data in `cargo test`
Updating to a newer rust version in #453 has turned up an issue where inbound packets can contain bonus data after processing. Replicating evidence from the issue:

---

| `cargo -V` | Csum bytes |
|--------|--------|
| `cargo 1.75.0 (1d8b05cdd 2023-11-20)` | `[0x3d, 0xac]` |
| `cargo 1.76.0 (c84b36747 2024-01-18)` | <different every time, random> |
| `cargo 1.77.0-nightly (2ce45605d 2024-01-04)` | `[0x3d, 0x2c]` | 
| `cargo 1.78.0-nightly (cdf84b69d 2024-02-02)` | <different every time, random> |

`DATA_VEC` below refers to all bytes, up to current recorded len, copied from `Packet.segs` into a single `Vec<u8>`, indexewd from the start of the ICMP header. The inbound rewritten packets on 1.75 have a trailer of all zeroes:
```
[3d, ac]
DATA_VEC? [0, 0, 3d, ac, 0, 7, 3, 9, 72, 65, 75, 6e, 69, 6f, 6e, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```
while those on other compiler versions have garbage data:
```
[3d, ac]
DATA_VEC? [0, 0, 3d, ac, 0, 7, 3, 9, 72, 65, 75, 6e, 69, 6f, 6e, 0, 0, 0, 0, 80, f0, 5c, df, ea]
```

---

Digging into the pseudo `mblk_t`s, these are init'ed using `Vec::with_capacity` in user/test code (so contents are not zero-initialized). I spent some time looking at mblk body addresses and I didnt directly see any historic overlaps, but we're clearly overlapping some older region of the heap.

The fastest fix is probably to more aggressively trim output packets in `Direction::In` -- I assume we're specifically 8 bytes over, at least in the ICMP case. I think the more correct longer-term fix is to be more honest in `Packet` around what it means to `expand` a packet segment. We obviously don't want to zero-fill a buffer, but we should probably be incrementing the wptr *after* a write into a `MaybeUninitialized` region, rather than incrementing wptr and handing out `&[u8]`s into that memory.

FelixMcFelix:
The `encap` method has, for its lifetime, been adding the length of a UDP header as well as `GeneveHdr::BASE_SIZE` when computing the length of the Geneve packet metadata for encapsulating test packets. However, `GeneveHdr` already includes UDP and its fields, so we were requesting an extra 8 bytes without initialising them. This PR corrects that length computation.

Fixes #454.

### DIFF
--- a/lib/opte/src/engine/geneve.rs
+++ b/lib/opte/src/engine/geneve.rs
@@ -14,6 +14,8 @@ use super::headers::PushAction;
 use super::headers::RawHeader;
 use super::packet::PacketReadMut;
 use super::packet::ReadErr;
+use super::udp::UdpHdr;
+use super::udp::UdpMeta;
 use core::mem;
 pub use opte_api::Vni;
 use serde::Deserialize;
@@ -41,8 +43,6 @@ pub const GENEVE_OPT_CLASS_OXIDE: u16 = 0x0129;
 pub struct GeneveMeta {
     pub entropy: u16,
     pub vni: Vni,
-    pub len: u16,
-
     pub oxide_external_pkt: bool,
 }
 
@@ -87,9 +87,10 @@ impl ModifyAction<GeneveMeta> for GeneveMod {
 }
 
 impl GeneveMeta {
+    /// Emit only the inner Geneve header.
     #[inline]
-    pub fn emit(&self, dst: &mut [u8]) {
-        debug_assert_eq!(dst.len(), self.hdr_len());
+    pub fn emit_inner(&self, dst: &mut [u8]) {
+        debug_assert_eq!(dst.len(), self.hdr_len_inner());
         let (base, remainder) = dst.split_at_mut(GeneveHdrRaw::SIZE);
         let mut raw = GeneveHdrRaw::new_mut(base).unwrap();
         raw.write(GeneveHdrRaw::from(self));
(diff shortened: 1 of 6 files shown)


rcgoodfellow:
It's a bit odd that `GeneveHdrRaw` contains the UDP header.

- https://github.com/oxidecomputer/opte/blob/d6959688ce7f44bfd2975cbebc141356a6a11c21/lib/opte/src/engine/geneve.rs#L232-L242

I would have expected that to be just

```
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |Ver|  Opt Len  |O|C|    Rsvd.  |          Protocol Type        |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |        Virtual Network Identifier (VNI)       |    Reserved   |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |                                                               |
   ~                    Variable-Length Options                    ~
   |                                                               |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```

per the definitions in RFC 8926. 

Maybe we can fix this in a more fundamental way by correcting this header definition?

FelixMcFelix:
I'm definitely in agreement: I'm not sure whether they were folded together *because* we're always parsing and emitting them as a pair. But I think the risk of a length mishap like the one fixed here is maybe more likely than parsing geneve without UDP? Let's split them then.

rcgoodfellow (approved):
Thanks for separating out the UDP/Geneve headers.

Issue #458, opened 2024-02-14
Suspicious amount of time spent in `bad_packet_probe`


mkeeter:
I was capturing kernel flamegraphs for a Crucible Downstairs in the Madrid mini-cluster, and noticed that we're spending ~2% of our runtime in `bad_packet_probe`:

<img width="1917" alt="Screenshot 2024-02-14 at 4 45 34 PM" src="https://github.com/oxidecomputer/opte/assets/745333/c0cbc103-d04f-486f-9f09-21f36d7e2e97">

(and another 1% doing other log formatting in `xde_rx`)

This seems like a lot to me!

Attaching to the probe for 10 seconds, I see that it's mostly `UnexpectedProtocol(TCP)` errors.
```
BRM42220081 # dtrace -q -n 'bad-packet { @[(string)arg0, (string)arg3] = count() } tick-10s { exit(0) }'

  unknown    Parse(BadHeader("BadLength { len: 60 }"))                        80
  unknown    Parse(BadHeader("ReadError { error: NotEnoughBytes }"))         120
  unknown    Parse(UnexpectedProtocol(TCP))                                 4308
```

In other runs, I also saw
```
  unknown    Parse(BadHeader("BadLength { len: 72 }"))                         1
  unknown    Parse(UnexpectedEtherType(6C00))                                  1
  unknown    Parse(UnexpectedProtocol(ICMPv6))                                 1
```

This is on BRM42220081 in cubby 14, which is the scrimlet for the Madrid mini-cluster.

FelixMcFelix:
Thanks a lot for digging this up! Repeating some of what we discussed earlier, this seems to be inbound underlay traffic on `cxgbe0` and `cxgbe1`. Ordinarily we expect to only see IPv6 + UDP + Geneve coming in, which is consistent with the parse errors there – OPTE is unhappy to see TCP, and I suspect it's trying to handle DDM solicits/advertisements as though they are actually Geneve (read: poorly).

Rather luckily, both `fmt` are right next to one another:

https://github.com/oxidecomputer/opte/blob/d6959688ce7f44bfd2975cbebc141356a6a11c21/xde/src/xde.rs#L2064-L2065

I'm keen to completely de-`String` this as far as SDTs are concerned. `ParseError::BadHeader(String)` is another bad apple but I need to actually think on that one, and it's not coming up for **every** non-OPTE underlay packet...

FelixMcFelix:
[...]

PR #459, opened 2024-02-15
Rework bad packet notification for dtrace probes

FelixMcFelix:
Today we're `fmt`ing the parse error encountered on any packet: non-geneve underlay, guest-originated, etc. Worse still, we're doing it twice. As shown in #458, this is really costly!

WIP. This currently includes the simplest fix -- don't `fmt` twice in a row. I'm working on getting us to zero fmts, with a *lot* of `&'static CStr` fun. This will necessarily complicate dtrace consumers a little, but I think we can alleviate it somewhat.

### DIFF
--- a/xde/src/xde.rs
+++ b/xde/src/xde.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-// Copyright 2023 Oxide Computer Company
+// Copyright 2024 Oxide Computer Company
 
 //! xde - A mac provider for OPTE.
 //!
@@ -159,13 +159,13 @@ fn bad_packet_probe(
     msg: &str,
 ) {
     let port_str = match port {
-        None => b"unknown\0" as *const u8 as *const i8,
-        Some(name) => name.as_ptr(),
+        None => c"unknown",
+        Some(name) => name.as_c_str(),
     };
     let msg_arg = CString::new(msg).unwrap();
     unsafe {
         __dtrace_probe_bad__packet(
-            port_str as uintptr_t,
+            port_str.as_ptr() as uintptr_t,
             dir as uintptr_t,
             mp as uintptr_t,
             msg_arg.as_ptr() as uintptr_t,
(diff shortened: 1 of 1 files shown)


mkeeter:
Looks good!  I don't see anyone using `fn bad_packet_parse_probe` any more, so you could probably delete it entirely.

FelixMcFelix:
I think the `&'static CStr` shenanigans are coming along pretty well:
```

**Comment by FelixMcFelix** [comment]
I think the `&'static CStr` shenanigans are working out really quite well:

```
PORT         DIR MBLK               MSG+DATA
unknown      IN  0xfffffe6a0e799420 Parse->UnexpectedEtherType [35020, 0]
unknown      IN  0xfffffe6a042c6480 Parse->UnexpectedProtocol [58, 0]
unknown      IN  0xfffffe6b8ab774c0 Parse->UnexpectedEtherType [35020, 0]
unknown      IN  0xfffffe6a1250e1a0 Parse->UnexpectedEtherType [35020, 0]
unknown      IN  0xfffffe6a0e7dbd60 Parse->UnexpectedEtherType [35020, 0]
unknown      IN  0xfffffe6a0e509da0 Parse->UnexpectedEtherType [35020, 0]
unknown      IN  0xfffffe69e9238700 Parse->UnexpectedEtherType [16724, 0]
unknown      IN  0xfffffe6b5e089c40 Parse->UnexpectedEtherType [16724, 0]
unknown      IN  0xfffffe6a0eedea60 Parse->BadHeader->BadLength { len: 4 } [0, 0]
```

`BadHeader::Parse` gets some special `unsafe` treatment until I can write a proc-macro to derive the new trait, but doing this without `format!()` is a huge win.

FelixMcFelix:
I spent some hours looking into fixing the source of `format!` calls in `Port::process` in addition to what's here. Finding it was pretty quick:
https://github.com/oxidecomputer/opte/blob/3a56d12230d3c301911703e51b2b3424511aacb3/lib/opte/src/engine/port.rs#L1425-L1463
So we're formatting a result for every packet, success or fail, then reallocing it.

As it turns out, fixing up `ProcessResult` and `ProcessError` is a deep rabbit hole, because we have a wide variety of error types being `String`ified, in addition to needing to fix up labels for Layers etc. into shared `str`/`CStr`s. I'm going to shunt that on as separate work since this is large enough as-is -- it and a hybrid `str`/`CStr` impl are currently in the `the-perf-cost-of-success` branch.

I don't know whether we want to gate this particular SDT behind `opte::engine::opte_debug` for now -- but this won't stop us paying for internal `format!()`s.

FelixMcFelix:
Offering some proof of where we're at with this now, I chucked an iPerf test over my local underlay and measured on the Rx path:
| Before | After |
|--------|--------|
| image | image |

Viewed as a flamegraph in each case (old, new):

rx-old
rx-new

Additionally, some FBT histograms show `port_process_return_probe` is costing about 3% of the post-change processing latency, not including any internal `format!()`s I can't get to yet.

rcgoodfellow (approved):
LGTM. Thanks!

PR #470, opened 2024-03-07
Fix xtask non-package install for fresh systems.

luqmana:
Using `cargo xtask install` without `--from-package` was relying on leftover state to actually work and would otherwise fail on clean system.


FelixMcFelix (approved):
Thanks very much for this Luqman, I figured I missed a step from Ry's feedback. LGTM!

Issue #460, opened 2024-02-15
Minimise `Debug::fmt` in XDE error handling


FelixMcFelix:
We currently have a few sets of culprits for calling `format!()` on errored packets:
 * `opte::engine::dbg`. There are places where we pass in a newly generated string to this function, however we gate the decision to *print* the string on a `static mut` switch. We need to move string formatting to fall after this check, otherwise we're always doing an expensive step.
 * Dtrace probes such as `__dtrace_probe_bad__packet` are making use of `format!()` to simplify error presentation, but these packets can potentially come from both the underlay and guest traffic.

```[tasklist]
### Tasks
- [x] #458 / #459
- [ ] De-`String` `Port::process` -- `ProcessResult` and `ProcessError` via `port_process_return_probe`.
```

PR #475, opened 2024-03-12
Use a `repr(C)` flow ID to reduce repeated per-packet conversion costs

### LINKED ISSUE #462: Conversion of `InnerFlowId` to SDT args appears expensive
We seem to be spending around 13% of `Port::process` by calling `flow_id_sdt_arg::from` for the benefit of its many dtrace probes -- many times per packet, as we have pre/post-processing and per-layer probes. We should either cache this, or find an adequate `#[repr(C)]` for InnerFlowId so that we don't *need* to perform a conversion.

image

FelixMcFelix:
This PR replaces `InnerFlowId` with an equivalent C-ABI friendly struct, from which we can directly pass pointers to our dtrace SDTs. This has proven necessary as we are reconstructing this many times per packet in both the UFT slowpath and fastpath.

Unfortunately, this has unearthed #476 for which we have a temporary hack, `NopDrop`, which uses a `black_box`'d `Drop` implementation to prevent problem SDTs from being tail-called into.

Additionally, this makes partial progress on #460 by using the new `DError` machinery whenever a packet is `Ok(Modified | Hairpin | Bypass)` to elide string formats on layer/port processing. The `Err(_)` and `Ok(Drop{ .. })` cases remain open work, but are less common and in those cases we have at least removed a superfluous realloc + memcpy on the formatted string.

From local IGB<->IGB results and comparing 4e2ad7e against master in CI, this seems like O(10%) reduction in packet latency *spent in XDE*. Certainly less of a total improvement once we account for the rest of MAC, and throughput impact on T6s is TBD.

Closes #462.


lib/opte/src/engine/flow_table.rs
```diff
@@ -229,16 +224,15 @@ fn flow_expired_probe(
     last_hit: Option<Moment>,
     now: Option<Moment>,
 ) {
+    let a = crate::NopDrop;
```
FelixMcFelix, line 227: All `NopDrop` code will be removed once illumos#16480 is merged in and available on the `helios-2.0` lab image.

lib/opte/src/engine/layer.rs
```diff
@@ -117,10 +119,15 @@ impl Display for DenyReason {
     }
 }
 
-#[derive(Debug)]
+// TODO: Represent `name` as joint C+RStr to implement fully.
```
FelixMcFelix, line 122: This would fall under the work remaining for #460.

lib/opte/src/lib.rs
```diff
@@ -170,6 +170,20 @@ mod opte_provider {
...
+///
+/// See: opte#476.
+pub struct NopDrop;
+
+impl Drop for NopDrop {
+    fn drop(&mut self) {
+        core::hint::black_box(self);
+    }
+}
+
```
FelixMcFelix, line 186: Again, to be removed once able.

dtrace/lib/common.d
```diff
@@ -2,14 +2,14 @@
 #pragma D depends_on provider ip
 
 typedef struct flow_id_sdt_arg {
-	int		af;
-	ipaddr_t	src_ip4;
-	ipaddr_t	dst_ip4;
-	in6_addr_t	src_ip6;
-	in6_addr_t	dst_ip6;
+	uint8_t		proto;
+	uint16_t		af;
```
mkeeter, line 6: Nit: is Github messed up alignment here, or is there an extra tab?

lib/opte/src/engine/port.rs
```diff
@@ -1460,18 +1478,22 @@ impl<N: NetworkImpl> Port<N> {
...
+                        flow_after,
                         epoch as uintptr_t,
                         pkt.mblk_addr(),
                         hp_pkt_ptr,
-                        res_arg.as_ptr() as uintptr_t,
+                        eb.as_ptr(),
                     );
                 }
 
+                drop(extra_str);
```
mkeeter, line 1496: Is this to prevent a tail-call?

lib/opte/src/engine/packet.rs
```diff
@@ -111,52 +112,102 @@ pub static FLOW_ID_DEFAULT: InnerFlowId = InnerFlowId {
...
+#[repr(C)]
 pub struct InnerFlowId {
-    pub proto: Protocol,
-    pub src_ip: IpAddr,
+    // Using a `u8` here for `proto` hides the enum repr from SDTs,
+    // but we need to have a `u16` discriminant for the IPs to set up
+    // 4B alignment on inaddr6_t on the dtrace side (so we lose out
+    // on the possibility of 38B packing).
+    pub proto: u8,
+    pub addrs: AddrPair,
```
luqmana, line 122: `AddrPair`'s discriminant is a `u16` right now, so there's 1B of padding between `proto` and `addrs` here. Feels like this comment is out of date?

Also, does the alignment requirement on the dtrace side matter since we copy to a local alloca anyways in `FLOW_FMT`?

dtrace/lib/common.d
```diff
@@ -2,14 +2,14 @@
...
+	union addrs {
+		ipaddr_t	ip4[2];
+		in6_addr_t	ip6[2];
```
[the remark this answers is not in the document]
luqmana: Also, `AddrPair` is `repr(u16)` and not `repr(C, u16)` which are slightly different in representation: a union of structs both starting with the discriminant vs a struct starting with the discriminant followed by a union of the payload.

https://rust-lang.github.io/unsafe-code-guidelines/layout/enums.html#explicit-repr-annotation-without-c-compatibility

https://rust-lang.github.io/unsafe-code-guidelines/layout/enums.html#explicit-repr-annotation-with-c-compatibility

lib/opte/src/engine/print.rs
```diff
@@ -207,9 +207,9 @@ pub fn print_lft_flow(
         t,
         "{}\t{}\t{}\t{}\t{}\t{}\t{}",
         flow_id.proto.to_string(),
-        flow_id.src_ip.to_string(),
+        flow_id.src_ip().to_string(),
```
luqmana, line 210: I know these `to_string()` are pre-existing but I'm not sure the above comment makes sense about custom `Display` implementations. `to_string()` would still use it

lib/opte/src/engine/port.rs
```diff
@@ -1441,17 +1443,33 @@ impl<N: NetworkImpl> Port<N> {
...
                 // for now this does the trick.
-                let res_str = match res {
-                    Ok(v) => format!("{:?}", v),
-                    Err(e) => format!("ERROR: {:?}", e),
+                let (eb, extra_str) = match res {
+                    Ok(v @ ProcessResult::Drop { reason }) => (
+                        ErrorBlock::from_err(v),
+                        Some(format!("{reason:?}\0"))
+                    ),
+                    Ok(v) => (ErrorBlock::from_err(v), None),
```
luqmana, line 1456: This is a bit confusing, we're in the `Ok` case but we're calling `ErrorBlock::from_err`?

lib/opte/src/engine/port.rs
FelixMcFelix: Good catch, changing that over.

dtrace/lib/common.d
FelixMcFelix: Yup, done.

lib/opte/src/engine/port.rs
FelixMcFelix: I think the type is maybe named confusingly at this point for what it can do  -- it was intended for `Err` values, but really any nested `enum` is what it's good at (and used for here). I'll rethink the naming.

dtrace/lib/common.d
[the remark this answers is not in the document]
FelixMcFelix: Thanks for bringing this up -- I did some extra checks there to verify which `repr` I meant. `repr(C, u16)` is the correct choice -- the non-C-compat is likely vestigial from an earlier iteration where I was storing something not 1B-aligned. `repr(C, u8)` would also minimise the struct size, if possible. Should be fixed now.

lib/opte/src/engine/packet.rs
FelixMcFelix: The comment is written very confusingly, my bad. Should be fixed now.

The alignment does unfortunately matter, unless we just use byte arrays for the actual v4/v6 addresses in `common.d`. Alignment is set [here](https://github.com/oxidecomputer/illumos-gate/blob/stlouis/usr/src/uts/common/netinet/in.h#L127). In `common.d` we have:
```
typedef struct flow_id_sdt_arg {
	uint8_t		proto;
	uint16_t	af;
	union addrs {
		struct {
			ipaddr_t src;
			ipaddr_t dst;
		} ip4;
		struct {
			in6_addr_t src;
			in6_addr_t dst;
		} ip6;
	} addrs;
	uint16_t	src_port;
	uint16_t	dst_port;
} flow_id_sdt_arg_t;
```
So `addrs` will be 4B-aligned due to `ip6`, and we get a bad read on the initial alloca.

lib/opte/src/engine/print.rs
FelixMcFelix: Yeah, this exact thing comes up in #480, wherre the compiler has realised this fact.

I've realised we accidentally lost the nicer `Protocol` in Display/print, so I'll fix that with the comments here.

lib/opte/src/engine/port.rs
FelixMcFelix: Okay, made some minor adjustments in 0edd7e6. If you think DError should also be renamed to suit its new multipurpose place, then we can do so.

luqmana (approved):
Thanks Kyle! Since the fix for #476 has landed in helios I'd say remove the workarounds but otherwise LGTM.

PR #477, opened 2024-03-19
Fix opte#469 for VPC-external flows

FelixMcFelix:
Removing an inbound UFT entry in response to a TCP closure requires
that we remove it using its pre-transform flow ID, rather than
`pkt.flow()`. This is what `TcpFlowEntryState.inbound_ufid` was intended
to capture. To-date, we have been filling in this optional inbound flow
ID on a TCP flow entry using its post-transformation metadata -- either
`pkt.flow()` or `InnerFlowId::from(pmeta)`. This results in a mismatch:
* The UFT is keyed on pre-transform flow ID, e.g., ephemeral IP + remote
  IP.
* The TCP flow table is always keyed on, e.g., private IP + remote IP.

So nothing is removed from the inbound UFT table when a TCP flow triggers
early UFT invalidation.

We now correctly use the initial `flow_before` on all inbound TCP flows.
Additionally, we create a new TCP flow entry and allow transformed packets
through in the event of a flow table mismatch in case similar bugs recur.

### DIFF
--- a/lib/opte/src/engine/port.rs
+++ b/lib/opte/src/engine/port.rs
@@ -1880,7 +1880,7 @@ impl<N: NetworkImpl> Port<N> {
 
         // Use the compiled UFT entry if one exists. Otherwise
         // fallback to layer processing.
-        match data.uft_in.get_mut(pkt.flow()) {
+        match data.uft_in.get_mut(ufid_in) {
             Some(entry) if entry.state().epoch == epoch => {
                 // TODO At the moment I'm holding the UFT locks not
                 // just for lookup, but for the entire duration of
@@ -1939,7 +1939,7 @@ impl<N: NetworkImpl> Port<N> {
                             // but we have regenerated the TCP entry to be less disruptive
                             // than a drop. Remove the UFT entry on the same proviso since the
                             // next packet to use it will regenerate it.
-                            data.uft_in.remove(ufid_in);
+                            self.uft_invalidate(data, None, Some(ufid_in), epoch);
                             return Ok(ProcessResult::Modified);
                         }
                         Err(ProcessError::TcpFlow(
--- a/lib/opte/src/engine/port.rs
+++ b/lib/opte/src/engine/port.rs
@@ -1655,7 +1655,7 @@ impl<N: NetworkImpl> Port<N> {
 
     // Process the TCP packet for the purposes of connection tracking
     // when an inbound UFT entry exists.
-    fn process_in_tcp_existing(
+    fn process_in_tcp(
         &self,
         data: &mut PortData,
         pmeta: &PacketMeta,
@@ -1685,9 +1685,10 @@ impl<N: NetworkImpl> Port<N> {
         ) {
             // We need to create a new TCP entry here because we can't call
             // `process_in_miss` on the already-modified packet.
-            e @ Err(ProcessError::TcpFlow(TcpFlowStateError::NewFlow {
-                ..
-            })) => {
+            e @ Err(
+                ProcessError::TcpFlow(TcpFlowStateError::NewFlow { .. })
+                | ProcessError::MissingFlow(_),
+            ) => {
                 self.create_new_tcp_entry(
                     &mut data.tcp_flows,
                     tcp,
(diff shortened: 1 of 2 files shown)


FelixMcFelix:
With sufficiently fast requests to `/v1/ping` to trigger port reuse, we are still firing counts of `__dtrace_probe_tcp__err`. This is sort of expected given the shortcuts for reviving dead flows that may not insert a cached `ufid_in`. However, the UFT entries are clean now, and there seem to be no `tcp_err` drops recorded:

```
kyle@farme:~/gits/opte$ kstat -m xde -n opte5
module: xde                             instance: 0
name:   opte5                           class:    net
        crtime                          350415.042037131
        in_bypass                       0
        in_drop                         13
        in_drop_handle_pkt              0
        in_drop_layer                   13
        in_drop_tcp_err                 0
        in_hairpin                      0
        in_modified                     2112
        in_process_err                  0
        in_uft_hit                      2076
        in_uft_miss                     49
        out_bypass                      0
        out_drop                        14
        out_drop_handle_pkt             8
        out_drop_layer                  6
        out_drop_tcp_err                0
        out_hairpin                     5
        out_modified                    1114
        out_process_err                 0
        out_uft_hit                     1094
        out_uft_miss                    39
        snaptime                        350793.425902187
```

By comparison, on `master` it remains quite easy to trigger a fairly large sum of `in_drop_tcp_err`s. Happy for other folks to try and replicate -- the simplest test on atrium (or other system that allows sub-1-minute port reuse) is to `curl http://<ip>/v1/ping --local-port 40000` a couple of times -- if any hang, we still have a problem.

rcgoodfellow:
It's occurred to me that I don't know how to update what omicron package picks up for OPTE before building and launching. Something equivalent to `./tools/update-dendrite.sh` or `./tools/update-maghemite.sh` in omicron. What is the best way to take this for a spin in a complete Omicron context?

lib/opte/src/engine/port.rs
```diff
@@ -1961,30 +1926,33 @@ impl<N: NetworkImpl> Port<N> {
...
                             );
-                            return Ok(ProcessResult::Drop {
-                                reason: DropReason::TcpErr,
-                            });
+                            // If we have a UFT but no TCP flow ID, there is likely a bug
+                            // and we are now out of sync. As above we can't reprocess,
+                            // but we have regenerated the TCP entry to be less disruptive
+                            // than a drop. Remove the UFT entry on the same proviso since the
+                            // next packet to use it will regenerate it.
+                            data.uft_in.remove(ufid_in);
```
luqmana, line 1942: Use `uft_invalidate` here so we also get the probe?

luqmana: Also, can't comment on it directly (grrr github context...), but mind using `ufid_in` instead of `pkt.flow()` on the match above for consistency.

FelixMcFelix:
> It's occurred to me that I don't know how to update what omicron package picks up for OPTE before building and launching. Something equivalent to `./tools/update-dendrite.sh` or `./tools/update-maghemite.sh` in omicron. What is the best way to take this for a spin in a complete Omicron context from install time?

If this were in main, then you'd edit `tools/opte_version`, install prerequisites using the script, and then edit Cargo.toml -> ["opte-ioctl", "oxide-vpc"].rev = `target git sha` before you `omicon-package package`.

I don't know how this impacts what goes into a4x2 -- if need be the [kmod is here](http://buildomat.eng.oxide.computer/public/file/oxidecomputer/opte/module/42041fd479f3199c5eaff39a6ff5ffcd29b92cd8/xde), and the [package is here](http://buildomat.eng.oxide.computer/public/file/oxidecomputer/opte/repo/42041fd479f3199c5eaff39a6ff5ffcd29b92cd8/opte.p5p). I think installing the package manually (`cargo xtask install --from-package --from-package-unfreze`) should suffice?

Since there are no API changes I typically just `cargo xtask install` between bringing up SoftNPU and `omicron-package install`.

luqmana (approved):
Nice catch! A couple small nits but otherwise lgtm

PR #480, opened 2024-03-28
General Repository Cleanup

FelixMcFelix:
This PR brings some code cleanliness changes and a bugfix:

* Make use of the new `addr_of_mut!()`/`addr_of!()` macros where shared mutable statics are used as part of MAC registration.
* Convert CStrs into `c""` syntax post-rust 1.77.0 (for non-bindgen code, at least).
* Fix up clippy lints which have piled up, most recently the 'redundant import' checks.
* Fix an issue where all TCP packets + sizes would be recorded on the 'inbound' stats.
* Prevent `derror_macro` from being pulled into Omicron's lockfile.

I spent a little time looking into the non-Send+Sync in Arc clippy lints, but those will require a separate PR to assess the safety of `impl Send+Sync` on, e.g., `KStatNamed`.


.github/buildomat/jobs/opte-api.sh
```diff
@@ -3,7 +3,7 @@
 #: name = "opte-api"
 #: variety = "basic"
 #: target = "helios-2.0"
-#: rust_toolchain = "nightly-2024-02-06"
+#: rust_toolchain = "nightly-2024-04-25"
```
mkeeter, line 6: Dumb question, but why doesn't `rust-toolchain.toml` work here?

lib/opte/src/d_error.rs
```diff
@@ -46,7 +35,7 @@ pub trait DError {
     fn leaf_data(&self, _data: &mut [u64]) {}
 }
 
-static_cstr!(EMPTY_STRING, b"\0");
+static_cstr!(EMPTY_STRING, c"");
```
mkeeter, line 38: Is it worth it to have a macro, now that it's the same as a one-liner?
```suggestion
static EMPTY_STRING: &CStr = c"";
```

FelixMcFelix: Probably not, particularly since it's only used for test cases.

.github/buildomat/jobs/opte-api.sh
FelixMcFelix: I believe this is down to us needing several nightly format options from `cargo fmt` across the repo. However, everything apart from XDE compiles on a stable toolchain version (as XDE is reliant on several unstable feature flags).

lib/opte/src/d_error.rs
FelixMcFelix: Done!

mkeeter (approved):

PR #484, opened 2024-04-04
Handle `mblk_t` packet chains

### LINKED ISSUE #237: Deal with packet chains
Currently OPTE does not deal at all with "packet chains". A packet chain is simply a chain of packets connected via `mblk_t.b_next`. Thanks for current limitations in mac I don't believe any chains will come via the Tx path (`xde_mc_tx()`), and there has been an assert there that has yet to trip to this day.

```
    // TODO I haven't dealt with chains, though I'm pretty sure it's
    // always just one.
    assert!((*mp_chain).b_next == ptr::null_mut());
    __dtrace_probe_tx(mp_chain as uintptr_t);
```

However, we should still have this support for the day that is no longer true. Furthermore, the Rx path has seen packet chains, and certainly is something we expect going forward.

FelixMcFelix:
This PR adds a `PacketChain` type which takes ownership of an `mblk_t` and then serves out owned `Packet<Initialized>` entitites for the rest of the stack to use.

This doesn't fundamentally change how we're structuring Packets internally, which still compute/allocate a Vec of `b_cont` entries. At the same time, we're still serving packets to our chosen destination MAC clients one by one. We will probably want to do some work here to better support reconstructing chains to hand off, where we can.

Closes #237.

### DIFF
--- a/lib/opte/src/engine/packet.rs
+++ b/lib/opte/src/engine/packet.rs
@@ -465,7 +465,7 @@ impl PacketChain {
             let tail_p = list.tail.as_ptr();
             unsafe {
                 (*tail_p).b_next = pkt_p;
-                (*pkt_p).b_prev = list_p;
+                (*pkt_p).b_prev = tail_p;
                 // pkt_p->b_next is already null.
             }
             list.tail = pkt;
--- a/lib/opte/src/engine/packet.rs
+++ b/lib/opte/src/engine/packet.rs
@@ -3929,6 +3929,10 @@ mod test {
         let chain_inner = chain.inner.as_ref().unwrap();
         assert_eq!(chain_inner.head.as_ptr(), els[1]);
         assert_eq!(chain_inner.tail.as_ptr(), els[2]);
+        unsafe {
+            assert!((*els[1]).b_prev.is_null());
+            assert!((*els[2]).b_next.is_null());
+        }
     }
 
     #[test]
--- a/lib/opte/src/engine/packet.rs
+++ b/lib/opte/src/engine/packet.rs
@@ -473,6 +473,13 @@ impl PacketChain {
             self.inner = Some(PacketChainInner { head: pkt, tail: pkt });
         }
     }
+
+    /// Return the head of the underlying `mblk_t` packet chain and
+    /// consume `self`. The caller of this function now owns the
+    /// `mblk_t` segment chain.
+    pub fn unwrap_mblk(mut self) -> Option<NonNull<mblk_t>> {
+        self.inner.take().map(|v| v.head)
+    }
 }
 
 impl Drop for PacketChain {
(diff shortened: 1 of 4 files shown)


lib/opte/src/engine/packet.rs
```diff
@@ -3740,4 +3884,88 @@ mod test {
...
+        assert_eq!(p0.mblk_addr(), els[0] as uintptr_t);
+        unsafe {
+            assert!((*els[0]).b_prev.is_null());
+            assert!((*els[0]).b_next.is_null());
+        }
+
+        // Chain head/tail ptrs are correct
+        let chain_inner = chain.inner.as_ref().unwrap();
+        assert_eq!(chain_inner.head.as_ptr(), els[1]);
+        assert_eq!(chain_inner.tail.as_ptr(), els[2]);
```
mkeeter, line 3931: Nit: we could also check that the head + tail aren't connected to anything (the tail won't have changed, but the head will have changed in the call to `next()`)
```suggestion
        assert_eq!(chain_inner.head.as_ptr(), els[1]);
        assert!((*els[1]).b_prev.is_null());
        assert_eq!(chain_inner.tail.as_ptr(), els[2]);
        assert!((*els[2]).b_next.is_null());
```

xde/src/xde.rs
```diff
@@ -1428,11 +1428,33 @@ unsafe extern "C" fn xde_mc_tx(
...
+    }
 
+    ptr::null_mut()
+}
+
+#[inline]
+unsafe fn xde_mc_tx_one(
+    src_dev: &XdeDev,
+    pkt: Packet<Initialized>,
+) -> *mut mblk_t {
```
mkeeter, line 1457: The comment below seems out of date, since `mp_chain` is owned by the `PacketChain` above

xde/src/xde.rs
```diff
@@ -2059,27 +2075,53 @@ unsafe extern "C" fn xde_rx(
     // dropped yet and thus our `MacClientHandle` is also still valid.
     let mch_ptr = arg as *const MacClientHandle;
     Arc::increment_strong_count(mch_ptr);
-    let mch = Arc::from_raw(mch_ptr);
+    let mch: Arc<MacClientHandle> = Arc::from_raw(mch_ptr);
```
mkeeter, line 2078: Vague vibes:

It hasn't changed in this PR, but it makes nervous to see `Arc::from_raw` paired with `Arc::as_ptr` (instead of `Arc::into_raw`).

I read through the `MacClientHandle::add_promisc` and think that it's correct as written.  However, I wonder if it would be cleaner to use `Arc::into_raw` and store a `*const MacClientHandle` in `MacPromiscHandle` (then rebuild the `Arc` in `Drop`), e.g. something like.

```rust
impl MacClientHandle {
    // ...

    /// Register promiscuous callback to receive packets on the underlying MAC.
    pub fn add_promisc(
        self: &Arc<Self>,
        ptype: mac_client_promisc_type_t,
        promisc_fn: mac_rx_fn,
        flags: u16,
    ) -> Result<MacPromiscHandle, c_int> {
        let mut mph = ptr::null_mut();

        // `MacPromiscHandle` keeps a reference to this `MacClientHandle`
        // in `mch` until it is removed and so we can safely access it from the
        // callback via the `arg` pointer.
        let mch = Arc::into_raw(self.clone());
        let ret = unsafe {
            mac_promisc_add(self.mch, ptype, promisc_fn, mch as *mut c_void, &mut mph, flags)
        };

        if ret == 0 {
            Ok(MacPromiscHandle { mph, mch })
        } else {
            Err(ret)
        }
    }
// ...
}

/// Safe wrapper around a `mac_promisc_handle_t`.
#[derive(Debug)]
pub struct MacPromiscHandle {
    /// The underlying `mac_promisc_handle_t`.
    mph: *mut mac_promisc_handle,

    /// The `MacClientHandle` used to create this promiscuous callback.
    mch: *const MacClientHandle,
}

impl Drop for MacPromiscHandle {
    fn drop(&mut self) {
        // Safety: We know that a `MacPromiscHandle` can only exist if a
        // mac promisc handle was successfully obtained, and thus `mph`
        // is valid.
        unsafe {
            mac_promisc_remove(self.mph);
            Arc::from_raw(self.mch); // dropped immediately
        };
    }
}
```

xde/src/xde.rs
```diff
@@ -2059,27 +2075,53 @@ unsafe extern "C" fn xde_rx(
...
-                return;
-            }
-        };
+    let mblk_addr = pkt.mblk_addr();
+    let mut pkt = match pkt.parse(Direction::In, parser) {
+        Ok(pkt) => pkt,
+        Err(e) => {
+            // TODO Add bad packet stat.
+            //
+            // NOTE: We are using mp_chain as read only here to get
```
mkeeter, line 2114: Outdated comment (`mp_chain` is gone)

xde/src/xde.rs
FelixMcFelix: I've tested this out and it seems to behave well, thanks for the suggestion!

lib/opte/src/engine/packet.rs
FelixMcFelix: Done.

xde/src/xde.rs
FelixMcFelix: Amended, should be slightly clearer now.

xde/src/xde.rs
FelixMcFelix: Amended.

mkeeter (approved):

PR #485, opened 2024-04-05
Prevent modunload once underlay has been set

FelixMcFelix:
This PR moves the underlay teardown logic from xde_detach into a new ioctl (and opteadm command) to unset the XDE underlay. Detach will now fail so long as the underlay is set -- this successfully prevents `modunload` from knocking us down after this state has been set.

The semantics are more or less what you'd expect:
* Clearing the underlay will fail if there exists a port.
* Clearing the underlay will fail if there is no underlay.
* Driver detach now requires that you `opteadm clear-xde-underlay`.

### DIFF
--- a/lib/opte-ioctl/src/lib.rs
+++ b/lib/opte-ioctl/src/lib.rs
@@ -4,6 +4,7 @@
 
 // Copyright 2022 Oxide Computer Company
 
+use opte::api::ClearXdeUnderlayReq;
 use opte::api::CmdOk;
 use opte::api::NoResp;
 use opte::api::OpteCmd;
@@ -168,6 +169,13 @@ impl OpteHdl {
     ) -> Result<NoResp, Error> {
         let req = SetXdeUnderlayReq { u1: u1.into(), u2: u2.into() };
         let cmd = OpteCmd::SetXdeUnderlay;
+        run_cmd_ioctl(self.device.as_raw_fd(), cmd, Some(&req))
+    }
+
+    /// Clear xde underlay devices.
+    pub fn clear_xde_underlay(&self) -> Result<NoResp, Error> {
+        let req = ClearXdeUnderlayReq { _unused: 0 };
+        let cmd = OpteCmd::ClearXdeUnderlay;
         run_cmd_ioctl(self.device.as_raw_fd(), cmd, Some(&req))
     }
 
--- a/xde/src/xde.rs
+++ b/xde/src/xde.rs
@@ -897,16 +897,18 @@ fn clear_xde_underlay() -> Result<NoResp, OpteError> {
         });
     }
 
+    // XXX: Need to safely quiesce xde_rx here.
+
     if let Some(underlay) = underlay.take() {
         // There shouldn't be anymore refs to the underlay given we checked for
         // 0 ports above.
-        let Ok(u1) = Arc::try_unwrap(underlay.u1) else {
+        let Some(u1) = Arc::into_inner(underlay.u1) else {
             return Err(OpteError::System {
                 errno: EBUSY,
                 msg: "underlay u1 has outstanding refs".into(),
             });
         };
-        let Ok(u2) = Arc::try_unwrap(underlay.u2) else {
+        let Some(u2) = Arc::into_inner(underlay.u2) else {
             return Err(OpteError::System {
                 errno: EBUSY,
                 msg: "underlay u2 has outstanding refs".into(),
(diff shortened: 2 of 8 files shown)


FelixMcFelix:
As rough proof of operation:
```
kyle@farme:~/gits/opte$ pfexec opteadm set-xde-underlay igb0 igb1
kyle@farme:~/gits/opte$ modinfo | grep xde
236 fffffffff44cb000 156450 290   1  xde (xde)

kyle@farme:~/gits/opte$ pfexec modunload -i 236
can't unload the module: Device busy

kyle@farme:~/gits/opte$ pfexec opteadm clear-xde-underlay
# succeeds

kyle@farme:~/gits/opte$ pfexec opteadm clear-xde-underlay
Error: command ClearXdeUnderlay failed: System { errno: 2, msg: "underlay not yet initialized" }

kyle@farme:~/gits/opte$ pfexec modunload -i 236
kyle@farme:~/gits/opte$
```

Naturally we need to change the virtual hardware teardown scripts in omicron's SoftNPU-based environment to call `opteadm clear-xde-underlay` once this is merged.

lib/opte-ioctl/src/lib.rs
```diff
@@ -171,6 +172,13 @@ impl OpteHdl {
         run_cmd_ioctl(self.device.as_raw_fd(), cmd, Some(&req))
     }
 
+    /// Clear xde underlay devices.
```
mkeeter, line 175: Vague architecture question: why is this code duplicated in `opte-ioctl` and `opteadm`?  It looks like `opteadm` already has a dependency on `opte-ioctl`; is there a subtle difference between `struct OpteAdm` and `struct OpteHdl`?

xde/src/xde.rs
```diff
@@ -856,6 +869,84 @@ fn set_xde_underlay(req: &SetXdeUnderlayReq) -> Result<NoResp, OpteError> {
...
+            //  3. `MacClientHandle` holds a ref to `MacHandle`.
+            // We explicitly drop them in order here to ensure there are no
+            // outstanding refs.
+
+            // 1. Remove promisc and unicast callbacks
+            drop(u.mph);
+            drop(u.muh);
+
+            // 2. Remove MAC client handle
+            if Arc::strong_count(&u.mch) > 1 {
```
mkeeter, line 921: This isn't obviously correct; there could be a race between checking `Arc::strong_count` and someone else cloning the `Arc` (maybe this race is impossible due to other factors, but it's not obvious here).

Using `Arc::into_inner` (and removing `drop(u.mch)` below) would be more obviously correct.

xde/src/xde.rs
```diff
@@ -856,6 +869,84 @@ fn set_xde_underlay(req: &SetXdeUnderlayReq) -> Result<NoResp, OpteError> {
...
+    if let Some(underlay) = underlay.take() {
+        // There shouldn't be anymore refs to the underlay given we checked for
+        // 0 ports above.
+        let Ok(u1) = Arc::try_unwrap(underlay.u1) else {
+            return Err(OpteError::System {
+                errno: EBUSY,
+                msg: "underlay u1 has outstanding refs".into(),
+            });
+        };
+        let Ok(u2) = Arc::try_unwrap(underlay.u2) else {
```
mkeeter, line 898: Same here

xde/src/xde.rs
FelixMcFelix: I don't think that this race is at all impossible -- `xde_rx` has another strong handle to the MAC client handle here, and it could be called into just before we deregister the promisc handle. 

That gives two bad orderings between {detach (today), underlay clear} and `xde_rx`, which makes me a little nervous if the two happen concurrently: we either bail on cleaning up our extra clients and leak holds on the underlying NIC, or `xde_rx` increments the strong count on an invalid `Arc` (!!!).

I think the only place this should be a risk is dev environments today, but I'm going to have a think about how we can safely quiesce here. Removing all ports is not a safe precondition, as any underlay traffic (or MAC flow matches, in future) will still land in `xde_rx`.

FelixMcFelix: I spent a little time double checking promisc handle behaviour in illumos, and it does give us the guarantees we need. I've left a comment explaining its behaviour for the benefit of future readers. 😄

lib/opte-ioctl/src/lib.rs
FelixMcFelix: Staring at it for a bit, I don't see that there *is* much reason for us to be doing this. We have a decent number of ioctls that appear in one but not the other. `OpteAdm` doesn't appear to have any methods which convert from CLI-specific datastructures, either: and if we did, there's no reason that `OpteAdm` couldn't just deref to an inner `OpteHdl`.

Opened #529.

mkeeter (approved):

PR #486, opened 2024-04-05
Expose choice of zone brand on kbench remote/server

FelixMcFelix:
This is a necessity for testing on newly-built bench gimlet images -- even if `sparse` is included, I've been unable to get their zones to come up due to supposed publisher/incorporation lookup issues.

On the other hand, a helios experiment-image built with extra packages like iperf will place all those into the omicron1 baseline where we need them, so this is the simplest fix to run both halves on sn9/14.

### DIFF
--- a/bench/Gimlet.adoc
+++ b/bench/Gimlet.adoc
@@ -5,38 +5,19 @@ Today, SN9/SN14 are set up in this configuration.
 
 == Image construction
 When https://github.com/oxidecomputer/helios[building a helios image], we need to include some extra packages and pull in the latest (host-installed) version of OPTE.
-The below patch adds these packages behind a conditional flag:
-
-[source, diff]
-----
-diff --git a/image/templates/gimlet/ramdisk-01-os.json b/image/templates/gimlet/ramdisk-01-os.json
-index a063784..c81974f 100644
---- a/image/templates/gimlet/ramdisk-01-os.json
-+++ b/image/templates/gimlet/ramdisk-01-os.json
-@@ -105,6 +105,14 @@
-             "/driver/developer/amd/zen"
-         ] },
-
-+        { "t": "pkg_install", "with": "netperf", "pkgs": [
-+            "/network/test/iperf",
-+            "/driver/network/opte",
-+            "/system/zones/brand/sparse",
-+            "/system/library/demangle",
-+            "/ooce/developer/flamegraph"
-+        ] },
-+
-         { "t": "pkg_install", "with": "compliance", "pkgs": [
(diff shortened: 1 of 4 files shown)


bench/Gimlet.adoc
```diff
@@ -0,0 +1,87 @@
+= Running OPTE Benchmarks on a Gimlet (pair)
+
+The kernel module benchmarks can be run on a pair of lab gimlets connected over their `cxgbe` interfaces.
+Today, SN9/SN14 are set up in this configuration.
+
+== Image construction
+When https://github.com/oxidecomputer/helios[building a helios image], we need to include some extra packages and pull in the latest (host-installed) version of OPTE.
+The below patch adds these packages behind a conditional flag:
```
luqmana, line 8: Is there a PR to just add this patch directly to helios? Alternatively seems like you could use the extra package functionality added in https://github.com/oxidecomputer/helios/pull/156.

bench/Gimlet.adoc
```diff
@@ -0,0 +1,87 @@
...
+index a063784..c81974f 100644
+--- a/image/templates/gimlet/ramdisk-01-os.json
++++ b/image/templates/gimlet/ramdisk-01-os.json
+@@ -105,6 +105,14 @@
+             "/driver/developer/amd/zen"
+         ] },
+
++        { "t": "pkg_install", "with": "netperf", "pkgs": [
++            "/network/test/iperf",
++            "/driver/network/opte",
```
luqmana, line 22: If we do go with adding a `netperf` feature, seems like you could omit `opte` here and just pass `-F optever=xx`.

bench/Gimlet.adoc
FelixMcFelix: Sure, I'll open a PR for it in the morning -- I didn't want to interfere with the canonical build config, but someone else will likely find it useful.

bench/Gimlet.adoc
FelixMcFelix: Sounds reasonable. I should also remove `sparse`, since we're using omicron1 here.

bench/Gimlet.adoc
FelixMcFelix: @luqmana I've just reworked the instructions in [1271e2e](https://github.com/oxidecomputer/opte/pull/486/commits/1271e2e1444948520210c995b5e13758db742a3a) to make use of `-F extra_packages+=x` instead, since it's only ~3 extra packages. I'll close oxidecomputer/helios#163.

luqmana (approved):

PR #490, opened 2024-04-09
VPC Subnet Routing

FelixMcFelix:
This PR introduces various features needed in OPTE side for VPC subnet routing:

* Add ioctl for deletion of router rules.
* Support for the 'system' vs 'custom' router distinction, and prioritisation of 'custom' rules of equal prefix length.
* Allow IP spoof-detection to be disabled (and re-enabled) on a per-CIDR basis. This is necassary to enable rules such as `192.168.0.0/16 -> ip(instance_abc)` for software routing or enabling VPN endpoints.

I'm leaving this in draft while the omicron-side pieces are in development.

### DIFF
--- a/bin/opteadm/src/bin/opteadm.rs
+++ b/bin/opteadm/src/bin/opteadm.rs
@@ -40,6 +40,7 @@ use oxide_vpc::api::PortInfo;
 use oxide_vpc::api::Ports;
 use oxide_vpc::api::ProtoFilter;
 use oxide_vpc::api::RemFwRuleReq;
+use oxide_vpc::api::RemoveCidrResp;
 use oxide_vpc::api::RouterClass;
 use oxide_vpc::api::RouterTarget;
 use oxide_vpc::api::SNat4Cfg;
@@ -223,12 +224,32 @@ enum Command {
 
     /// Configure external network addresses used by a port for VPC-external traffic.
     SetExternalIps {
-        /// The OPTE port to which the route is added
+        /// The OPTE port to configure
         #[arg(short)]
         port: String,
 
         #[command(flatten)]
         external_net: ExternalNetConfig,
+    },
+
+    /// Allows a guest to send and receive traffic on a given CIDR block.
+    AllowCidr {
+        /// The OPTE port to configure
+        #[arg(short)]
+        port: String,
+
+        /// The IP block to allow through the gateway.
+        prefix: IpCidr,
+    },
+
+    /// Prevents a guest from sending/receiving traffic on a given CIDR block.
+    RemoveCidr {
+        /// The OPTE port to configure
+        #[arg(short)]
+        port: String,
+
+        /// The IP block to deny at the gateway.
+        prefix: IpCidr,
     },
 }
 
@@ -741,6 +762,18 @@ fn main() -> anyhow::Result<()> {
             } else {
                 // TODO: show *actual* parse failure.
                 anyhow::bail!("expected IPv4 *or* IPv6 config.");
+            }
+        }
+
+        Command::AllowCidr { port, prefix } => {
+            hdl.allow_cidr(&port, prefix)?;
+        }
+
+        Command::RemoveCidr { port, prefix } => {
+            if let RemoveCidrResp::NotFound = hdl.remove_cidr(&port, prefix)? {
+                anyhow::bail!(
+                    "could not remove cidr {prefix} from gateway -- not found"
+                );
             }
         }
     }
--- a/bin/opteadm/src/bin/opteadm.rs
+++ b/bin/opteadm/src/bin/opteadm.rs
@@ -244,7 +244,7 @@ enum Command {
         external_net: ExternalNetConfig,
     },
 
-    /// Allows a guest to send and receive traffic on a given CIDR block.
+    /// Allows a guest to send or receive traffic on a given CIDR block.
     AllowCidr {
         /// The OPTE port to configure
         #[arg(short)]
@@ -252,6 +252,11 @@ enum Command {
 
         /// The IP block to allow through the gateway.
         prefix: IpCidr,
+
+        /// Control whether traffic on the CIDR should be allowed for
+        /// inbound or outbound traffic.
+        #[arg(long = "dir")]
+        direction: Direction,
     },
 
     /// Prevents a guest from sending/receiving traffic on a given CIDR block.
@@ -262,6 +267,11 @@ enum Command {
 
         /// The IP block to deny at the gateway.
         prefix: IpCidr,
+
+        /// Control whether traffic on the CIDR should be allowed for
+        /// inbound or outbound traffic.
+        #[arg(long = "dir")]
+        direction: Direction,
     },
 }
 
@@ -788,12 +798,14 @@ fn main() -> anyhow::Result<()> {
             }
         }
 
-        Command::AllowCidr { port, prefix } => {
-            hdl.allow_cidr(&port, prefix)?;
+        Command::AllowCidr { port, prefix, direction } => {
+            hdl.allow_cidr(&port, prefix, direction)?;
         }
 
-        Command::RemoveCidr { port, prefix } => {
-            if let RemoveCidrResp::NotFound = hdl.remove_cidr(&port, prefix)? {
+        Command::RemoveCidr { port, prefix, direction } => {
+            if let RemoveCidrResp::NotFound =
+                hdl.remove_cidr(&port, prefix, direction)?
+            {
                 anyhow::bail!(
                     "could not remove cidr {prefix} from gateway -- not found"
                 );
--- a/bin/opteadm/src/bin/opteadm.rs
+++ b/bin/opteadm/src/bin/opteadm.rs
@@ -26,6 +26,7 @@ use oxide_vpc::api::AddRouterEntryReq;
 use oxide_vpc::api::Address;
 use oxide_vpc::api::ClearVirt2BoundaryReq;
 use oxide_vpc::api::DelRouterEntryReq;
+use oxide_vpc::api::DelRouterEntryResp;
 use oxide_vpc::api::DhcpCfg;
 use oxide_vpc::api::ExternalIpCfg;
 use oxide_vpc::api::Filters as FirewallFilters;
@@ -711,7 +712,11 @@ fn main() -> anyhow::Result<()> {
         } => {
             let req =
                 DelRouterEntryReq { port_name: port, dest, target, class };
-            hdl.del_router_entry(&req)?;
+            if let DelRouterEntryResp::NotFound = hdl.del_router_entry(&req)? {
+                anyhow::bail!(
+                    "could not delete entry -- no matching rule found"
+                );
+            }
         }
 
         Command::SetExternalIps { port, external_net } => {
--- a/bin/opteadm/src/bin/opteadm.rs
+++ b/bin/opteadm/src/bin/opteadm.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-// Copyright 2023 Oxide Computer Company
+// Copyright 2024 Oxide Computer Company
 
 use clap::Args;
 use clap::Parser;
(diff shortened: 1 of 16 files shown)


lib/oxide-vpc/src/engine/gateway/mod.rs
```diff
@@ -275,3 +282,87 @@ impl Display for VpcMeta {
         write!(f, "vpc-meta")
     }
 }
+
+fn make_holepunch_rules(
```
zeeshanlakhani, line 286: n00b Q probably: would it make sense to implement this set of functions for `IpCidr` instead of as standalone functions? Also, maybe cidr.rs should be it's own module inside `gateway` (for clarity).

lib/oxide-vpc/src/engine/gateway/mod.rs
```diff
@@ -275,3 +282,87 @@ impl Display for VpcMeta {
...
+    guest_mac: MacAddr,
+    gateway_mac: MacAddr,
+    dest: IpCidr,
+    vpc_mappings: Arc<VpcMappings>,
+) -> (Rule<Finalized>, Rule<Finalized>) {
+    let vpc_meta = Arc::new(VpcMeta::new(vpc_mappings));
```
zeeshanlakhani, line 292: n00b2 Q: Would it make sense to implement `MetaAction` (used with `vpc_meta` / `Action::Meta` below) on vpc_mappings directly or impl it on `Arc<VpcMappings>` so you don't have to initialize the Arc here (again)?

zeeshanlakhani (commented):
Just some minor stuff while looking at this code in draft.

lib/oxide-vpc/src/engine/gateway/mod.rs
FelixMcFelix: Ordinarily, yes. There is a bit of a distinction here between `opte(-api)` and `oxide-vpc` though, in that the former is meant to be a platform for different VPC designs (see `NetworkImpl`). The previous code-owner (rpz) wrote a [really nice overview of this split](https://github.com/oxidecomputer/opte/blob/master/lib/opte/README.adoc).

The hole-punching behaviour is specific to our VPC design (and to work with how our gateway layer is structured), which is why I've placed it in the `oxide-vpc` crate -- sadly, it's freestanding because we can't impl on a foreign type. However, moving these to a submodule is very much doable! 😄

lib/oxide-vpc/src/engine/gateway/mod.rs
FelixMcFelix: The double-`Arc` is really unfortunate for sure. We could `impl MetaAction for VpcMappings`, but I think we'd need to recast the design of the `MetaAction` to use something additional like `RuleDisplay` instead of `Display`, I think?

lib/oxide-vpc/src/engine/gateway/mod.rs
FelixMcFelix: Reorganised into `gateway/transit.rs` in [2680fda](https://github.com/oxidecomputer/opte/pull/490/commits/2680fda4acefbce29009f2518c7f7a0ea8374764).

lib/oxide-vpc/src/engine/gateway/mod.rs
zeeshanlakhani: @FelixMcFelix ah interesting, I'll look further at this then.

lib/oxide-vpc/tests/integration_tests.rs
```diff
@@ -3981,3 +3998,227 @@ fn no_panic_on_flow_table_full() {
...
+        ]
+    );
+
+    // Encap routes between sleds correctly, inner IPs are not modified,
+    // and L2 dst matches the guest's NIC.
+    let v6_encap_meta = pkt1.meta().outer.ip.as_ref().unwrap().ip6().unwrap();
+    assert_eq!(v6_encap_meta.src, g1_cfg.phys_ip);
+    assert_eq!(v6_encap_meta.dst, g2_cfg.phys_ip);
+    assert_eq!(pkt1.meta().inner_ether().dst, g2_cfg.guest_mac);
+    assert_eq!(pkt1.meta().inner_ether().src, g1_cfg.guest_mac);
```
luqmana, line 4190: Hmm, was the inner mac here modified to be the gateway mac after `process(In, ..)` below?

bin/opteadm/src/bin/opteadm.rs
```diff
@@ -220,24 +224,45 @@ enum Command {
...
-        /// The OPTE port to which the route is added
+        /// The OPTE port to configure
         #[arg(short)]
         port: String,
 
         #[command(flatten)]
         external_net: ExternalNetConfig,
     },
+
+    /// Allows a guest to send and receive traffic on a given CIDR block.
```
luqmana, line 247: I can only really think of contrived scenarios and like you said this is already a bit more fine-grained than what the RFD mentions, but is it worth allowing just a given direction?

(not a blocker)

lib/oxide-vpc/tests/integration_tests.rs
luqmana: Ah yea, that's part of the `cidr_in` action.

lib/oxide-vpc/src/engine/gateway/transit.rs
```diff
@@ -0,0 +1,101 @@
...
+    vpc_mappings: Arc<VpcMappings>,
+) -> Result<NoResp, OpteError> {
+    let (in_rule, out_rule) = make_holepunch_rules(
+        port.mac_addr(),
+        port.network().cfg.gateway_mac,
+        dest,
+        vpc_mappings,
+    );
+    port.add_rule(NAME, Direction::In, in_rule)?;
+    port.add_rule(NAME, Direction::Out, out_rule)?;
```
luqmana, line 70: Should we delete the inbound rule if adding the outbound rule fails here?

I wouldn't be surprised if we have other places with similar issues. It might be time to think about a general transaction/rollback mechanism for adding rules.

lib/oxide-vpc/src/engine/gateway/transit.rs
```diff
@@ -0,0 +1,101 @@
...
+    let maybe_in_id = port.find_rule(NAME, Direction::In, &in_rule)?;
+    let maybe_out_id = port.find_rule(NAME, Direction::Out, &out_rule)?;
+    if let Some(id) = maybe_in_id {
+        port.remove_rule(NAME, Direction::In, id)?;
+    }
+    if let Some(id) = maybe_out_id {
+        port.remove_rule(NAME, Direction::Out, id)?;
+    }
+
+    Ok(if maybe_in_id.is_none() || maybe_out_id.is_none() {
```
luqmana, line 96: Given the current (intended) design we should never have only one or the other right? And similar to the add case above, if failing to remove/find the outbound, should we undo removing the inbound?

luqmana (commented):
Thanks for putting this together Kyle! The demo was great and excited to see this land. Looks good overall but one question about dealing with partial failures during add/remove.

bin/opteadm/src/bin/opteadm.rs
FelixMcFelix: That makes sense to me -- it's not a complex change and it lets us kick 'rule insert rollback' down the road for a little longer.

Done.

lib/oxide-vpc/src/engine/gateway/transit.rs
FelixMcFelix: Yeah, that would be more correct. Changing the API to set it one direction at a time will stop this from being an issue here, but instead push it up to the control plane. I'm happy to implement rollback on failure there.

I think that wider idea is very much worthwhile -- maybe a bit open-ended depending on whether we want this at the control-plane level, ioctl level, or both. I'll write up an issue.

lib/oxide-vpc/src/engine/gateway/transit.rs
FelixMcFelix: Yeah, agreed. The control plane doesn't call remove yet since these properties are only set on an OPTE port's creation, but your assumption would be correct going forward.

lib/oxide-vpc/src/engine/router.rs
```diff
@@ -116,54 +117,65 @@ impl fmt::Display for RouterTargetInternal {
...
 // ```
-fn prefix_len_to_priority(cidr: &IpCidr) -> u16 {
+fn compute_rule_priority(cidr: &IpCidr, class: RouterClass) -> u16 {
     use opte::api::ip::IpCidr::*;
     use opte::api::ip::Ipv4PrefixLen;
     use opte::api::ip::Ipv6PrefixLen;
     let (max_prefix_len, prefix_len) = match cidr {
         Ip4(ipv4) => (Ipv4PrefixLen::NETMASK_ALL.val(), ipv4.prefix_len()),
         Ip6(ipv6) => (Ipv6PrefixLen::NETMASK_ALL.val(), ipv6.prefix_len()),
     };
```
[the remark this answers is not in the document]
FelixMcFelix: That's far neater -- done!

FelixMcFelix (commented):
Thanks for the review Luqman!

lib/oxide-vpc/src/api.rs
```diff
@@ -899,6 +939,32 @@ impl Display for Ports {
...
+/// traffic on a specific CIDR other than its private IP.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct RemoveCidrReq {
+    pub port_name: String,
+    pub cidr: IpCidr,
+    pub dir: Direction,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub enum RemoveCidrResp {
```
zeeshanlakhani, line 961: this seems so option-like, so maybe it's worth keeping around info in the Ok case (maybe for info tracking) on what was removed?

lib/oxide-vpc/src/engine/router.rs
```diff
@@ -116,54 +117,60 @@ impl fmt::Display for RouterTargetInternal {
...
-// priroity = max_prefix_len - prefix len + 10
+// priority = (((max_prefix_len - prefix len) << 1) | is_system) + 10
 // ```
 //
 // `max_prefix_len` is the maximum prefix length for a given IP
 // CIDR type: `32` for IPv4, `128` for IPv6.
 //
 // `prefix_len` comes from the passed in `cidr` argument.
 //
+// One bit is used to ensure that 'custom' router rules take precedence
```
zeeshanlakhani, line 132: Q: is there a test that specifically tests precedence? It's probably already captured, but I thought to ask.

FelixMcFelix: Thanks for taking a look. The new system vs. custom route precedence is tested in the new `intra_subnet_routes_with_custom` test on an identical destination CIDR: https://github.com/oxidecomputer/opte/blob/789b927aed87d251b5e318431f373815f8f7dfc2/lib/oxide-vpc/tests/integration_tests.rs#L4064-L4090

I don't believe we have any explicit tests on e.g. the priority values we're using to get LPM-like behaviour, though.

lib/oxide-vpc/src/engine/gateway/transit.rs
FelixMcFelix: Opened #549 .

lib/oxide-vpc/src/api.rs
FelixMcFelix: Sure, that's a good idea. Done via 527d6ff.

bin/opteadm/src/bin/opteadm.rs
```diff
@@ -244,14 +244,19 @@ enum Command {
...
         #[arg(short)]
         port: String,
 
         /// The IP block to allow through the gateway.
         prefix: IpCidr,
+
+        /// Control whether traffic on the CIDR should be allowed for
+        /// inbound or outbound traffic.
+        #[arg(long = "dir")]
+        direction: Direction,
```
luqmana, line 259: Given `opteadm` is primarily a dev tool, I think making this an `Option<Direction>` and doing the ioctl calls for both directions if it's `None` is fine.

luqmana (approved):
One small suggestion for opteadm but otherwise lgtm! Thanks Kyle.

Issue #332, opened 2023-01-31
Feature request: the ability to delete Virt2Phys mappings


jmpesp:
OPTE currently supports `set_v2p` and `dump_v2p`, but it should support `delete_v2p` as well.

PR #494, opened 2024-04-12
add ability for omicron to observe v2p mappings

internet-diglett:
We're adding rpws to Omicron so we can keep the v2p mappings consistent across sled reboots and other unfortunate-but-necessary events. Currently, sled-agent nor omicron has the ability to determine what mappings already exists, so it's difficult to reconcile or detect conflicting mappings.

Related
---
Closes https://github.com/oxidecomputer/opte/issues/332

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -151,9 +151,9 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.15.4"
+version = "3.16.0"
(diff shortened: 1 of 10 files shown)


FelixMcFelix (approved):
Thanks Levon, this looks good! Thank you for also moving the V2B/V2P ioctl structs out from `overlay` and into `api`; having to reach through `engine` in a userland consumer would have been... odd. :)

PR #499, opened 2024-04-16
Cache routes received from `next_hop`

FelixMcFelix:
This PR inserts a caching layer in front of `next_hop` to store routes for a given `(IpAddr6, L4Hash)` for around 100ms, and reorganises routing-related code into `xde::route`. This is a necessary step as today we spend around 21% of `xde_mc_tx` in route lookup on a per-outbound-packet basis. As a consequence, we need install a new `Periodic` to flush expired entries every 1s, as otherwise we can have a large amountof detritus pileup from different L4Hash values.

Currently this is implemented as a per-port cache rather than shared across the driver -- I've done this because it will give us better sharding of concurrent readers/writers at higher port/guest counts. The PR as-is pulls us up from  ~1.74Gbps to 2Gbps on a single link between sn9/14.

I'm seeing similar gain on the `all-perf-in-flight` branch on in-review illumos bits (which includes mac flows + packet chains), where this change brings us from 2.25 Gbps to 2.50 Gbps on this setup.

### DIFF
--- a/xde/src/route.rs
+++ b/xde/src/route.rs
@@ -1,0 +1,500 @@
+use crate::ip;
+use crate::sys;
+use crate::xde::xde_underlay_port;
+use crate::xde::DropRef;
+use crate::xde::XdeDev;
+use alloc::collections::BTreeMap;
+use alloc::sync::Arc;
+use core::ptr;
+use core::time::Duration;
+use illumos_sys_hdrs::*;
+use opte::ddi::sync::KRwLock;
+use opte::ddi::sync::KRwLockType;
+use opte::ddi::time::Moment;
+use opte::engine::ether::EtherAddr;
+use opte::engine::ip6::Ipv6Addr;
+
+// XXX: completely arbitrary timeout.
+/// The duration a cached route remains valid for.
+const MAX_ROUTE_LIFETIME: Duration = Duration::from_millis(100);
+
+extern "C" {
+    pub fn __dtrace_probe_next__hop(
+        dst: uintptr_t,
+        gw: uintptr_t,
+        gw_ether_src: uintptr_t,
+        gw_ether_dst: uintptr_t,
+        msg: uintptr_t,
+    );
+}
+
+fn next_hop_probe(
+    dst: &Ipv6Addr,
+    gw: Option<&Ipv6Addr>,
+    gw_eth_src: EtherAddr,
+    gw_eth_dst: EtherAddr,
+    msg: &[u8],
+) {
+    let gw_bytes = gw.unwrap_or(&Ipv6Addr::from([0u8; 16])).bytes();
+
+    unsafe {
+        __dtrace_probe_next__hop(
+            dst.bytes().as_ptr() as uintptr_t,
+            gw_bytes.as_ptr() as uintptr_t,
+            gw_eth_src.to_bytes().as_ptr() as uintptr_t,
+            gw_eth_dst.to_bytes().as_ptr() as uintptr_t,
+            msg.as_ptr() as uintptr_t,
+        );
+    }
+}
+
+// The following are wrappers for reference drop functions used in XDE.
+
+fn ire_refrele(ire: *mut ip::ire_t) {
+    unsafe { ip::ire_refrele(ire) }
+}
+
+fn nce_refrele(ire: *mut ip::nce_t) {
+    unsafe { ip::nce_refrele(ire) }
+}
+
+fn netstack_rele(ns: *mut ip::netstack_t) {
+    unsafe { ip::netstack_rele(ns) }
+}
+
+// At this point the core engine of OPTE has delivered a Geneve
+// encapsulated guest Ethernet Frame (also simply referred to as "the
+// packet") to xde to be sent to the specific outer IPv6 destination
+// address. This packet includes the outer Ethernet Frame as well;
+// however, the outer frame's destination and source addresses are set
+// to zero. It is the job of this function to determine what those
+// values should be.
+//
+// Adjacent to xde is the native IPv6 stack along with its routing
+// table. This table is routinely updated to indicate the best path to
+// any given IPv6 destination that may be specified in the outer IP
+// header. As xde is not utilizing the native IPv6 stack to send out
+// the packet, but rather is handing it directly to the mac module, it
+// must somehow query the native routing table to determine which port
+// this packet should egress and fill in the outer frame accordingly.
+// This query is done via a private interface which allows a kernel
+// module outside of IP to query the routing table.
+//
+// This process happens in a sequence of steps described below.
+//
+// 1. With an IPv6 destination in hand we need to determine the next
+//    hop, also known as the gateway, for this address. That is, of
+//    our neighbors (in this case one of the two switches, which are
+//    also acting as routers), who should we forward this packet to in
+//    order for it to arrive at its destination? We get this
+//    information from the routing table, which contains Internet
+//    Routing Entries, or IREs. Specifically, we query the native IPv6
+//    routing table using the kernel function
+//    `ire_ftable_lookup_simple_v6()`. This function returns an
+//    `ire_t`, which includes the member `ire_u`, which contains the
+//    address of the gateway as `ire6_gateway_addr`.
+//
+// 2. We have the gateway IPv6 address; but in the world of the Oxide
+//    Network that is not enough to deliver the packet. In the Oxide
+//    Network the router (switch) is not a member of the host's
+//    network. Instead, we rely on link-local addresses to reach the
+//    switches. The lookup in step (1) gave us that link-local address
+//    of the gateway; now we need to figure out how to reach it. That
+//    requires consulting the routing table a second time: this time
+//    to find the IRE for the gateway's link-local address.
+//
+// 3. The IRE of the link-local address from step (2) allows us to
+//    determine which interface this traffic should traverse.
+//    Specifically it gives us access to the `ill_t` of the gateway's
+//    link-local address. This structure contains the IP Lower Level
+//    information. In particular it contains the `ill_phys_addr`
+//    which gives us the source MAC address for our outer frame.
+//
+// 4. The final piece of information to obtain is the destination MAC
+//    address. We have the link-local address of the switch port we
+//    want to send to. To get the MAC address of this port it must
+//    first be assumed that the host and its connected switches have
+//    performed NDP in order to learn each other's IPv6 addresses and
+//    corresponding MAC addresses. With that information in hand it is
+//    a matter of querying the kernel's Neighbor Cache Entry Table
+//    (NCE) for the mapping that belongs to our gateway's link-local
+//    address. This is done via the `nce_lookup_v6()` kernel function.
+//
+// With those four steps we have obtained the source and destination
+// MAC addresses and the packet can be sent to mac to be delivered to
+// the underlying NIC. However, the careful reader may find themselves
+// confused about how step (1) actually works.
+//
+//   If step (1) always returns a single gateway, then how do we
+//   actually utilize both NICs/switches?
+//
+// This is where a bit of knowledge about routing tables comes into
+// play along with our very own Delay Driven Multipath in-rack routing
+// protocol. You might imagine the IPv6 routing table on an Oxide Sled
+// looking something like this.
+//
+// Destination/Mask             Gateway                 Flags  If
+// ----------------          -------------------------  ----- ---------
+// default                   fe80::<sc1_p5>             UG     cxgbe0
+// default                   fe80::<sc1_p6>             UG     cxgbe1
+// fe80::/10                 fe80::<sc1_p5>             U      cxgbe0
+// fe80::/10                 fe80::<sc1_p6>             U      cxgbe1
+// fd00:<rack1_sled1>::/64   fe80::<sc1_p5>             U      cxgbe0
+// fd00:<rack1_sled1>::/64   fe80::<sc1_p6>             U      cxgbe1
+//
+// Let's say this host (sled1) wants to send a packet to sled2. Our
+// sled1 host lives on network `fd00:<rack1_sled1>::/64` while our
+// sled2 host lives on `fd00:<rack1_seld2>::/64` -- the key point
+// being they are two different networks and thus must be routed to
+// talk to each other. For sled1 to send this packet it will attempt
+// to look up destination `fd00:<rack1_sled2>::7777` (in this case
+// `7777` is the IP of sled2) in the routing table above. The routing
+// table will then perform a longest prefix match against the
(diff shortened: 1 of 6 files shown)


xde/src/route.rs
```diff
@@ -0,0 +1,514 @@
...
+            dst.bytes().as_ptr() as uintptr_t,
+            gw_bytes.as_ptr() as uintptr_t,
+            gw_eth_src.to_bytes().as_ptr() as uintptr_t,
+            gw_eth_dst.to_bytes().as_ptr() as uintptr_t,
+            msg.as_ptr() as uintptr_t,
+        );
+    }
+}
+
+// The following are wrappers for reference drop functions used in XDE.
```
mkeeter, line 51: It hasn't changed in this PR, but these functions seem iffy to me, since they make it trivial to trigger UB from safe Rust:

```rust
    let i = 123;
    let ptr = i as *const ip::ire_t;
    ire_refrele(ire); // probably bad!
```

xde/src/route.rs
```diff
@@ -0,0 +1,514 @@
...
+// default                   fe80::<sc1_p5>             UG     cxgbe0
+// default                   fe80::<sc1_p6>             UG     cxgbe1
+// fe80::/10                 fe80::<sc1_p5>             U      cxgbe0
+// fe80::/10                 fe80::<sc1_p6>             U      cxgbe1
+// fd00:<rack1_sled1>::/64   fe80::<sc1_p5>             U      cxgbe0
+// fd00:<rack1_sled1>::/64   fe80::<sc1_p6>             U      cxgbe1
+//
+// Let's say this host (sled1) wants to send a packet to sled2. Our
+// sled1 host lives on network `fd00:<rack1_sled1>::/64` while our
+// sled2 host lives on `fd00:<rack1_seld2>::/64` -- the key point
```
mkeeter, line 147:
```suggestion
// sled2 host lives on `fd00:<rack1_sled2>::/64` -- the key point
```

xde/src/route.rs
```diff
@@ -0,0 +1,514 @@
...
+
+        next_hop_probe(ip6_dst, Some(&gw_ip6), src, dst, b"\0");
+
+        Ok(Route { src, dst, underlay_dev })
+    }
+}
+
+/// A simple caching layer over `next_hop`.
+#[derive(Clone)]
+pub struct RouteCache(Arc<KRwLock<BTreeMap<RouteKey, CachedRoute>>>);
```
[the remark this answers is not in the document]
FelixMcFelix: I'll admit that I hadn't -- I was following the OPTE preference for `BTreeMap`s. But I threw something together quickly there in criterion with `RouteKey` and `CachedRoute` to insert/lookup one row into an arbitrary sized table:

```bash
   Compiling opte-bench v0.1.0 (/Users/kyle/gits/opte/bench)
    Finished bench [optimized + debuginfo] target(s) in 2.92s
     Running benches/userland.rs (target/release/deps/userland-57e28d11e88fdd6e)

mapquest/btreemap/get/0 time:   [287.02 ps 287.61 ps 288.30 ps]
Found 9 outliers among 100 measurements (9.00%)
  5 (5.00%) high mild
  4 (4.00%) high severe
mapquest/btreemap/get/2 time:   [3.8893 ns 3.9047 ns 3.9191 ns]
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
mapquest/btreemap/get/8 time:   [8.9178 ns 8.9306 ns 8.9465 ns]
Found 19 outliers among 100 measurements (19.00%)
  5 (5.00%) high mild
  14 (14.00%) high severe
mapquest/btreemap/get/64
                        time:   [8.1967 ns 8.2355 ns 8.2792 ns]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
mapquest/btreemap/get/512
                        time:   [7.2225 ns 7.2445 ns 7.2702 ns]
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high severe
mapquest/btreemap/get/8192
                        time:   [2.8060 ns 2.8106 ns 2.8155 ns]
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  4 (4.00%) high mild

mapquest/btreemap/insert/0
                        time:   [53.993 ns 54.513 ns 55.530 ns]
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe
mapquest/btreemap/insert/2
                        time:   [47.196 ns 47.271 ns 47.339 ns]
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild
mapquest/btreemap/insert/8
                        time:   [57.601 ns 57.679 ns 57.760 ns]
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild
mapquest/btreemap/insert/64
                        time:   [290.32 ns 290.57 ns 290.83 ns]
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) low mild
  4 (4.00%) high mild
mapquest/btreemap/insert/512
                        time:   [2.4069 µs 2.4108 µs 2.4145 µs]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
mapquest/btreemap/insert/8192
                        time:   [50.270 µs 50.590 µs 50.858 µs]




mapquest/hashmap/get/0  time:   [287.28 ps 288.20 ps 289.29 ps]
Found 17 outliers among 100 measurements (17.00%)
  5 (5.00%) high mild
  12 (12.00%) high severe
mapquest/hashmap/get/2  time:   [22.259 ns 23.395 ns 24.472 ns]
mapquest/hashmap/get/8  time:   [21.733 ns 22.932 ns 24.172 ns]
mapquest/hashmap/get/64 time:   [22.045 ns 23.297 ns 24.493 ns]
mapquest/hashmap/get/512
                        time:   [19.729 ns 20.846 ns 22.009 ns]
mapquest/hashmap/get/8192
                        time:   [19.731 ns 20.825 ns 21.964 ns]

mapquest/hashmap/insert/0
                        time:   [46.374 ns 46.581 ns 46.796 ns]
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
mapquest/hashmap/insert/2
                        time:   [43.214 ns 43.935 ns 44.742 ns]
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
mapquest/hashmap/insert/8
                        time:   [66.543 ns 66.619 ns 66.698 ns]
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) low mild
  3 (3.00%) high mild
mapquest/hashmap/insert/64
                        time:   [59.119 ns 60.854 ns 62.838 ns]
mapquest/hashmap/insert/512
                        time:   [295.78 ns 301.83 ns 308.87 ns]
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
mapquest/hashmap/insert/8192
                        time:   [231.58 ns 232.38 ns 233.19 ns]
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) low severe
  1 (1.00%) high mild
  1 (1.00%) high severe
```

This sort of tracks with my understanding of why we're using `BTreeMap`s in general throughout OPTE: faster lookup at the cost of more expensive insert.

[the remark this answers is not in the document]
FelixMcFelix: Perhaps a more obvious reason why we use `BTree{Map,Set}` that totally slipped my mind: there is no `HashMap` in `alloc`/`no_std`. So we would need to find and measure a suitable replacement if we wanted to go down that road. In the meantime I think the best call might be to limit the cache size.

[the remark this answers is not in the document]
rcgoodfellow: Sorry I don't have a ton of context here, but do we expect the nexthop cache to have a lot of regular churn?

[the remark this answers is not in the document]
FelixMcFelix: There shouldn't be in practice -- the above is very much worst-case, and I'm thinking about the case where a guest might have many active flows that are actively refreshing their routes and then add one (/n) more.

In the general case of a steady number of flows:
* An extant flow will call `next_hop` every 100ms in case preferences or reachability change. If there's an existing entry (even if expired), we should only be paying for lookup rather than a full insert on the `BTreeMap`.
* Expired routes are cleared out from the map every 1s.

The last part might add to the risk of churn; I'll change it up to expire flows which have expired by longer than the cleanup period to make refresh more consistent.

xde/src/route.rs
FelixMcFelix: I won't have to squeeze this in, but I've opened #515 to track this.

rcgoodfellow: I believe this was done to facilitate implementing `DropRef`. If there's a better/safer way to do this, all for it.

xde/src/route.rs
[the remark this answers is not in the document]
rcgoodfellow: My intuition here is that we want to optimize for queries rather than inserts. And if that intuition is not correct I really want to understand why. If this can be modeled as an LPM query, then using [poptrie](https://github.com/oxidecomputer/poptrie) might be a win.

xde/src/route.rs
```diff
@@ -0,0 +1,587 @@
...
+            Entry::Occupied(_) => true,
+            _ => false,
+        };
+
+        // We've had a definitive flow miss, but we need to cap the cache
+        // size to prevent excessive modification latencies at high flow
+        // counts.
+        // If full and we have no old entry to update, drop the lock and do
+        // not insert.
+        // XXX: Want to profile in future to see if LRU expiry is
```
rcgoodfellow, line 487: I think this is worth an issue and measuring cache performance there is 1) a pool of destination route keys larger than the cap that is b) churning within the expiry time. One could easily imagine this scenarios arising for a busy web server with lots of clients. I think the reason we have the L4 hash in the key is to pin flows to paths, so expressing this as an LPM to reduce table size would probably be tricky?

xde/src/route.rs
```diff
@@ -0,0 +1,587 @@
...
+    pub src: EtherAddr,
+    pub dst: EtherAddr,
+    pub underlay_idx: u8,
+    pub timestamp: Moment,
+}
+
+impl CachedRoute {
+    fn is_valid(&self, t: Moment) -> bool {
+        t.delta_as_millis(self.timestamp)
+            <= EXPIRE_ROUTE_LIFETIME.as_millis() as u64
```
rcgoodfellow, line 543: Should we detect rollover here with `try_into` or similar and return false if we detect rollover? Same for `is_retained`. Or if this is a smaller type, using `into` would make it more obvious that this is truncation/rollover safe.

xde/src/route.rs
```diff
@@ -0,0 +1,587 @@
...
+use opte::engine::ether::EtherAddr;
+use opte::engine::ip6::Ipv6Addr;
+
+// XXX: completely arbitrary timeouts.
+/// The duration a cached route remains valid for before it must be
+/// refreshed.
+///
+/// Expired routes will not be removed from the cache, and will leave
+/// an entry to enable a quick in-place refresh in the `BTreeMap`.
+const EXPIRE_ROUTE_LIFETIME: Duration = Duration::from_millis(100);
```
rcgoodfellow, line 30: For the moment, this is well within the DDM control plane neighbor expiry lifetime. We'll revisit when the DDM data plane lands.

rcgoodfellow (commented):
Thanks Kyle. Mostly high-level overall direction comments from my end. Code looks good to me, but let's look into the expiry rollover safety that I noted in the comments below.

xde/src/route.rs
```diff
@@ -0,0 +1,587 @@
...
+// packet") to xde to be sent to the specific outer IPv6 destination
+// address. This packet includes the outer Ethernet Frame as well;
+// however, the outer frame's destination and source addresses are set
+// to zero. It is the job of this function to determine what those
+// values should be.
+//
+// Adjacent to xde is the native IPv6 stack along with its routing
+// table. This table is routinely updated to indicate the best path to
+// any given IPv6 destination that may be specified in the outer IP
+// header. As xde is not utilizing the native IPv6 stack to send out
```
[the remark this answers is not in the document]
FelixMcFelix: This is part of what #504 is getting at; we do already need to revisit the tx path as part of getting the NICs back into hardware classification. Although, the way I've been going about that so far is by getting another client handle at the DLS layer, so it's not fundamentally different from `mac_tx`.

Coming at it from the IP level is probably safer still. It looks to me like the easiest way of doing this from an NCE/IRE (using public APIs) is `ip_xmit`, but this works on a packet at a time. There could be something worth copying in how it uses the ILL to perform the actual send?

xde/src/route.rs
```diff
@@ -0,0 +1,587 @@
...
+/// 2304 |@                                        14791
+/// 2560 |@@                                       32901
+/// 2816 |@                                        10730
+/// 3072 |                                         3459
+/// ```
+///
+/// Naturally, bringing this down to O(ns) is desirable. Usually, illumos
+/// holds `ire_t`s per `conn_t`, but we're aiming to be more fine-grained
+/// with DDM -- so we need a tradeoff between 'asking about the best route
+/// per-packet' and 'holding a route until it is expired'. We choose, for noe,
```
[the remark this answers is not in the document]
FelixMcFelix: Done.

xde/src/route.rs
FelixMcFelix: Done; std/core represents millisecond counds as `u128` while our `hrtime_t` is `i64` (nanoseconds). There isn't a  practical issue here given rollover times, but I've put `u128::from` on the LHS to make it obvious we're not doing any truncation.

xde/src/route.rs
FelixMcFelix: I'll add some extra probes around cache events so we can observe cases like this.

The L4 Hash in the key captures behaviour we have today, which is that OPTE uses multiple sled-to-sled paths depending on flow hash. We could recast that behaviour into an LPM of O(sleds) entries: if we can query and cache *all* reachability and preference information for each underlay destination (e.g., over both NIC links), then we can handle distribution over paths ourselves. We could possibly walk `ire->ire_bucket` in the same vein as `ire_round_robin` to fill that out?

I think that would be the right call; it'd be followup work that would likely replace large parts of this, so it's up to you on whether we use what we have for now and keep an LPM in mind as the next step (until we can offload, that is 😄).

rcgoodfellow: Let's take the win in front of us and continue to consider recasting things to fit in an LPM lookup later on.

rcgoodfellow (approved):

xde/src/route.rs
FelixMcFelix: Thanks again, #539 opened.

PR #62, opened 2022-03-14
use mac flows to filter xde traffic #61

rzezeski:
This work is on the back burner at the moment as there is more pressing work that can get done; sticking with promisc isn't a problem for the near term future.

PR #504, opened 2024-04-24
Move underlay NICs back into H/W Classification

### LINKED ISSUE #489: `set-xde-underlay` forces underlay devices into software classification
Part of our underlay initialisation is forcing existing MAC rings on underlay devices into software classification. This seems to be the case on both my home setup (`igb{0,1}`) and on a gimlet (`cxgbe{0,1}`).

On SN9, prior to setting the underlay:
```
> ::mac_ring
            ADDR TYPE STATE FLAG            GROUP CLASS              MIP              SRS FLOW NAME
fffffcfa18603008 RX   inuse 0000 fffffcfa196abc80 hw    fffffcfa3b8c3028 fffffcfa435be080 cxgbe1
fffffcfa186030e0 RX   inuse 0000 fffffcfa196abc80 hw    fffffcfa3b8c3028 fffffcfa43c6b280 cxgbe1
fffffcfa186031b8 RX   inuse 0000 fffffcfa196abc80 hw    fffffcfa3b8c3028 fffffcfa19dc0dc0 cxgbe1
fffffcfa18603290 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa18603368 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa18603440 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa18603518 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa186035f0 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa186036c8 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa186037a0 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa18603878 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa18603950 RX   inuse 0000 fffffcfa196abb40 hw    fffffcfa3b8c7020 fffffcfa19dc5200 cxgbe0
fffffcfa18603a28 RX   inuse 0000 fffffcfa196abb40 hw    fffffcfa3b8c7020 fffffcfa444edd80 cxgbe0
fffffcfa18603b00 RX   inuse 0000 fffffcfa196abb40 hw    fffffcfa3b8c7020 fffffcfa19dc68c0 cxgbe0
fffffcfa18603bd8 TX   inuse 0000 fffffcfa303aabe0 no    fffffcfa3b8cf018                0
fffffcfa18603cb0 RX   inuse 0000 fffffcfa19b3bb40 hw    fffffcfa3b8cf018 fffffcfa19dbe040 igb0
fffffcfa3d7e7880 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7958 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7a30 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7b08 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7be0 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7cb8 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7d90 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7e68 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
```

And immediately following a `opteadm set-xde-underlay cxgbe0 cxgbe1`:
```
> ::mac_ring
            ADDR TYPE STATE FLAG            GROUP CLASS              MIP              SRS FLOW NAME
fffffcfa18603008 RX   inuse 0000 fffffcfa196abc80 sw    fffffcfa3b8c3028                0
fffffcfa186030e0 RX   inuse 0000 fffffcfa196abc80 sw    fffffcfa3b8c3028                0
fffffcfa186031b8 RX   inuse 0000 fffffcfa196abc80 sw    fffffcfa3b8c3028                0
fffffcfa18603290 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa18603368 TX   inuse 0000 fffffcfa30213000 no    fffffcfa3b8c7020                0
fffffcfa18603440 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa18603518 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa186035f0 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa186036c8 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa186037a0 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa18603878 TX   inuse 0000 fffffcfa30213460 no    fffffcfa3b8c7020                0
fffffcfa18603950 RX   inuse 0000 fffffcfa196abb40 sw    fffffcfa3b8c7020                0
fffffcfa18603a28 RX   inuse 0000 fffffcfa196abb40 sw    fffffcfa3b8c7020                0
fffffcfa18603b00 RX   inuse 0000 fffffcfa196abb40 sw    fffffcfa3b8c7020                0
fffffcfa18603bd8 TX   inuse 0000 fffffcfa303aabe0 no    fffffcfa3b8cf018                0
fffffcfa18603cb0 RX   inuse 0000 fffffcfa19b3bb40 hw    fffffcfa3b8cf018 fffffcfa19dbe040 igb0
fffffcfa3d7e7880 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7958 TX   inuse 0000 fffffcfa38cadac0 no    fffffcfa3b8c3028                0
fffffcfa3d7e7a30 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7b08 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7be0 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7cb8 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7d90 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
fffffcfa3d7e7e68 TX   inuse 0000 fffffcfa38cadf20 no    fffffcfa3b8c3028                0
```
Although the flow names no longer point to cxgbe, the MIP values remain the same and we can verify these are our chosen mac_impls:
```
> fffffcfa3b8c3028::print mac_impl_t
{
    mi_rw_lock = {
        _opaque = [ 0 ]
    }
    mi_node = {
        list_next = 0
        list_prev = 0
    }
    mi_name = [ "cxgbe1" ]
    mi_state_flags = 0
    mi_driver = 0xfffffcfa3aa73b00
# ...
}
```

`ft_flow_count` on the main flow_tab is moving between 5 (no underlay) and 6 (underlay). I'll look into whether this is caused by either the fact that we're opening a MAC client (xdeu0 over cxgbe0, etc.) or the mac_unicast_add attached to that.

Currently, being stuck in `MAC_SW_CLASSIFIER` has the unfortunate side effect of forcing all packets received on a link to be processed individually rather than as a chain. Removing the underlay correctly returns the original clients to `MAC_HW_CLASSIFIER` as well as correctly linking up SRSs etc.

FelixMcFelix:
Today, we get our TX and RX pathways on underlay devices for XDE by creating a secondary MAC client on each device. As part of this process we must attach a unicast MAC address (or specify `MAC_OPEN_FLAGS_NO_UNICAST_ADDR`) during creation to spin up a valid datapath, otherwise we can receive packets on our promiscuous mode handler but any sent packets are immediately dropped by MAC. However, datapath setup then fails to supply a dedicated ring/group for the new client, and the device is reduced to pure software classification. This hard-disables any ring polling threads, and so all packet processing occurs in the interrupt context. This limits throughput and increases OPTE's blast radius on control plane/crucible traffic between sleds.

This PR places a hold onto the underlay NICs via `dls`, and makes use of `dls_open`/`dls_close` to acquire a valid transmit pathway onto the original (primary) MAC client, to which we can also attach a promiscuous callback. As desired, we are back in hardware classification:

```bash
EVT22200009 # mdb -k
Loading modules: [ unix genunix specfs dtrace mac cpu.generic apix cpc crypto mm random smbios zfs sata ip hook neti sockfs lofs vmm scsi_vhci arp ufs ipcc logindmux nsmb ptm ]
> ::mac_ring
            ADDR TYPE STATE FLAG            GROUP CLASS              MIP              SRS FLOW NAME
fffffcfa18602008 RX   inuse 0000 fffffcfa19b08dc0 hw    fffffcfa3b776028 fffffcfa19dc1dc0 cxgbe1
fffffcfa186020e0 RX   inuse 0000 fffffcfa19b08dc0 hw    fffffcfa3b776028 fffffcfa19dc78c0 cxgbe1
fffffcfa186021b8 RX   inuse 0000 fffffcfa19b08dc0 hw    fffffcfa3b776028 fffffcfa4cc69b40 cxgbe1
fffffcfa18602290 TX   inuse 0000 fffffcfa3b97d4a0 no    fffffcfa3b77e020                0
fffffcfa18602368 TX   inuse 0000 fffffcfa3b97d4a0 no    fffffcfa3b77e020                0
fffffcfa18602440 TX   inuse 0000 fffffcfa3b97d4a0 no    fffffcfa3b77e020                0
fffffcfa18602518 TX   inuse 0000 fffffcfa3b97d4a0 no    fffffcfa3b77e020                0
fffffcfa186025f0 TX   inuse 0000 fffffcfa3b97d4a0 no    fffffcfa3b77e020                0
fffffcfa186026c8 TX   inuse 0000 fffffcfa3b97d4a0 no    fffffcfa3b77e020                0
fffffcfa186027a0 TX   inuse 0000 fffffcfa3b97d4a0 no    fffffcfa3b77e020                0
fffffcfa18602878 TX   inuse 0000 fffffcfa3b97d4a0 no    fffffcfa3b77e020                0
fffffcfa18602950 RX   inuse 0000 fffffcfa1968f780 hw    fffffcfa3b77e020 fffffcfa4cc68480 cxgbe0
fffffcfa18602a28 RX   inuse 0000 fffffcfa1968f780 hw    fffffcfa3b77e020 fffffcfa445d0b00 cxgbe0
fffffcfa18602b00 RX   inuse 0000 fffffcfa1968f780 hw    fffffcfa3b77e020 fffffcfa19dbf040 cxgbe0
fffffcfa18602bd8 TX   inuse 0000 fffffcfa2fe62798 no    fffffcfa3b786018                0
fffffcfa18602cb0 RX   inuse 0000 fffffcf95805cbc0 hw    fffffcfa3b786018 fffffcfa43645240 igb0
fffffcfa3d7a0880 TX   inuse 0000 fffffcfa3b97d9e0 no    fffffcfa3b776028                0
fffffcfa3d7a0958 TX   inuse 0000 fffffcfa3b97d9e0 no    fffffcfa3b776028                0
fffffcfa3d7a0a30 TX   inuse 0000 fffffcfa3b97d9e0 no    fffffcfa3b776028                0
fffffcfa3d7a0b08 TX   inuse 0000 fffffcfa3b97d9e0 no    fffffcfa3b776028                0
fffffcfa3d7a0be0 TX   inuse 0000 fffffcfa3b97d9e0 no    fffffcfa3b776028                0
fffffcfa3d7a0cb8 TX   inuse 0000 fffffcfa3b97d9e0 no    fffffcfa3b776028                0
fffffcfa3d7a0d90 TX   inuse 0000 fffffcfa3b97d9e0 no    fffffcfa3b776028                0
fffffcfa3d7a0e68 TX   inuse 0000 fffffcfa3b97d9e0 no    fffffcfa3b776028                0
```

And if we zoom into the flamegraph, we now receive packets via `mac_rx_srs_poll_ring` (!!!).

rx

Performance numbers are [included below](https://github.com/oxidecomputer/opte/pull/504#issuecomment-2077621754) -- on its own, a 7% increase in guest throughput and a 74% increase in underlay max throughput.

This work is orthogonal to #62 (and related efforts) which will get us out of promiscuous mode -- both are necessary parts of making optimal use of the illumos networking stack.

Closes #489 .


FelixMcFelix:
Extra setup details on sn9/14, and better presentation of all known speeds:

```bash
# TCP tunables.
ipadm set-prop -p max_buf=8388608 tcp
ipadm set-prop -p send_buf=8388608 tcp
ipadm set-prop -p recv_buf=8388608 tcp
```

Global zone iperf numbers and status:
| Setup | Speed (Gbps), `iperf -c ... -P 1` | Works/Understood |
|--------|--------|--------|
| `master` | ~17.8--18.4 | ✅ |
| +#504 (+ hw_class, + poll thread) | ~21.2 | ✅ |
| +#62 (+ no promisc, - no poll thread) | ~25.8 | ❌ |
| No XDE | ~30.0--33.0 | ✅ |

FelixMcFelix:
I spent some time revisiting and fixing this up, and trying to understand the incompatibility with the flows work -- hacky as it is, I wouldn't have expected breakage.

There is a fairly consistent way to trigger a kernel panic, roughly in order of:
* Add a user-flow to an underlay link.
* Set the underlay.
* Pass traffic between ports across the underlay.
* Remove all ports and clear the underlay

I have some dumps to dig into on Monday, remarkably only the test harness is causing issues when we put the two features together (i.e., only once traffic is sent over do we get a panic).

xde/src/xde.rs
```diff
@@ -1074,60 +1075,66 @@ unsafe extern "C" fn xde_attach(
...
-            errno: EFAULT,
-            msg: format!("mac_promisc_add failed for {link_name}: {e}"),
-        })?;
-
-    // Set up a unicast callback. The MAC address here is a sentinel value with
-    // nothing real behind it. This is why we picked the zero value in the Oxide
-    // OUI space for virtual MACs. The reason this is being done is that illumos
-    // requires that if there is a single mac client on a link, that client must
-    // have an L2 address. This was not caught until recently, because this is
-    // only enforced as a debug assert in the kernel.
```
FelixMcFelix, line 1117: This should never be the case now: being visible via (and getting a stream from) DLS implies the existence of a client on the underlying MAC. This is also the root-cause of landing in software classification, as we end up with two clients.

xde/src/xde.rs
```diff
@@ -931,46 +925,26 @@ fn clear_xde_underlay() -> Result<NoResp, OpteError> {
...
-            //  1. `MacPromiscHandle` holds a ref to `MacClientHandle`, and
-            //  2. `MacUnicastHandle` holds a ref to `MacClientHandle`, and
-            //  3. `MacClientHandle` holds a ref to `MacHandle`.
-            // We explicitly drop them in order here to ensure there are no
-            // outstanding refs.
+            // We have a chain of refs here: `MacPromiscHandle` holds a ref to
+            // `DldStream`. We explicitly drop them in order here to ensure
+            // there are no outstanding refs.
 
             // 1. Remove promisc and unicast callbacks
```
luqmana, line 932: Nit: there aren't any unicast callbacks anymore right?

luqmana (commented):
Woo! 🎉 Thanks for cleaning this up and getting us to h/w classification Kyle! Overall looks good, just a couple comments

xde/src/dls/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+        let mut dlp = ptr::null_mut();
+        let mut dlh = ptr::null_mut();
+        let link = mph.link_id();
+
+        let res = unsafe { dls_devnet_hold(link.into(), &mut dlh) };
+        if res != 0 {
+            return Err(res);
+        }
+
+        let res = unsafe { dls_link_hold(dls_devnet_mac(dlh), &mut dlp) };
```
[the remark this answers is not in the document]
FelixMcFelix: That's what I was doing originally, until 3dc09de. The problem is that I was leaking temp refs, which made it so we could never tear down e.g. simnet links in the test suite -- so ended up replicating some of the internal logic to be safe.

I *think* we can use `dls_devnet_hold_link` if we change `DlsStream::drop` from `dls_devnet_rele`->`dls_devnet_rele_tmp`. I'll test it out.

xde/src/dls/mod.rs
```diff
@@ -0,0 +1,256 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Copyright 2024 Oxide Computer Company
+
+//! Safe abstractions around DLS public and private functions.
```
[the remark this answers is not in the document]
FelixMcFelix: I think the closest thing to a 'Committed' path today would be opening a full STREAMS session on the link (`dld_str_t` is the backend for this?), but then we'd be unable to register a promisc handle on that. I do want us to be in a place where we can send on whatever flow handle we get back from mac/dls.

xde/src/dls/mod.rs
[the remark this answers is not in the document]
FelixMcFelix: I've tested this and it does work, i.e. a cleared underlay will not leak references and we can remove the resource we're placed onto after we're done with XDE.

However, having a `dd_tref` held has the really nasty side effect of locking up `dladm` if we mistakenly attempt to delete an in-use simnet or VNIC. In contrast, having a full ref gets us the following message:
```
kyle@farme:~/gits/opte$ pfexec dladm delete-simnet xde_test_sim1
dladm: simnet deletion failed: link busy
```

So I think we should keep this as-is.

xde/src/dls/sys.rs
```diff
@@ -0,0 +1,189 @@
...
+pub enum dld_passivestate_t {
+    DLD_UNINITIALIZED,
+    DLD_PASSIVE,
+    DLD_ACTIVE,
+}
+
+// Direct translation of the illumos type type, with some
+// pointer fields left as opaque void*s for simplicity.
+#[repr(C)]
+pub struct dld_str_s {
```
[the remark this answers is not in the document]
FelixMcFelix: I think `dld_str_t` is just an opaque struct in `dls.h`, since they're usually allocated/constructed when coming in through `dld_open`. I'm not quite sure if there are other consequences for spinning up a full STREAMS session on the link though.

xde/src/dls/mod.rs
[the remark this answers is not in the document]
luqmana: Sure, maybe something to bring up in network sync tomorrow.

luqmana (approved):
Thanks Kyle, functionally looks good to me and the perf increases are great!

Unfortunately this does couple us even further to internal/private details so I would say maybe let's talk with the wider team tomorrow before merging.

FelixMcFelix:
> Unfortunately this does couple us even further to internal/private details so I would say maybe let's talk with the wider team tomorrow before merging.

Agreed, we should do. Thanks again for looking this over (and diving back into mac/dls/dls/...). 😄

FelixMcFelix:
Moving this PR back to draft to make the use of `dld_str_s` a bit safer. Talking through this with @rmustacc, it seems an expedient way to get this to a place where we need to be less wary of kernel changes is to:

1) Expose the kmem_cache_t used by dld_str, which is currently static, and alloc/destroy `dld_str_t`s from the kmem cache.
2) Verify the offsets needed for field accesses to dld_str_s match the current kernel.

1 is implemented here, and I'll open a minor stlouis patch shortly to enable this. 2 is a work-in-progress -- we could either have these exposed as a symbol by dls, or have opteadm install this information based on ctf information. Given that the correct cleanup for kmem cache use requires that we touch four such fields (rather than ~2 before), I'm inclined to go for the OPTE + CTF approach, and install the offsets during e.g. underlay set.

I've fixed up the base parser from ctf-bindgen to work correctly on `/kernel/drv/amd64/dld`. This code is temporarily housed under `crates/ctf` due to the extra admin work of enabling access to that repo.

FelixMcFelix:
I think this is good to go now, a quick recap on what's changed:
* We now use the functions added as part of https://github.com/oxidecomputer/illumos-gate/commit/0187c4284524db630ea6f878036b3cf180227943 to construct and destroy a valid `dld_str_t` with the least effort. Any cleanup which we were replicating from `dls`/`dld` after closing the stream has been moved into `dld_str_destroy_detached`.
  * This means we only access one field within `dld_str_t` -- 
* `opteadm` and the ioctl interface used by omicron now use `ctf-bindgen` to inspect `/system/object/dld/object` for a complete list of field offsets in `dld_str_t`. These are computed and sent when setting the underlay (i.e., this is a prerequisite for all other operations), wherein XDE will fail to load if the offset and type of `ds_mch` have changed.

CI is going to fail on `test`/`bench` until we get a new Helios image, but I can confirm that xde-tests and kbench behave correctly after a `pfexec pkg update` on my local box (as well as sn9/14).

lib/opte-ioctl/src/lib.rs
```diff
@@ -396,7 +396,7 @@ where
 }
 
 pub fn fetch_fragile_types() -> Result<FragileInternals, Error> {
-    let dld_ctf = Ctf::from_file("/kernel/drv/amd64/dld")
+    let dld_ctf = Ctf::from_file("/system/object/dld/object")
```
luqmana, line 379: 😮 TIL objfs

xde/src/dls/sys.rs
```diff
@@ -0,0 +1,198 @@
...
+#[repr(C)]
+pub enum dld_passivestate_t {
+    DLD_UNINITIALIZED,
+    DLD_PASSIVE,
+    DLD_ACTIVE,
+}
+
+// Direct translation of the illumos type, with some
+// pointer fields left as opaque void*s for simplicity.
+// Note that we only access ds_mch, and *do not create this
```
luqmana, line 114: Not to cast aside the verification machinery you added but if this is the only remaining field we need, should we also just add a `dld_str_mac_client_handle` function to stlouis so we can treat `dld_str_s` completely opaquely in `xde`?

lib/opte-ioctl/src/lib.rs
FelixMcFelix: I learned that trick from Robert! I think it'll be short-lived if we use your below suggestion. I'll preserve a copy of this branch since we could end up needing these tricks again (although I hope we do not).

xde/src/dls/sys.rs
FelixMcFelix: I think that's a good idea -- we may as well go all in and prevent the breaking API change, then.

FelixMcFelix: Done. The prior branch version is preserved as `runtime-ctf-check-poc` in case we need this functionality again.

PR #576, opened 2024-08-02
Fix cargo-check for non-illumos machines

pfmooney:


FelixMcFelix (approved):
Thanks for putting this together, this is really nice for ergonomics. The `ioctl` code is a lot more idiomatic now, too.

I was wondering whether we'd need to put in an early return in `opteadm` on non-illumos platforms, but the binary build fails when linking there (due to no kstat definitions) so that's not yet a concern.

pfmooney:
> I was wondering whether we'd need to put in an early return in `opteadm` on non-illumos platforms, but the binary build fails when linking there (due to no kstat definitions) so that's not yet a concern.

I can do some stubbing similar to what I did in `rusty-doors` for the kstat linkage, although as you point out, building an `opteadm` binary anywhere but illumos is not especially useful.

FelixMcFelix:
> > I was wondering whether we'd need to put in an early return in `opteadm` on non-illumos platforms, but the binary build fails when linking there (due to no kstat definitions) so that's not yet a concern.
> 
> I can do some stubbing similar to what I did in `rusty-doors` for the kstat linkage, although as you point out, building an `opteadm` binary anywhere but illumos is not especially useful.

Sorry, my comment was me thinking aloud more than anything else. I think this is fine without a stub change for now.

pfmooney:
> Sorry, my comment was me thinking aloud more than anything else. I think this is fine without a stub change for now.

No worries.  I meant it as a follow-up.

PR #588, opened 2024-08-29
Support for omicron internet gateway model

rcgoodfellow:
The Omicron internet gateway model associates IP pools, and subsequently addresses from those IP pools, with internet gateways. When packets are directed at an internet gateway through a VPC route, they should take on a source address from the IP pool that is tied to the target internet gateway. This trivially happens when there is only one internet gateway and an instance only has an address from one IP pool. But when there are multiple IP pools in play, particularly with floating IPs from different pools, a decision needs to be made about which one to use. That question is answered by the IP-pool/internet-gateway association in combination with VPC routes that direct packets at a particular internet gateway.

This PR provides OPTE machinery for realizing Omicron internet gateway model.

Leaving in draft as I put together an e2e prototype.

### DIFF
--- a/lib/oxide-vpc/tests/integration_tests.rs
+++ b/lib/oxide-vpc/tests/integration_tests.rs
@@ -67,8 +67,10 @@ use smoltcp::wire::NdiscNeighborFlags;
 use smoltcp::wire::NdiscRepr;
 use smoltcp::wire::NdiscRouterFlags;
 use smoltcp::wire::RawHardwareAddress;
+use std::collections::BTreeMap;
 use std::prelude::v1::*;
 use std::time::Duration;
+use uuid::Uuid;
 use zerocopy::AsBytes;
 
 const IP4_SZ: usize = EtherHdr::SIZE + Ipv4Hdr::BASE_SIZE;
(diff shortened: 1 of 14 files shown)


lib/oxide-vpc/tests/integration_tests.rs
```diff
@@ -1230,6 +1230,9 @@ fn check_external_ip_inbound_behaviour(
     }
 }
 
+// TODO(ry) not sure if this test still makes sense with igw being used for
```
rcgoodfellow, line 1233: Calling explicit attention to this.

FelixMcFelix (commented):
Thanks for putting this together Ry! I had read Internet Gateways as though they were a system-router-only concept for choosing SNAT, but being able to use them within custom routers to select source IPs is powerful.

However, I think that trying to use the router table to perform source IP selection is fraught for the reasons outlined in one of the inline comments. I think we might be able to preserve the priority/ECMP semantics we have today, without needing to throw sled-only routes in the mix:

1. Each InternetGateway route target stores the gateway's UUID into the packet's `ActionMeta`. We have one route in `router` per InternetGateway.
2. This PR relies on mapping each EIP to an Internet Gateway. We can instead add the parent gateway as a match predicate for each NAT/SNAT rule.
   - I suppose FIP sets would need to be split into multiple rules based on parent gateway.
3. Add low-priority route mapping `InternetGateway(None)` -> `Allow` (VPC-internal traffic).
4. Change default action to `Deny`.

But then we need to update those mappings at runtime. This would work well if external IP management was handled by an RPW, but it isn't yet. I don't know at first glance how to balance this -- you could carry the (external IP, parent gateway) mappings alongside the route updates? Let me know what you think.

PR #614, opened 2024-11-21
Tunneled LSO/CSO Support

### LINKED ISSUE #328: Performance: T6 Geneve aware TSO

### LINKED ISSUE #329: Performance: T6 Geneve aware checksumming

FelixMcFelix:
This PR makes use of in-progress illumos bits to detect common offloads between underlay devices, and advertises LSO/CSO capabilities on `opte` ports, relying illumos's emulation functionality to split packets and insert checksums if not on `cxgbe` (or other compatible NIC).

This allows for a few nice things:
 * Guests will send us TCP packets without an L4 checksum and up to ~64KiB in size.
   - The guest and OPTE spend less time computing/updating checksums.
   - OPTE has to process fewer packets – hooray! Ultimately, that's less processing time per payload byte.
   - If sending out over the NIC, then either the NIC or illumos's LSO emulation path are responsible for splitting the packet into TCP packets which will not violate the MTU. See below on what we tweak here.
   - If sending in the loopback path, we get to hand these packets directly to the target guest without splitting them apart. We do need to insert the checksum ourself, however.
 * Guests will send us UDP packets without an L4 checksum.
   - Again, the guest and OPTE spend less time computing/updating checksums.

Ultimately, we have control over the MSS we advertise to the NIC for LSO. The useful part of this is that when we know we are using a purely rack-internal path, we can elevate the MSS up to `MTU - overheads`. I've added a system of 'well-known' `ActionMeta` KV pairs which allow for layers like `overlay` to propagate this knowledge out. Given that the use of a larger MTU vastly reduces inbound packet rate on the receive half (the main bottleneck today), this gets us to around 8Gbps iPerf for rack-local traffic (and 17Gbps for two or more parallel streams) on `glasgow`.

On `dublin` with a full control plane, this resolves to around 4Gbps sled-to-sled and 14-Gbps sled-local between Linux VMs – we want to investigate the drop here once this is on dogfood. illumos as a guest doesn't do much better than before since it does not advertise GRO (illumos-as-host will fragment the packets) or LSO (it will never send TCP packets > MTU).

- [x] Cleanup.
- [x] Finalise illumos interfaces. [stlouis#663](https://code.oxide.computer/c/illumos-gate/+/485)
- [x] Manually fill checksums in guest loopback cases.
- [x] Have viona perform LSO if guests do not advertise/expect LRO. [illumos#17032](https://code.illumos.org/c/illumos-gate/+/3942)

Closes #328, closes #329.

Based on #688.


xde/src/mac/sys.rs
```diff
@@ -177,6 +185,148 @@ unsafe extern "C" {
...
+
+    pub lso_tunnel_tcp: lso_tunnel_tcp_t,
+}
+
+bitflags::bitflags! {
+/// Classes of TCP segmentation offload supported by a MAC provider.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct TcpLsoFlags: u32 {
+    /// The device supports TCP LSO over IPv4.
+    const BASIC_IPV4 = 1 << 0;
```
rcgoodfellow, line 258: breadcrumb to `LSO_TX_...` constants in illumos code would be helpful, same for `TunnelTcpLsoFlags` below.

xde/src/mac/sys.rs
```diff
@@ -177,6 +185,148 @@ unsafe extern "C" {
...
+        Self::INET_HDRCKSUM.bits();
+
+    /// The `cso_tunnel` field has been filled by the driver.
+    const TUNNEL_VALID = 1 << 5;
+}
+
+/// Classes of checksum offload suppported for tunnelled packets by a
+/// MAC provider.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct TunnelCsoFlags: u32 {
```
rcgoodfellow, line 303: breadcrumb for `MAC_CSO_TUN...` constants in illumos

xde/src/xde.rs
```diff
@@ -1600,6 +1652,34 @@ unsafe fn xde_mc_tx_one(src_dev: &XdeDev, mut pkt: MsgBlk) -> *mut mblk_t {
...
+        }
+        _ => (MacEtherOffloadFlags::empty(), IpProtocol(0)),
+    };
+
+    let ulp_meoi = mac_ether_offload_info_t {
+        meoi_flags: MacEtherOffloadFlags::L2INFO_SET
+            | MacEtherOffloadFlags::L3INFO_SET
+            | l4_flag,
+        meoi_len,
+        meoi_l2hlen: meta.inner_eth.packet_length() as u8,
```
rcgoodfellow, line 1675: taken literally this seems concerning, but i'm guessign `packet_lenght()` really means `header_lenght()`? The usize squashing is also a bit concerning.

xde/src/xde.rs
```diff
@@ -1652,15 +1732,57 @@ unsafe fn xde_mc_tx_one(src_dev: &XdeDev, mut pkt: MsgBlk) -> *mut mblk_t {
...
                 guest_loopback(src_dev, &devs, out_pkt, vni);
                 return ptr::null_mut();
             }
 
+            // Boost MSS to use full jumbo frames if we know our path
+            // can be served purely on internal links.
+            // Recall that SDU does not include L2 size, hence 'non_eth_payl'
+            let mut flags = offload_flags;
+            let inner_mtu = if mtu_unrestricted {
+                src_dev.underlay_capab.mtu - 70
```
rcgoodfellow, line 1750: named constants for 70 and 1500 would be nice

xde/src/xde.rs
```diff
@@ -1762,11 +1884,63 @@ where
...
+            let (v4_lso_max, v6_lso_max) = if upstream_lso
+                .lso_flags
+                .contains(TcpLsoFlags::BASIC_IPV4 | TcpLsoFlags::BASIC_IPV6)
+            {
+                (
+                    upstream_lso.lso_basic_tcp_ipv4.lso_max,
+                    upstream_lso.lso_basic_tcp_ipv6.lso_max,
+                )
+            } else {
+                (u16::MAX as u32, u16::MAX as u32)
```
rcgoodfellow, line 1928: nit: u32::from(u16::MAX)

xde/src/xde.rs
```diff
@@ -1931,6 +2105,9 @@ unsafe fn xde_rx_one(
         return;
     };
 
+    let is_tcp = matches!(meta.inner_ulp, ValidUlp::Tcp(_));
+    let mss_estimate = 1500 - (&meta.inner_l3, &meta.inner_ulp).packet_length();
```
rcgoodfellow, line 2109: named constant for standard MTU

xde/src/xde.rs
```diff
@@ -1948,11 +2125,25 @@ unsafe fn xde_rx_one(
...
-            let npkt = emit_spec.apply(pkt);
+        Ok(ProcessResult::Modified(emit_spec)) => {
+            let mut npkt = emit_spec.apply(pkt);
+
+            // Due to possible pseudo-GRO, we need to inform mac/viona on how
+            // it can split up this packet, if the guest cannot receive it
+            // (e.g., no GRO/large frame support).
+            // HW_LSO will cause viona to treat this packet as though it were
+            // a locally delivered segment making use of LSO.
+            if is_tcp && npkt.len() > 1500 + Ethernet::MINIMUM_LENGTH {
```
rcgoodfellow, line 2136: named constant for standard MTU

rcgoodfellow (commented):
Thanks Kyle!

As a high level note, there is quite a bit of `as` numeric conversion in this PR that we should probably work toward safer mechanisms.

FelixMcFelix (commented):
Thanks for the review Ry! Other fixes aside, there should be significantly less `as u32` in the newer parts of `xde`.

rcgoodfellow (approved):
LGTM. Thanks!

FelixMcFelix:
The zone-to-zone scenarios (`test`, `bench`) work on my main workstation after a `pfexec pkg update`. I'm going to merge this in spite of the CI failure here so that we can kick off image creation in one shot.

PR #621, opened 2024-11-28
Don't unwrap when removing invalid TCP state from fastpath.

### LINKED ISSUE #618: opte panic opte/src/engine/port.rs:1400:59 Option::unwrap() on a None value
Sled 21 in dogfood paniced in opte:

```
Dec 28 00:00:56 BRM42220031 savecore: [ID 570001 auth.error] reboot after panic: panicked at /work/oxidecomputer/opte/lib/opte/src/engine/port.rs:1400:59:                                                                                                                                  
Dec 28 00:00:56 BRM42220031 called `Option::unwrap()` on a `None` value 
```

The resulting core reports:
```
> ::status
debugging crash dump vmcore.0 (64-bit) from BRM42220031
operating system: 5.11 helios-2.0.23033 (oxide)
build version: heads/stlouis-0-g3a56ba873a
image uuid: f3d5155b-2b76-4a9f-9bed-124070e2687a
panic message: panicked at /work/oxidecomputer/opte/lib/opte/src/engine/port.rs:1400:59:
called `Option::unwrap()` on a `None` value
dump content: kernel pages and pages from PID -1
> 
```

And:
```
> ::stack                             
vpanic()
rust_begin_unwind+0x16d()
0xfffffffff84f3e8f()
core::panicking::panic_bounds_check::h675786dd45cde0b3()
0xfffffffff84f3398()
xde_mc_tx+0x4554()
mac_ring_tx+0x50(fffffcfbdfca3040, 0, fffffcfafeec62e0)
mac_provider_tx+0x6d(fffffcfbdfca3040, 0, fffffcfafeec62e0, fffffcfc072bfd78)
mac_tx+0x46c(fffffcfc072bfd78, fffffcfafeec62e0, 0, 1, 0)
viona_tx+0x322(fffffcfa706efdc0, fffffcfa706eff38)
viona_worker_tx+0x75(fffffcfa706eff38, fffffcfa706efdc0)
viona_worker+0x13a(fffffcfa706eff38)
thread_start+0xb()
```

### LINKED ISSUE #624: Large ioctl response resize logic is incorrect.
While stress-testing as part of the quest to reproduce #618 / #621, I have a few ports which hold vast numbers of firewall allows, UFT flows, and TCP flows. Accordingly I'm often getting messages like:

```bash
kyle@farme:~/gits/omicron$ pfexec opteadm dump-tcp-flows -p opte6
Error: failed to get response for command DumpTcpFlows in 3 attempts
kyle@farme:~/gits/omicron$ pfexec opteadm dump-layer firewall -p opte6
Error: failed to get response for command DumpLayer in 3 attempts
```

The code to handle this case is incorrectly performing the resize. https://github.com/oxidecomputer/opte/blob/e17f1ab3580a60cf11e0417c4e305f70a6774733/lib/opte-ioctl/src/lib.rs#L310-L322

If we throw in some debug logging, `eprintln!("OS wants to return {}, I have {}.", rioctl.resp_len_actual, resp_buf.capacity());`, we see that

```
OS wants to return 257396, I have 16384.
OS wants to return 257396, I have 241012.
OS wants to return 257396, I have 241012.
```

`resp_buf.len()` is zero, so the math is permanently off. Another piece is that on queries such as TCP flow state, response size can be a bit of a moving target as we rerun the query on a failure:

```
kyle@farme:~/gits/omicron$ pfexec opteadm dump-tcp-flows -p opte6
OS wants to return 58542, I have 16384.
OS wants to return 58546, I have 58542.
```
We'll want some extra headroom on top of the xde-given response length.

FelixMcFelix:
Now that the fastpath temporarily drops (and reacquires) the port lock when TCP state needs to be invalidated, we opened ourselves up to the possibility that another packet could have removed this state ahead of us. Equally, a packet could insert new TCP state which we might accidentally remove.

This PR removes the unwrap on removal to account for the race, and only removes TCP flows if they are pointer-equal.

Closes #618, closes #624.

### DIFF
--- a/lib/opte/src/engine/port.rs
+++ b/lib/opte/src/engine/port.rs
@@ -1396,9 +1396,9 @@ impl<N: NetworkImpl> Port<N> {
 
             let ufid_in = flow_lock.inbound_ufid.as_ref();
 
-            // Because we've dropped the port lock, another flow could have also
+            // Because we've dropped the port lock, another packet could have also
             // invalidated this flow and removed the entry. It could even install
-            // new UFT/TCP entries, depending on ordering.
+            // new UFT/TCP entries, depending on lock/process ordering.
             //
             // Verify that the state we want to remove still exists, and is
             // `Arc`-identical.
--- a/lib/opte/src/engine/port.rs
+++ b/lib/opte/src/engine/port.rs
@@ -1395,9 +1395,19 @@ impl<N: NetworkImpl> Port<N> {
             let ufid_out = &flow_lock.outbound_ufid;
 
             let ufid_in = flow_lock.inbound_ufid.as_ref();
-            self.uft_tcp_closed(&mut local_lock, ufid_out, ufid_in);
 
-            let _ = local_lock.tcp_flows.remove(ufid_out).unwrap();
+            // Because we've dropped the port lock, another flow could have also
+            // invalidated this flow and removed the entry. It could even install
+            // new UFT/TCP entries, depending on ordering.
+            //
+            // Verify that the state we want to remove still exists, and is
+            // `Arc`-identical.
+            if let Some(found_entry) = local_lock.tcp_flows.get(ufid_out) {
+                if Arc::ptr_eq(found_entry, &entry) {
+                    self.uft_tcp_closed(&mut local_lock, ufid_out, ufid_in);
+                    _ = local_lock.tcp_flows.remove(ufid_out);
+                }
+            }
 
             if reprocess {
                 lock = Some(local_lock);
--- a/lib/opte-ioctl/src/lib.rs
+++ b/lib/opte-ioctl/src/lib.rs
@@ -285,7 +285,8 @@ where
 
     // It would be a shame if the command failed and we didn't have enough bytes
     // to serialize the error response, so we set this to default to 16 KiB.
-    let mut resp_buf = Vec::with_capacity(16 * 1024);
+    const BASE_CAPACITY: usize = 16 * 1024;
+    let mut resp_buf = Vec::with_capacity(BASE_CAPACITY);
     let mut rioctl = OpteCmdIoctl {
         api_version: API_VERSION,
         cmd,
@@ -313,8 +314,18 @@ where
             if raw_err == libc::ENOBUFS {
                 assert!(rioctl.resp_len_actual != 0);
                 assert!(rioctl.resp_len_actual > resp_buf.capacity());
-                // Make room for the size the kernel claims to need
-                resp_buf.reserve(rioctl.resp_len_actual - resp_buf.capacity());
+
+                // Make room for at least the size the kernel claims to need.
+                // This can be slightly tricky: since every retry reruns the
+                // command, the size of the next resp could change from under
+                // us (increase or decrease). Keep some headroom to account
+                // for this.
+                let wanted_capacity =
+                    BASE_CAPACITY / 4 + rioctl.resp_len_actual;
+
+                // XDE could write into `resp_buf` (but does not).
+                // .len() *should* be zero -- but don't bank on it.
+                resp_buf.reserve(wanted_capacity - resp_buf.len());
                 rioctl.resp_bytes = resp_buf.as_mut_ptr();
                 rioctl.resp_len = resp_buf.capacity();
                 rioctl.resp_len_actual = 0;


FelixMcFelix:
I sadly haven't been able to locally reproduce the crash itself, using either zone-to-zone traffic or between local VMs via standalone omicron. Both this PR and `master` stand up against `while true; do iperf -c 10.0.0.1 -P128 -t 2; iperf -c 10.0.0.1 -P128 -t 2 -R; done` for around 10 minutes. This should slam the TCP flow state table pretty hard, and exposes us to a lot of leftover entries when data segments arrive later than expected -- I would have thought this would get us a trigger, .

rcgoodfellow:
> I sadly haven't been able to locally reproduce the crash itself, using either zone-to-zone traffic or between local VMs via standalone omicron. Both this PR and master stand up against while true; do iperf -c 10.0.0.1 -P128 -t 2; iperf -c 10.0.0.1 -P128 -t 2 -R; done for around 10 minutes.

Maybe try this for many flows simultaneously with distinct address pairs? This would get a bit closer to rack traffic conditions.

FelixMcFelix:
> Maybe try this for many flows simultaneously with distinct address pairs? This would get a bit closer to rack traffic conditions.

I've bumped up the topology a bit (all instances running omicron's builtin alpine linux):
```
┌──────────────┐    ┌──────────────────────────────────────────────────────┐
│   Macbook    │    │Farme                                                 │
│(10.0.23.168) │    │(10.0.147.187)                                        │
└──────────────┘    │  ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
        ▲           │   VPC                                             │  │
        └─────SSH───┼──┼─────────┬──────────────┬──────────────┐           │
                    │            │              │              │        │  │
                    │  │         ▼              ▼              ▼           │
                    │    ┌──────────────┬──────────────┬──────────────┐ │  │
                    │  │ │     cafe     │  restaurant  │     bar      │    │
                    │    │I: 172.30.0.5 │I: 192.168.0.5│I: 172.30.0.7 │ │  │
                    │  │ │E: 10.1.222.12│E: 10.1.222.13│E: 10.1.222.14│    │
                    │    └──────────────┴──────────────┴──────────────┘ │  │
                    │  │         ▲              │              │           │
                    │            │              │              │        │  │
                    │  │         └──────────────┴──────────────┘           │
                    │                  iperf3 client->server            │  │
                    │  └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
                    └──────────────────────────────────────────────────────┘
```

No crashes as yet on `master` under the above `-P128` rapid-fire iperf gauntlet, but I did come across #624. I've been attempting concurrent API operations (e.g., firewall rule add/delete) in case epoch bumps tell part of the story (as per oxide-dogfood on the most recent coredump). I'll leave it running, but we're missing something to make the original fault dead reproducible.

lib/opte/src/engine/port.rs
```diff
@@ -1395,9 +1395,19 @@ impl<N: NetworkImpl> Port<N> {
...
+            // new UFT/TCP entries, depending on ordering.
+            //
+            // Verify that the state we want to remove still exists, and is
+            // `Arc`-identical.
+            if let Some(found_entry) = local_lock.tcp_flows.get(ufid_out) {
+                if Arc::ptr_eq(found_entry, &entry) {
+                    self.uft_tcp_closed(&mut local_lock, ufid_out, ufid_in);
+                    _ = local_lock.tcp_flows.remove(ufid_out);
+                }
+            }
```
luqmana, line 1410: If it's ok to swap the order here, you could make use of the `Entry` API to lookup and remove the tcp flow to save on searching through it twice:


```suggestion
            if let Entry::Occupied(found_entry) = local_lock.tcp_flows.entry(*ufid_out) {
                if Arc::ptr_eq(found_entry.get(), &entry) {
                    _ = found_entry.remove_entry();
                    self.uft_tcp_closed(&mut local_lock, ufid_out, ufid_in);
                }
            }
```

luqmana (approved):
LGTM. Just left one suggestion for coalescing the flow state lookup+removal.

lib/opte/src/engine/port.rs
FelixMcFelix: Thanks a lot for the review Luqman.

I think this would be a good idea (the swap is valid as I read it), except `local_lock.tcp_flows` here is a `FlowTable` and not a `BTreeMap`. It'd be nice to forward the `Entry` API while respecting the capacity constraints in `FlowTable::add` etc. to enable that. I'll open a ticket.

PR #636, opened 2024-12-16
`RwLock::read` for UFT hits, kstats use `AtomicU64`

### LINKED ISSUE #435: Want less/finer-grained locking in the datapath
Currently, individual ports are wrapped in a `KMutex`, which is locked on:
1. Packet transmission
2. Packet receipt
3. Periodic UFT expiry
4. Reconfiguration of user-driven network state: Firewall rules, NAT/External IPs, DHCP Params, ...

Performance-wise, what I'm most concerned about is the total shared lock between 1. and 2., although 3. has caused minor issues (#426). It would be worth spending some time thinking about how this overlap of locks can be minimised overall -- keeping in mind that we do need pairwise flow installation to both In/Out tables, shared TCP state, etc.

FelixMcFelix:
This PR replaces the per-`Port` `KMutex` with a `KRwLock`. This allows
ioctls to read port state without preventing packets from being served.
This also prevents successful UFT hits from blocking one another on the
port for any length of time.

A requirement is that we move to atomic accesses on kstat values, which
allows us to update kstats from outside a write-lock on the port.

~~The odd thing is that there doesn't seem to be a performance
improvement (in Mbps).~~ Contention according to lockstat is solved, and
apparently the time spent in `xde_rx`/`xde_mc_tx` is reduced -- we're no
longer:

 * S2C -- the 3rd and 10th most spin-happy locks,
 * C2S -- the 2rd and 3rd most spin-happy locks.

After some refactoring, this has moved from 2.8->2.9 Gbps (pre-LSO) on `glasgow`.

Nor are we egregiously blocking on the port lock.

Closes #435.


crates/illumos-sys-hdrs/src/lib.rs
```diff
@@ -144,20 +148,48 @@ pub union KStatNamedValue {
...
+    }
+
+    #[inline]
+    pub fn incr_u64(&self, val: u64) {
+        core::hint::black_box(self.as_u64().fetch_add(val, Ordering::Relaxed));
+    }
+
+    #[inline]
+    pub fn decr_u64(&self, val: u64) {
+        core::hint::black_box(self.as_u64().fetch_sub(val, Ordering::Relaxed));
```
FelixMcFelix, line 192: There's something tricky going on here where `rustc` particularly enjoys optimising these increments and stores out. I tried plumbing snapshot reads all the way through via, e.g., a `#[no_mangle]` dummy ioctl in `xde` to convince the compiler that we do need these values, but that had no effect.

luqmana: I believe the issue is the transmute in `as_u64` is undefined behaviour. Pulling out a some of the code to run through miri gives:

```
error: Undefined Behavior: trying to retag from <2173> for SharedReadWrite permission at alloc1026[0x10], but that tag only grants SharedReadOnly permission for this location
  --> src/main.rs:37:18
   |
37 |         unsafe { std::mem::transmute::<&u64, &AtomicU64>(&self._u64) }
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |                  |
   |                  trying to retag from <2173> for SharedReadWrite permission at alloc1026[0x10], but that tag only grants SharedReadOnly permission for this location
   |                  this error occurs as part of retag at alloc1026[0x10..0x18]
```

Basically, we can't directly convert a `&u64` (which is `SharedReadOnly`) to `&AtomicU64` (which is `SharedReadWrite`). `UnsafeCell` is what we need here:


```diff
diff --git a/crates/illumos-sys-hdrs/src/lib.rs b/crates/illumos-sys-hdrs/src/lib.rs
index 24890e1..228e8e2 100644
--- a/crates/illumos-sys-hdrs/src/lib.rs
+++ b/crates/illumos-sys-hdrs/src/lib.rs
@@ -12,7 +12,7 @@ pub mod kernel;
 #[cfg(feature = "kernel")]
 pub use kernel::*;

-use core::mem::transmute;
+use core::cell::UnsafeCell;
 use core::ptr;
 use core::sync::atomic::AtomicU64;
 use core::sync::atomic::Ordering;
@@ -131,13 +131,15 @@ impl Default for kstat_named_t {
         Self {
             name: [0; KSTAT_STRLEN],
             dtype: 0,
-            value: KStatNamedValue { _c: [0; 16] },
+            value: KStatNamedValue(UnsafeCell::new(_KStatNamedValue {
+                _c: [0; 16],
+            })),
         }
     }
 }

 #[repr(C)]
-pub union KStatNamedValue {
+union _KStatNamedValue {
     _c: [c_char; 16],
     _i32: i32,
     _u32: u32,
@@ -145,6 +147,9 @@ pub union KStatNamedValue {
     _u64: u64,
 }

+#[repr(transparent)]
+pub struct KStatNamedValue(UnsafeCell<_KStatNamedValue>);
+
 impl core::ops::AddAssign<u64> for KStatNamedValue {
     #[inline]
     fn add_assign(&mut self, other: u64) {
@@ -162,34 +167,34 @@ impl core::ops::SubAssign<u64> for KStatNamedValue {
 impl KStatNamedValue {
     /// Validates at compile time whether `self._u64` can be safely used as
     /// an AtomicU64.
-    const KSTAT_ATOMIC_U64_SAFE: () = if align_of::<KStatNamedValue>() % 8 == 0
-    {
-    } else {
-        panic!("Platform does not meet u64 8B alignment for AtomicU64");
-    };
+    const KSTAT_ATOMIC_U64_SAFE: () =
+        if align_of::<KStatNamedValue>() % align_of::<AtomicU64>() == 0 {
+        } else {
+            panic!("Platform does not meet u64 8B alignment for AtomicU64");
+        };

     #[inline(always)]
     #[allow(clippy::let_unit_value)]
-    pub fn as_u64(&self) -> &AtomicU64 {
+    fn as_u64(&self) -> &AtomicU64 {
         _ = Self::KSTAT_ATOMIC_U64_SAFE;
         // SAFETY: KStatNamedValue must have 8B alignment on target platform.
         //         Validated by compile time check in `KSTAT_ATOMIC_U64_SAFE`.
-        unsafe { transmute::<&u64, &AtomicU64>(&self._u64) }
+        unsafe { AtomicU64::from_ptr(self.0.get().cast()) }
     }

     #[inline]
     pub fn set_u64(&self, val: u64) {
-        core::hint::black_box(self.as_u64().store(val, Ordering::Relaxed));
+        self.as_u64().store(val, Ordering::Relaxed);
     }

     #[inline]
     pub fn incr_u64(&self, val: u64) {
-        core::hint::black_box(self.as_u64().fetch_add(val, Ordering::Relaxed));
+        self.as_u64().fetch_add(val, Ordering::Relaxed);
     }

     #[inline]
     pub fn decr_u64(&self, val: u64) {
-        core::hint::black_box(self.as_u64().fetch_sub(val, Ordering::Relaxed));
+        self.as_u64().fetch_sub(val, Ordering::Relaxed);
     }
 }
```

With that diff the stats updates don't seem to be optimized out.

FelixMcFelix: Yeah, that's it exactly. Thank you for figuring it out!

crates/illumos-sys-hdrs/src/lib.rs
```diff
@@ -127,37 +131,70 @@ impl Default for kstat_named_t {
...
+    } else {
+        panic!("Platform does not meet u64 8B alignment for AtomicU64");
+    };
+
+    #[inline(always)]
+    #[allow(clippy::let_unit_value)]
+    fn as_u64(&self) -> &AtomicU64 {
+        _ = Self::KSTAT_ATOMIC_U64_SAFE;
+        // SAFETY: KStatNamedValue must have 8B alignment on target platform.
+        //         Validated by compile time check in `KSTAT_ATOMIC_U64_SAFE`.}
```
luqmana, line 181: Nit, stray `}` appeared:
```suggestion
        //         Validated by compile time check in `KSTAT_ATOMIC_U64_SAFE`.
```

luqmana (approved):
LGTM, thanks!

crates/illumos-sys-hdrs/src/lib.rs
FelixMcFelix: Done in 587082d.

PR #639, opened 2024-12-18
Use `BTreeMap`s to accelerate port lookup by name & VNI/MAC address

### LINKED ISSUE #637: Inbound packets require a linear scan of all ports to deliver packets
In the packet receive path, we perform a linear scan over all ports to find a matching VNI + destination MAC address for every packet.

https://github.com/oxidecomputer/opte/blob/9b57c4bbb761cd1d347b9a912ae865324ffe3c78/xde/src/xde.rs#L1445-L1447

This does have a significant performance impact for the last port in the list, as measured on `glasgow`:

| $n$ additional ports | iPerf3 throughput |
|--------|--------|
| 0 | ~3.00 Gbps |
| 32 | ~2.82 Gbps |
| 64 | ~2.72 Gbps |
| 256 | ~2.36 Gbps |
| 512 | ~1.90 Gbps |
| 1024 | ~1.22 Gbps |

While 256 ports is maybe a bit on the extreme end today, port counts in the range 64–128 are not unreasonable. In theory this number can grow fairly large – one for each service, ≥1 for each instance (max 128 instances today, I believe). On higher density Turin parts we're looking at more threads (384+), so more ports is likely when O(instances) ≈ O(threads).

We should use a `BTreeMap`, presort the `Vec<Box<XdeDev>>` on `(Vni, MacAddr)`, or keep an adjacent lookup table. [`slice::binary_search_by_key`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search_by_key) will be key for the latter – which is best will require some benchmarking, since we have varying levels of indirection here. We could do the same for port names (add/removal requires a linear scan here too), but this is not a hotpath.

### LINKED ISSUE #638: IOCTL hangs when adding >= 512 OPTE ports.
While attempting to grab some numbers on performance at 512/1024+ ports for #637, the `cargo kbench` runner froze up. Attempts to contact the driver via `opteadm` hit the same fate.

In `mdb -k`:
```bash
> ::stacks

...

fffffcfa2e6e4480 SLEEP    CV                      1
                 swtch+0x139
                 cv_wait+0x70
                 vmem_nextfit_alloc+0x11e
                 vmem_alloc+0x19e
                 id_alloc+0x1b
                 periodic_create+0x36
                 i_timeout+0x5d
                 ddi_periodic_add+0x53
                 create_xde+0x145f
                 xde_ioc_opte_cmd+0xa0ec
                 xde_ioctl+0x7e
                 cdev_ioctl+0x3f
                 spec_ioctl+0x55
                 fop_ioctl+0x40
                 ioctl+0x144

...

```

It looks like we've hit some kind of limit when trying to attach another periodic task while creating the port. We need to know what this limit is and probably curtail our periodic task generation.

FelixMcFelix:
This PR moves all lookup of ports by data- and control-plane operations to make use of `BTreeMap`s and prevent linear scans from being used in any case. `cargo kbench` now supports the `-P` flag to spin up a number of ports with random MAC/VNI combinations.

On `glasgow`, using `./kbench remote -b omicron1 -u cxgbe0 cxgbe1 172.20.2.114 -p -P $N_PORTS` + `iperf -c 10.0.0.1 -R`, the measured throughput remains in the 2.85–3.0Gbps range regardless of choice of `-P` (up to 4096). At 16384 ports we could no longer consistently hit 3Gbps, but we hadn't dipped too badly (2.8–2.9Gbps). I haven't measured control-plane-op/remove times, but I would expect that they enjoy similar speedup at large port counts. E.g., firewall/router/IP state updates all require a name lookup to apply. Addition of ports will now be more expensive (assuming no realloc of the `Vec`), but this is an expected part of maintaining `BTreeMap`s and the tradeoff is worth it for the hotpath benefit.

Additionally, we now use a single periodic task for all flow/route expiry management, to prevent ourselves from spending the system's `ddi_periodic_add` budget. This should be no worse than the previous state of affairs – iterating over all ports requires only a read lock over the `DevMap`, which any Rx handler will already be doing.

Closes #637, closes #638.

### DIFF
--- a/xde/src/dev_map.rs
+++ b/xde/src/dev_map.rs
@@ -1,0 +1,76 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Copyright 2024 Oxide Computer Company
+
+use crate::xde::XdeDev;
+use alloc::boxed::Box;
+use alloc::collections::btree_map::BTreeMap;
+use alloc::string::String;
+use opte::api::MacAddr;
+use opte::api::Vni;
+
+// From microbenchmarking (https://github.com/oxidecomputer/opte/issues/637)
+// it is apparent that we have *far* faster `Ord` and `Eq` implementations
+// on these wider integer types.
+type Key = (u32, u64);
+type Val = Box<XdeDev>;
+
+pub struct DevMap {
+    devs: BTreeMap<Key, Val>,
+    names: BTreeMap<String, Key>,
+}
+
+impl Default for DevMap {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl DevMap {
+    pub const fn new() -> Self {
+        Self { devs: BTreeMap::new(), names: BTreeMap::new() }
+    }
+
+    pub fn insert(&mut self, val: Val) -> Option<Val> {
+        let key = get_key(&val);
+        _ = self.names.insert(val.devname.clone(), key);
+        self.devs.insert(key, val)
+    }
+
+    pub fn remove(&mut self, name: &str) -> Option<Val> {
+        self.devs.remove(&self.names.remove(name)?)
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn get(&self, vni: Vni, mac: MacAddr) -> Option<&Val> {
+        self.devs.get(&(vni.as_u32(), mac_to_u64(mac)))
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn get_by_name(&self, name: &str) -> Option<&Val> {
+        self.devs.get(self.names.get(name)?)
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = &Val> {
+        self.devs.values()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.devs.is_empty()
+    }
+}
+
+#[inline(always)]
+fn mac_to_u64(val: MacAddr) -> u64 {
+    let val = val.bytes();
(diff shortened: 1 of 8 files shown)


xde/src/dev_map.rs
```diff
@@ -0,0 +1,90 @@
...
+
+use crate::xde::XdeDev;
+use alloc::boxed::Box;
+use alloc::collections::btree_map::BTreeMap;
+use alloc::string::String;
+use opte::api::MacAddr;
+use opte::api::Vni;
+
+// From microbenchmarking (https://github.com/oxidecomputer/opte/issues/637)
+// it is apparent that we have *far* faster `Ord` and `Eq` implementations
```
luqmana, line 15: hmm, worth updating those impls at some point then?

luqmana (approved):
LGTM, nice improvement!

xde/src/dev_map.rs
FelixMcFelix: Thanks for the review, Luqman!

I tested this out moving the conversions into the Eq/Ord impls just there – this didn't make any of the `(Vni, MacAddr)` cases as fast as the `(u32, u64)` ones. Linear scans got a bit slower.

[I put a quick proxy of this into godbolt](https://godbolt.org/z/3P99Kx3GE) to see what rustc is actually doing -- it looks to me as though for `MacAddr`, `Ord` works out using `core::intrinsics::compare_bytes` and doing the same BE conversion on the first 4B and last 2B. I think most of the improvement I get in the benchmarks comes down to precaching this conversion for the entire map.

This has me wondering if we can throw the same trick at `InnerFlowId`s -- if there's anything there I'll open a followup.

FelixMcFelix: Sadly, I did not see any measurable improvement on packet processing time when trying related shenanigans with `InnerFlowId` -> `[u64; 5]` for `FlowTable`s. This remained the case even at fairly high UFT occupancy.

PR #677, opened 2025-02-11
update oxnet to 0.1.0 from crates.io

sunshowers:
Also update libnet to explicitly specify `main`, in order to ensure that two
copies of the crate aren't floating around in other dependency graphs.


ahl:
I'm sort of the opinion, if it ain't broke. But up to you. Looks good as intended

FelixMcFelix (approved):
Thanks, Rain!

PR #690, opened 2025-02-21
Only use `KMutexType::Driver` in `mutex_init`

FelixMcFelix:
Pulled out of #636. We were creating some `Spin` locks which, misconfigured, would end up falling back to adaptive behaviour as part of LFT creation. Moreover, they would hit a debug assert and panic on test bits. (!!!)

Given that we have no means of specifying priority in `mutex_init`, we now make sure that all created `KMutex`es are correct by removing the illusion of free choice.

### DIFF
--- a/lib/opte/src/ddi/sync.rs
+++ b/lib/opte/src/ddi/sync.rs
@@ -119,7 +119,8 @@ impl<T> KMutex<T> {
         // Safety: ???.
         unsafe {
             // MUTEX_DRIVER is the only type currently sanctioned by the DDI
-            // for use here. Other types require the priority argument to be set.
+            // for use here. The priority argument, when we provide it, will
+            // control whether we get adaptive/spin behaviour.
             mutex_init(
                 &mut kmutex,
                 ptr::null(),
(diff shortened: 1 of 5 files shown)


lib/opte/src/ddi/sync.rs
```diff
@@ -118,7 +118,14 @@ impl<T> KMutex<T> {
         //
         // Safety: ???.
         unsafe {
-            mutex_init(&mut kmutex, ptr::null(), mtype.into(), ptr::null());
+            // MUTEX_DRIVER is the only type currently sanctioned by the DDI
+            // for use here. Other types require the priority argument to be set.
```
citrus-it, line 122: `MUTEX_DRIVER` is the only sanctioned type, but you can use `MUTEX_DRIVER ` along with a priority argument when creating a mutex for use inside an interrupt handler.  In that case you may well end up with a spin lock depending on the priority.

This is what the comment at 114:115 is alluding to. This'll need to grow support for the priority argument if you ever want interrupt context locks, but you'd still be using `MUTEX_DRIVER` for the type.

citrus-it (approved):
Thanks for fixing this.

PR #701, opened 2025-03-04
Rework rust toolchain installation

FelixMcFelix:
CI on recent PRs is breaking, due to rustup 1.28.0+ no longer autoinstalling the correct rust toolchain version. This hurts us immediately since we have *two* toolchains (pinned nightly and stable), and deliberately specified the nightly for some tooling.

This PR changes this over to use buildomat's auto-installation for the stable variant, and the new toolchain show -> install pattern for nightly. This also lets us place `$NIGHTLY` into most of our `cargo fmt` invocations, which should reduce the busywork in future compiler bumps for XDE.

### DIFF
--- a/.github/buildomat/common.sh
+++ b/.github/buildomat/common.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# Install both toolchains required for OPTE.
+# We pin to both a specific nightly *and* a stable compiler version
+# due to XDE's reliance on unstable features.
+rustup show active-toolchain || rustup toolchain install
+
+pushd xde
+rustup show active-toolchain || rustup toolchain install
+export NIGHTLY=`rustup show active-toolchain | head -n 1`
+popd
+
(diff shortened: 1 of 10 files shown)


.github/buildomat/common.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# Install both toolchains required for OPTE.
+# We pin to both a specific nightly *and* a stable compiler version
+# due to XDE's reliance on unstable features.
+rustup show active-toolchain || rustup toolchain install
+
+pushd xde
+rustup show active-toolchain || rustup toolchain install
+export NIGHTLY=`rustup show active-toolchain -v | head -n 1 | cut -d' ' -f1`
```
zeeshanlakhani, line 10: 👍🏽

zeeshanlakhani (approved):
Looks great (and some nice cleanups too)

Issue #730, opened 2025-04-17
Firewall rules should allow for ICMP Code/Type to be selected


FelixMcFelix:
User-facing services (nexus) want to be able to support path MTU discovery for end users. This requires that we allow for a subset of ICMP messages to transit from Nexus's zone out of OPTE -- in this case, [type 3 & code 4](https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml#icmp-parameters-codes-3) (Destination Unreachable//Fragmentation Needed and Don't Fragment was Set). However, we do *not* want to open up all of ICMP to the Internet writ large!

We need to be able to express constraints on ICMP code/type in individual firewall rules (individual, set, ranges, etc.) to allow for targeted control of ICMP traffic.

Follows on from https://github.com/oxidecomputer/omicron/issues/7998.

Issue #369, opened 2023-05-13
NAT broken for ICMP messages that contain original datagram as payload.


rcgoodfellow:
There are several [ICMP messages](https://www.rfc-editor.org/rfc/rfc792) that contain a copy of the original datagram that caused the ICMP message to be sent. When the sender of the original message is behind an OPTE NAT, the expectation would be that the copy of the original datagram in the ICMP message payload would be translated to be consistent with what the guest sent.

As a concrete example, we see traceroute packets like the following from within the guest.

```
23:32:41.655456 IP (tos 0x0, ttl 250, id 2385, offset 0, flags [DF], proto ICMP (1), length 96)
    68.86.93.241 > 172.30.0.5: ICMP time exceeded in-transit, length 76
        IP (tos 0x20, ttl 1, id 28766, offset 0, flags [DF], proto ICMP (1), length 60)
    10.100.0.11 > 1.1.1.1: ICMP echo request, id 914, seq 19, length 40 (wrong icmp cksum c209 
(->7ed5)!)
```

Here the guest address is `172.30.0.5` and the assigned external IP is `10.100.0.11`. As we can see the ICMP packet itself had NAT performed showing `68.86.93.241 > 172.30.0.5`. However, the copy of the original packet in the ICMP payload shows `10.100.0.11 > 1.1.1.1`. This prevents traceroute from working properly.

PR #734, opened 2025-04-22
Explicitly drop non-echo ICMP at SNAT

### LINKED ISSUE #727: OPTE spewing onto serial console
I noticed this recently in an a4x2 environment. OPTE seems to be dumping logs onto the root serial console.

```
root@oxz_ntp_dc329588:~# WARNING: failed to generate descriptor for stateful action: ICMP:172.30.3.5:0:162.159.200.123:0 Unexpected { msg: "Expected ICMP Echo Request, found: 3" }
                                                                                WARNING: failed to generate descriptor for stateful action: ICMP:172.30.3.5:0:162.159.200.123:0 Unexpected { msg: "Expected ICMP Echo Request, found: 3" }
                                    WARNING: failed to generate descriptor for stateful action: ICMP:172.30.3.5:0:162.159.200.123:0 Unexpected { msg: "Expected ICMP Echo Request, found: 3" }

root@oxz_ntp_dc329588:~#
root@oxz_ntp_dc329588:~#
root@oxz_ntp_dc329588:~#
root@oxz_ntp_dc329588:~# WARNING: failed to generate descriptor for stateful action: ICMP:172.30.3.5:0:162.159.200.1:0 Unexpected { msg: "Expected ICMP Echo Request, found: 3" }

root@oxz_ntp_dc329588:~#
root@oxz_ntp_dc329588:~#
root@oxz_ntp_dc329588:~# exit
logout

[Connection to zone 'oxz_ntp_dc329588-765b-4f65-b8ad-33c200342bca' pts/1 closed]
root@g0:~#
root@g0:~#
root@g0:~# WARNING: failed to generate descriptor for stateful action: ICMP:172.30.3.5:0:162.159.200.123:0 Unexpected { msg: "Expected ICMP Echo Request, found: 3" }
WARNING: failed to generate descriptor for stateful action: ICMP:172.30.3.5:0:162.159.200.123:0 Unexpected { msg: "Expected ICMP Echo Request, found: 3" }
```

FelixMcFelix:
Our SNAT layer has limited support for ICMP traffic, and is limited to using the identifier of Echo Request/Reply traffic as a form of pseudo-port. For other ICMP traffic, there is no comparable field and the packet must be dropped.

This PR changes how this is requested, from a `GenDescError` (ordinarily used for programming errors or resource exhaustion) into a standard drop. This is more appropriate given that guests are free to (attempt to) send these other ICMP variants at their leisure, including in response to an ibound frame on a recently closed port.

Closes #727.

### DIFF
--- a/lib/opte/src/engine/snat.rs
+++ b/lib/opte/src/engine/snat.rs
@@ -199,20 +199,16 @@ pub struct SNat<T: ConcreteIpAddr> {
     icmp_pool: Arc<NatPool<T>>,
 }
 
-enum GenIcmpErr<T: Display> {
+enum GenIcmpErr {
     MetaNotFound,
-    NotRequest(T),
 }
 
-impl<T: Display> From<GenIcmpErr<T>> for GenDescError {
-    fn from(val: GenIcmpErr<T>) -> Self {
+impl From<GenIcmpErr> for GenDescError {
+    fn from(val: GenIcmpErr) -> Self {
         GenDescError::Unexpected {
             msg: match val {
                 GenIcmpErr::MetaNotFound => {
                     "No ICMP metadata found despite Protocol::ICMP".to_string()
-                }
-                GenIcmpErr::NotRequest(v) => {
-                    format!("Expected ICMP Echo Request, found: {}", v)
                 }
             },
         }
@@ -236,7 +232,11 @@ impl<T: ConcreteIpAddr + 'static> SNat<T> {
         }
     }
 
-    // A helper method for generating an SNAT + ICMP(v6) action descriptor.
+    /// A helper method for generating an SNAT + ICMP(v6) action descriptor.
+    ///
+    /// Only echo requests will be admitted, as these a) signify a new flow,
+    /// and b) are SNAT-compatible via their Echo ID parameter. All other ICMP
+    /// packets are explicitly dropped.
     fn gen_icmp_desc(
         &self,
         nat: SNatAlloc<T>,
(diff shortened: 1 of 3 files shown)


rcgoodfellow:
This may also have interactions with the solutions for #730 and #369.

lib/opte/src/engine/snat.rs
```diff
@@ -246,37 +246,38 @@ impl<T: ConcreteIpAddr + 'static> SNat<T> {
...
             Protocol::ICMPv6 => {
-                let icmp6 = meta
-                    .inner_icmp6()
-                    .ok_or(GenIcmpErr::<Icmpv6Message>::MetaNotFound)?;
-                if icmp6.ty() != u8::from(Icmpv6Message::EchoRequest) {
-                    Err(GenIcmpErr::NotRequest(icmp6.ty()).into())
+                let icmp6 =
+                    meta.inner_icmp6().ok_or(GenIcmpErr::MetaNotFound)?;
+
+                Ok(if icmp6.ty() == u8::from(Icmpv6Message::EchoRequest) {
```
rcgoodfellow, line 261: Are we not handling echo reply here because this is for initializing a few flow only?

FelixMcFelix: In general for SNAT, Echo Reply messages should all be handled by the LFT entry created as part of a stateful allow here.

The other thing is that this code is only used in the outbound direction. The only time a guest should send an echo reply would be unsolicited (?), or if the other side quickly follows up with its own echo request on the same identifier (which is, again, handled by the pair of LFT entries created by the first outbound packet).

FelixMcFelix:
Thanks for the review, Ry.

> As a high level comment regarding #727. Can we make it so that OPTE logs do not land on the root console at all?

We can, I believe that downgrading the error level for `record_gen_desc_failure`/`record_gen_ht_failure` to `LogLevel::Note` should suffice? (The other two levels correspond to `CE_WARN`).

> This may also have interactions with the solutions for #730 and #369.

I don't really see offhand that it would, but I might be wrong -- #730 would only really factor in if we wanted to set up an explicit SNAT-incompatible-ICMP rule for the purpose of tracking dedicated stats. #639 we're most likely to solve using a body transform on the _NAT_ (EIP/FIP) action, since the affected packet types are incompatible with how we perform SNAT.

rcgoodfellow:
> We can, I believe that downgrading the error level for record_gen_desc_failure/record_gen_ht_failure to LogLevel::Note should suffice? (The other two levels correspond to CE_WARN).

That sounds like something we should do. As this can completely destroy one's ability to manage a sled from the root login.

rcgoodfellow (approved):

PR #735, opened 2025-04-24
opteadm/`OpteHdl` cleanup

### LINKED ISSUE #529: `OpteAdm` duplicates a significant amount of code from `OpteHdl`
Currently, ioctls are implemented identically between both `OpteAdm` (bin/opteadm/src/lib.rs) and `OpteHdl`(lib/opte-ioctl/src/lib.rs), and each has several ioctls that the other does not. All of this functionality should be implemented once on `OpteHdl`. `OpteAdm` could instead be used instead to contain any equivalent methods for CLI-specific types, if needed:

```rust
pub struct OpteAdm(OpteHdl);

impl Deref for OpteAdm {
    type Target = OpteHdl;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

impl OpteAdm {
    // ... CLI-specific convenience/conversion methods ...
}
```

### LINKED ISSUE #755: `opteadm` is pulling in tokio as a dependency
`opteadm` is a fully sync program, as are its interactions with illumos via [`libnet`](https://github.com/oxidecomputer/netadm-sys).

Just now it looks like, via libnet and dlpi, we're pulling in tokio as a dependency. We should chase this down to cut down build-size/time.

FelixMcFelix:
This PR moves all base ioctl operations into `opte_ioctl::OpteHdl`, allowing us to remove the vast majority of code in the `OpteAdm` handle.

In addition, I've done some extra reorganisation and dependency grooming s.t. `opteadm` no longer compiles any `engine` features/code from OPTE, which gets the release binary size from 46.2 -> 31.7 MiB. Thin LTO takes that down to 21.3 MiB. Opteadm compile time drops from ~40s on `master` to ~30s here 🎉.

One of the blockers here was that several ioctls relied on serialising/deserialising types which were only exposed as part of the `engine` feature flag. Accordingly, a few type definitions have been moved into `opte_api` (`RuleId`, `TcpState`) and `opte::api` (`InnerFlowId`). I've tried to be cognisant of the longer term goal of having primitives like `InnerFlowId` be chosen by the `NetworkImpl` by parametrising the `Dump` ioctls.

Closes #529.
Closes #755.


lib/opte/src/api.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+impl InnerFlowId {
+    #[cfg(feature = "engine")]
+    pub fn crc32(&self) -> u32 {
+        let mut hasher = Hasher::new();
+        self.hash(&mut hasher);
+        hasher.finalize()
+    }
+}
```
FelixMcFelix, line 70: The main reason this type lives here and not in `opte_api` is so that we can have this `crc32` method implemented directly on the flow ID, while keeping `crc32fast` limited to `engine`. Adding a conditional import to `opte_api` itself felt... bad.

In the fullness of time (#754), we probably want our base `FlowId` trait, at which point `opte::engine` can add its own `FlowIdExt: FlowId` trait with methods like this.

xtask/src/main.rs
```diff
@@ -422,7 +422,12 @@ impl BuildTarget {
         match self {
             Self::OpteAdm => {
                 println!("Building opteadm ({p_name}).");
-                build_cargo_bin(&["--bin", "opteadm"], p_name, None, true)
+                build_cargo_bin(
+                    &["--bin", "opteadm"],
+                    p_name,
+                    Some("bin/opteadm"),
+                    true,
+                )
```
FelixMcFelix, line 430: Building from the workspace root appeared to be pulling in more dependencies via feature unification.

FelixMcFelix (commented):
Fell into a bit of a rabbithole around build-features while cleaning this up. I've opened #755 as a result of staring at one lockfile too many.

FelixMcFelix:
Moved back to draft -- looks like there's a CI issue between older scripts pointing at `release` vs `release-lto`. Will fix that up first.

xtask/src/main.rs
rcgoodfellow: Adding a comment in the code about this would be helpful.

rcgoodfellow (approved):
Thanks @FelixMcFelix. Just a few small comments. Overall LGTM.

bin/opteadm/src/lib.rs
```diff
@@ -2,328 +2,32 @@
...
-        name: &str,
-    ) -> Result<api::DumpLayerResp, Error> {
-        let cmd = OpteCmd::DumpLayer;
-        let req = api::DumpLayerReq {
-            port_name: port_name.to_string(),
-            name: name.to_string(),
-        };
-        run_cmd_ioctl(self.device.as_raw_fd(), cmd, Some(&req))
-    }
+pub struct OpteAdm(OpteHdl);
```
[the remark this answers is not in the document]
FelixMcFelix: I thought we might have done initially (to have a convenient home for `opteadm`-specific logic), but yeah it's turned out unnecessary. Happy to do away with it.

lib/opte-ioctl/src/lib.rs
```diff
@@ -179,6 +225,16 @@ impl OpteHdl {
         run_cmd_ioctl(self.device.as_raw_fd(), cmd, Some(&req))
     }
 
+    /// Dump the Virtual-to-Boundary mappings.
+    pub fn dump_v2b(&self) -> Result<DumpVirt2BoundaryResp, Error> {
+        let cmd = OpteCmd::DumpVirt2Boundary;
+        run_cmd_ioctl(
+            self.device.as_raw_fd(),
+            cmd,
+            Some(&DumpVirt2BoundaryReq { unused: 99 }),
```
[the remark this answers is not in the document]
FelixMcFelix: We can go a little bit further and just send `None` as the request payload, as we do in several other ioctls (e.g. `OpteCmd::ListPorts`).

FelixMcFelix:
> Thanks @FelixMcFelix. Just a few small comments. Overall LGTM.

Thanks a lot for the review. It looks like we got to cut a few more `A { _unused }` types out of the interface too.

PR #769, opened 2025-06-10
Disable per-function sections

dancrossnyc:
Disable per-function sections by setting the,
`function-sections` parameter to `false` in the
target JSON file.

This reduces the number of generated sections from thousands to 36.


dancrossnyc:
Happy to do it!

PR #792, opened 2025-07-09
Do not generally treat zero-checksums as 'omitted'

### LINKED ISSUE #791: `0x0000` should be treated as a valid checksum for all protocols except UDP
Zero is a valid checksum value for IPv4, TCP, and the ICMP variants. We're assuming this isn't the case in a few spots:

* Body checksum extraction, which later determines whether we update the ULP checksum if required
* Application of the IPv4 checksum update.

This seems to be the root cause of https://github.com/oxidecomputer/customer-support/issues/470.

FelixMcFelix:
Fixes #791.

### DIFF
--- a/lib/opte/src/engine/packet.rs
+++ b/lib/opte/src/engine/packet.rs
@@ -1242,9 +1242,10 @@ impl<T: Read + Pullup> Packet<FullParsed<T>> {
         // contained in the mblk_t. I would say that this gives us a perfect
         // view on which checksums (other than UDP) have been omitted, but:
         //
-        // * IP is in the unfortunate position where `HCK_IPV4_HDRCKSUM`
+        // * IP is in the unfortunate position where the `HCK_IPV4_HDRCKSUM`
         //   (requesting offload) and `HCK_IPV4_HDRCKSUM_OK` (the NIC verified
-        //   the checksum) are identical.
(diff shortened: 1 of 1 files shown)


lib/opte/src/engine/packet.rs
```diff
@@ -1249,32 +1225,47 @@ impl<T: Read + Pullup> Packet<FullParsed<T>> {
...
-        // If a ULP exists, then compute and set its checksum.
-        if let (true, Some(ulp)) =
-            (update_ulp, &mut self.state.meta.headers.inner_ulp)
+        // In future we will want to factor in the offload information
+        // contained in the mblk_t. I would say that this gives us a perfect
+        // view on which checksums (other than UDP) have been omitted, but:
+        //
+        // * IP is in the unfortunate position where `HCK_IPV4_HDRCKSUM`
+        //   (requesting offload) and `HCK_IPV4_HDRCKSUM_OK` (the NIC verified
+        //   the checksum) are identical.
```
rcgoodfellow, line 1247: This confused me until i looked up the illumos source. Maybe saying something like `HCK_IPV4_HDRCKSUM` and `HCK_IPV4_HDRCKSUM_OK` map to the same message block flag value would be clearer.

rcgoodfellow (approved):
Thanks Kyle. LGTM just one small suggestion on one of the comments.

Issue #760, opened 2025-05-25
Add Geneve Option for Multicast Tag


zeeshanlakhani:
With the additional Geneve option added as [part of Multicast replication](https://github.com/oxidecomputer/dendrite/pull/14/files#diff-57021d70232ce254374c31d562a0ad1751736699286121626c1b9af8400fe684R181) in https://github.com/oxidecomputer/dendrite/pull/14, we need to [update the Geneve engine in OPTE]( https://github.com/oxidecomputer/opte/blob/master/lib/opte/src/engine/geneve.rs#L205) to parse and set this option.

PR #847, opened 2025-09-04
Initial multicast support

FelixMcFelix:
Closes: #760 (among other needs). 

XXX currently sketching out the bare minimum requirements around frame delivery, insertion of Geneve options, etc.

TODO:
- [x] Add XDE-wide multicast forwarding table (`Ipv6Addr` -> `BTreeMap<(NextHopV6, Replication)>`).
- [x] Add exceptions in source/dest MAC/L3 addr checking for multicast addreses matching known groups. (#cidr-checking only mcast dst side)
- [x] Populate `Mcast2Phys` and above table via ioctl.
- [x] Some tests might be nice? :)

### DIFF
--- a/xde-tests/src/lib.rs
+++ b/xde-tests/src/lib.rs
@@ -162,8 +162,20 @@ impl OpteZone {
         Ok(Self { zone })
     }
 
+    /// Wait for the SMF repository to be available in the zone.
+    /// This must be called before `wait_for_network` to avoid failures
+    /// when the zone is still booting.
+    fn wait_for_smf(&self) -> Result<()> {
+        poll_until(
+            || self.zone.zexec("svcs -a").map(|_| true).unwrap_or(false),
+            Duration::from_secs(60),
+            "SMF repository",
+        )
+    }
+
     /// Wait for the network to come up, then set up the IPv4 overlay network.
     fn setup(&self, devname: &str, addr: Ipv4Addr) -> Result<()> {
+        self.wait_for_smf()?;
         self.zone.wait_for_network()?;
         // Configure IPv4 via DHCP
         self.zone
@@ -183,6 +195,7 @@ impl OpteZone {
         ipv4_addr: Ipv4Addr,
         ipv6_addr: Ipv6Addr,
     ) -> Result<()> {
+        self.wait_for_smf()?;
         self.zone.wait_for_network()?;
         // Configure IPv4 via DHCP (OPTE provides DHCP server via hairpin)
         self.zone


.gitignore
```diff
@@ -1,4 +1,5 @@
 *.html
 target
 download
-.DS_STORE
+scripts
```
FelixMcFelix: Is this a part of your local setup you're trying to keep out-of-tree?

lib/opte/README.adoc
```diff
@@ -209,11 +209,47 @@ resources.
...
+Delivery Modes (Replication):: The Oxide Geneve multicast option carries
+the delivery mode as a 2‑bit field in the top two bits of the option
+body's first byte:
+
+* External — local guest delivery within the same VNI: OPTE decapsulates
+  and delivers to all local subscribers (guests) on the port map.
+* Underlay — infrastructure delivery: OPTE sends Geneve‑encapsulated
+  packets towards the configured underlay multicast address in fleet
+  VNI 77. The underlay performs any further replication.
+* All — both behaviors above.
```
FelixMcFelix: This feels a bit like a few concepts are being conflated here. `Replication` and its types were originally defined in terms of what the *switch* would do when receiving a marked packet:

https://github.com/oxidecomputer/opte/blob/ea3a19dc2cdd4081545921a050978480c18cbadb/lib/oxide-vpc/src/engine/geneve.rs#L95-L107

I.e., `External` => switch will replicate to front panel ports, `Underlay` => switch will replicate to sleds, `All` asks that it does both. Are we also using this terminology as an access control mechanism to determine which VNIs will receive a sled-originated packet?

How you've defined `External` in this case confuses me a bit, in that replicating a transmitted packet to all potential local recipients in the same VNI is something that you'd *always* want to do. Does this tie into the relay delivery that's currently on the Rx path?

lib/oxide-vpc/src/api.rs
```diff
@@ -20,13 +20,101 @@ use serde::Deserialize;
...
+///
+/// This is the only VNI currently supported for multicast traffic.
+/// All multicast groups (M2P mappings and forwarding entries) must use this VNI.
+/// OPTE validates that multicast operations specify this VNI and rejects others.
+///
+/// **Security model:** While M2P (Multicast-to-Physical) mappings are stored
+/// per-VNI in the code, the enforcement of DEFAULT_MULTICAST_VNI means all
+/// multicast traffic shares a single namespace across the rack, with no
+/// VPC-level isolation (as multicast groups are fleet-wide).
+pub const DEFAULT_MULTICAST_VNI: u32 = 77u32;
```
FelixMcFelix: Okay, so we aren't yet doing any VPC-level isolation. I guess the only thing we need to be aware of is what the source IP looks like, IIRC there was discussion at some point that in this case all packets should be NAT'd.

lib/oxide-vpc/src/engine/geneve.rs
```diff
@@ -5,7 +5,72 @@
...
+//!
+//! ### Replication Semantics
+//!
+//! - **External**: Packet should be decapsulated and delivered to local guest
+//!   instances subscribed to this multicast group. Switch sets `nat_egress_hit`
+//!   to trigger decapsulation before delivery.
+//! - **Underlay**: Packet should remain encapsulated and forwarded to underlay
+//!   infrastructure destinations.
+//! - **All**: Bifurcated delivery to both local guests (decapsulated) and
+//!   underlay destinations (encapsulated).
```
FelixMcFelix: Hmm, still of the opinion that OPTE should only care about these on Tx, for telling the switch how to replicate the frame. Same comment about `External` meaning what we would probably consider 'internal' delivery.

lib/oxide-vpc/src/engine/geneve.rs
```diff
@@ -84,28 +149,24 @@ impl<'a> OptionCast<'a> for ValidOxideOption<'a> {
...
+/// This option encodes the replication scope as a 2-bit field in the top two
+/// bits of the first byte of the option body. The remaining 30 bits are
+/// reserved for future use. The replication strategy determines whether the
+/// packet is delivered to local guest instances (External), underlay
+/// infrastructure destinations (Underlay), or both (All).
 #[derive(Debug, Clone, Ingot, Eq, PartialEq)]
 #[ingot(impl_default)]
 pub struct MulticastInfo {
+    /// Replication scope encoded as a u2 (top 2 bits of the first byte).
+    /// Values map to `Replication::{External, Underlay, All, Reserved}`.
```
FelixMcFelix: comment nit: I think the representation and enum variants are already shown unambiguously by the code.

xde/src/xde.rs
```diff
@@ -1810,78 +1892,755 @@ fn guest_loopback(
...
+        // Re-acquire locks for underlay forwarding
+        let xde = get_xde_state();
+        let mgmt = xde.management_lock.lock();
+        let mcast_fwd = mgmt.mcast_fwd.read();
+
+        if let Some(next_hops) = mcast_fwd.get(&ctx.inner_dst) {
+            // We found forwarding entries, replicate to each next hop
+            for (next_hop, replication) in next_hops.iter() {
+                // Clone packet with headers using pullup
+                let pullup_len = (ctx.encap_len as usize)
```
FelixMcFelix: All of these offsets will be identical to the above cases, nor do they need to be recomputed in the `for` loop. I don't think it's unreasonable to just unconditionally compute `pullup_len`, `geneve_offset` and friends considering they must be used at least once somewhere if we've found a valid M2P entry etc and successfully processed the packet.

FelixMcFelix (commented):
Thanks for the work so far. I haven't looked at the new multicast integration tests yet, but I have a bunch of questions from what I've looked at thus far.

.gitignore
zeeshanlakhani: I had an xde reset thing I was using. I'll remove this.

lib/opte/README.adoc
zeeshanlakhani: Yeah. I was moving toward reshaping this as delivery, which is what you're detailing. I can definitely do some cleanup here to capture the right semantics.

xde/src/xde.rs
```diff
@@ -1810,78 +1892,755 @@ fn guest_loopback(
...
+    if do_external {
+        let entry_state =
+            entry_state.get_or_insert_with(|| src_dev.port_map.read());
+        if let Some(others) = entry_state.mcast_listeners(&ctx.inner_dst) {
+            let my_key = VniMac::new(ctx.vni, src_dev.port.mac_addr());
+            for el in others {
+                // Filter by VNI - only deliver to listeners in the same VNI
+                if el.vni() != ctx.vni {
+                    continue;
+                }
```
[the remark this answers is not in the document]
zeeshanlakhani: Will make the Vni check only in overlay, and not do this in xde for sure.

xde/src/xde.rs
```diff
@@ -1921,34 +2694,39 @@ unsafe extern "C" fn xde_mc_tx(
...
-    //
-    // Maybe we should clone out the `DevMap` at this instant.
-    let mut entry_state = None;
+    // Clone per-CPU mcast forwarding table Arc and drop lock immediately.
+    // This makes the reader lock-free and avoids blocking management refreshes.
+    let cpu_index = current_cpu().seq_id;
+    let cpu_entry = &src_dev.u1.stream.ports_map[cpu_index];
+    let mcast_fwd = clone_from_rwlock(&cpu_entry.mcast_fwd);
+
+    // Lazily clone per-port DevMap Arc for hairpin/local delivery.
```
zeeshanlakhani: @FelixMcFelix, of note, this is a different approach I took with Arc cloning `mcast_fwd` and then dropping the lock. Take a gander to see if you have worries here or below.

xde/src/xde.rs
```diff
@@ -1921,34 +2694,39 @@ unsafe extern "C" fn xde_mc_tx(
...
 
     let (local_pkts, [u1_pkts, u2_pkts]) = tx_postbox.deconstruct();
 
-    if let Some(entry_state) = entry_state {
-        entry_state.deliver_all(local_pkts);
+    // Local same-sled delivery (via mac_rx to guest ports) is safe.
+    // Lazily clone DevMap if we have anything to deliver.
+    if !local_pkts.is_empty() {
+        let devs = cached_devmap
+            .get_or_insert_with(|| clone_from_rwlock(&src_dev.port_map));
```
zeeshanlakhani: This `clone_from_rwlock` is also related to the change/note above.

lib/opte/src/ddi/sync.rs
```diff
@@ -780,3 +780,29 @@ impl<T> Drop for TokenGuard<'_, T> {
...
+/// This pattern is used throughout the datapath to make readers lock-free
+/// while keeping snapshots alive via Arc refcounting. The brief lock hold
+/// (just the Arc clone) minimizes contention and avoids blocking management
+/// operations like `refresh_maps()`.
+#[inline(always)]
+pub fn clone_from_rwlock<T>(
+    lock: &KRwLock<alloc::sync::Arc<T>>,
+) -> alloc::sync::Arc<T> {
+    alloc::sync::Arc::clone(&*lock.read())
+}
```
FelixMcFelix: ?? Note to self, audit these (also `_mutex`) and their uses.

lib/oxide-vpc/src/engine/overlay.rs
```diff
@@ -1021,35 +1025,12 @@ impl Default for Mcast2Phys {
...
-    pub fn addr(&self) -> Ipv6Addr {
-        self.0
-    }
-
-    /// Return the destination MAC address derived from the IPv6 multicast address.
-    fn dst_mac(&self) -> MacAddr {
-        self.0.unchecked_multicast_mac()
-    }
-}
+pub struct MulticastUnderlay(pub Ipv6Addr);
```
FelixMcFelix: I would generally shy away from having a newtype like this expose its innards as `pub`, since that makes it trivial to construct an invalid instance.

xde/src/xde.rs
```diff
@@ -1236,50 +1462,110 @@ fn clear_xde_underlay() -> Result<NoResp, OpteError> {
...
-        //
-        // We know these must succeed given that the only holders of an
-        // `Arc<XdeUnderlayPort>` are `XdeState` (whose ref we have exclusively locked)
-        // and `XdeDev` (of which none remain).
-        let name = underlay.u1.name.clone();
-        let u1 = Arc::into_inner(underlay.u1).unwrap_or_else(|| {
-            panic!("underlay u1 ({name}) must have one ref during teardown",)
-        });
+    // Clear multicast forwarding table to release any references
+    token.mcast_fwd.write().clear();
```
FelixMcFelix: I don't think it is holding any references? It contains only data, rather than any smart pointer types.

xde/src/xde.rs
```diff
@@ -1320,50 +1465,107 @@ fn clear_xde_underlay() -> Result<NoResp, OpteError> {
...
+                    *map = Arc::clone(&empty_map);
+                }
+                {
+                    let mut mcast = entry.mcast_fwd.write();
+                    *mcast = Arc::clone(&empty_mcast);
+                }
+                entry.has_mcast_subscribers.store(false, Ordering::Relaxed);
+            }
+        }
+    }
```
FelixMcFelix: I don't see why we have this here. If `token.devs` is empty and we have the management lock, all `DevMaps` seen by `PerEntryState` should also be empty by definition -- any `delete`s have succeeded, and no one else can attempt a write to `token.devs` in parallel. So, there really should be no other references to the underlay. If they're not empty, that's a fatal programming error on our part?

xde/src/xde.rs
```diff
@@ -1810,78 +2094,567 @@ fn guest_loopback(
...
+            // VNI is at offset 4 in Geneve header (3 bytes)
+            if let Some(vni_bytes) =
+                fwd_pkt.get_mut(geneve_offset + 4..geneve_offset + 7)
+            {
+                let vni_be = next_hop.vni.as_u32().to_be_bytes();
+                vni_bytes.copy_from_slice(&vni_be[1..4]); // VNI is 24 bits
+            }
+            // Update Geneve multicast option to reflect underlay replication to
+            // prevent re-relay loops.
+            update_mcast_replication(&mut fwd_pkt, geneve_offset, *replication);
```
FelixMcFelix: Nit: dendrite should already be preventing reflection/re-relay, I think. IIRC we have the logic for it not to mcast-forward on the port a packet came inbound on (or, at least, to drop during egress processing).

xde/src/xde.rs
```diff
@@ -2490,6 +3387,82 @@ fn xde_rx_one(
...
+            oxide_vpc::engine::geneve::validate_options(&meta.outer_encap)
+        {
+            stat_parse_error(Direction::In, &e);
+            opte::engine::dbg!(
+                "Invalid Geneve options in multicast packet: {:?}",
+                e
+            );
+            bad_packet_parse_probe(None, Direction::In, mblk_addr, &e);
+            return Some(pkt);
+        }
```
FelixMcFelix: OPTE should have already done this work as part of `parse_inbound`.

xde/src/xde.rs
```diff
@@ -56,66 +56,104 @@
...
+//! ### Teardown and reference cycles
+//! The Arc-cloning strategy creates a reference cycle during normal operation:
+//! underlay port → stream → ports_map (per-CPU) → [`DevMap`] → [`XdeDev`] → underlay port.
+//! This is benign during operation but must be broken during teardown.
+//!
+//! When `clear_xde_underlay()` is called (after all ports have been removed),
+//! we explicitly clear per-CPU cached `DevMap`s by replacing them with empty
+//! snapshots. This breaks the cycle and allows underlay port Arcs to be unwrapped.
+//! If brief in-flight TX chains still hold `DevMap` references, the unwrap returns
+//! EBUSY and the caller can retry. Refs drain quickly once caches are cleared.
```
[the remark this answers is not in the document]
zeeshanlakhani: Right, that's why I brought it up first to check. Makes sense.

lib/oxide-vpc/src/engine/overlay.rs
zeeshanlakhani: definitely. meant to go back and clean this up.

xde-tests/src/lib.rs
```diff
@@ -54,19 +115,95 @@ impl OpteZone {
...
     }
 
-    /// Wait for the network to come up, then set up the overlay network.
+    /// Wait for the network to come up, then set up the IPv4 overlay network.
     fn setup(&self, devname: &str, addr: String) -> Result<()> {
         self.zone.wait_for_network()?;
+        // Configure IPv4 with static address (immediate, no DHCP wait)
+        self.zone.zexec(&format!(
+            "ipadm create-addr -t -T static -a {addr}/24 {devname}/test"
+        ))?;
```
FelixMcFelix: I'm not sure we want to make this static -- it may be faster, but we're losing out on a) regularly testing that DHCP works against an OS network stack in our own CI and b) making sure that hairpin packet processing proceeds without issue*. Was the delay here causing bad interactions with one or more of the tests?

* ARP fills the same niche for b), I suppose.

xde-tests/src/lib.rs
```diff
@@ -106,20 +244,75 @@ impl OptePort {
...
+        })
     }
 
     /// Add an overlay routing entry to this port.
     pub fn add_router_entry(&self, dest: &str) -> Result<()> {
         let adm = OpteHdl::open()?;
         adm.add_router_entry(&AddRouterEntryReq {
             port_name: self.name.clone(),
-            dest: IpCidr::Ip4(format!("{}/32", dest).parse().unwrap()),
+            dest: IpCidr::Ip4(format!("{dest}/32").parse().unwrap()),
```
FelixMcFelix: Thanks for catching some of the `format!` inlining here -- Clippy must have been missed on this, my bad.

xde-tests/src/lib.rs
```diff
@@ -202,26 +467,172 @@ impl Xde {
...
+}
+
+impl Drop for MulticastGroup {
+    fn drop(&mut self) {
+        let Ok(hdl) = OpteHdl::open() else {
+            eprintln!("failed to open xde device for multicast cleanup");
+            return;
+        };
+
+        // Clear forwarding entry
```
FelixMcFelix: The automated cleanup here is really nice!

lib/opte/src/engine/geneve.rs
```diff
@@ -383,6 +383,15 @@ impl<'a, T: OptionCast<'a>> GeneveOptionParse<'a, T> {
...
 }
 
+impl<'a, T: OptionCast<'a>> HeaderLen for GeneveOptionParse<'a, T> {
+    const MINIMUM_LENGTH: usize = GeneveOpt::MINIMUM_LENGTH;
+
+    fn packet_length(&self) -> usize {
+        // Option header (4 bytes) + body length (already padded to 4-byte boundary)
+        GeneveOpt::MINIMUM_LENGTH + self.body_remainder.len()
+    }
+}
```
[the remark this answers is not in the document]
zeeshanlakhani: Good point.

xde-tests/src/lib.rs
zeeshanlakhani: Yeah. I moved toward static for testing initially, but could move this back to DHCP.

xde/src/xde.rs
```diff
@@ -1159,23 +1353,37 @@ fn delete_xde(req: &DeleteXdeReq) -> Result<NoResp, OpteError> {
...
 
-/// Rebuild each entrypoint's view of the central `DevMap`.
-fn refresh_maps(devs: KRwLockWriteGuard<DevMap>, underlay: &UnderlayState) {
+/// Rebuild each entrypoint's view of the central [`DevMap`] and multicast
+/// forwarding table `McastForwardingTable`.
+fn refresh_maps(
+    devs: KRwLockWriteGuard<DevMap>,
+    underlay: &UnderlayState,
+    mcast_fwd: &Arc<KRwLock<McastForwardingTable>>,
+) {
```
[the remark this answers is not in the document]
zeeshanlakhani: Yep. My bad on forgetting this. Will have a new name though.

xde/src/xde.rs
```diff
@@ -2704,6 +3652,358 @@ fn dump_v2b_hdlr() -> Result<DumpVirt2BoundaryResp, OpteError> {
...
+        let next_hop_map =
+            mcast_fwd.entry(underlay).or_insert_with(BTreeMap::new);
+
+        // Insert/update next hops: same next hop addr → replace replication mode,
+        // different next hop addr → add new entry (like `swadm route add`)
+        for (next_hop, rep) in req.next_hops {
+            next_hop_map.insert(next_hop, rep);
+        }
+
+        drop(mcast_fwd);
```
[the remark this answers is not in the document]
zeeshanlakhani: yeah. That's leftover from the changes.

lib/oxide-vpc/src/engine/geneve.rs
```diff
@@ -506,4 +522,80 @@ mod test {
...
+            // Geneve option: class 0x0129 (Oxide)
+            0x01, 0x29,
+            // Geneve option: flags+type (Multicast = 0x01, non-critical)
+            0x01,
+            // Geneve option: rsvd + len (1 word = 4 bytes body)
+            0x01,
+        ];
+        // Geneve option body: 4-byte body with replication in top 2 bits
+        buf.push(0x00); // Replication::External = 0b00 in top 2 bits
+        buf.extend_from_slice(&[0x00, 0x00, 0x00]);
```
FelixMcFelix: Nit: why not include these in the definition of `buf`?

xde/src/xde.rs
```diff
@@ -2174,14 +2223,12 @@ struct MulticastRxContext<'a> {
...
 ///
-/// `cpu_devs` may be None if the fast-path check indicated no local subscribers exist.
-///
 /// [`Replication`]: oxide_vpc::api::Replication
 fn handle_mcast_tx<'a>(
     ctx: MulticastTxContext,
     src_dev: &'a XdeDev,
     postbox: &mut TxPostbox,
-    cpu_devs: Option<&'a DevMap>,
+    cpu_devs: &'a DevMap,
```
FelixMcFelix: Nit: maybe `devs`, to make it clearer that this is not originated from the per-CPU state.

PR #873, opened 2025-10-30
Do not set `b_prev` when constructing `mblk_t` chains

FelixMcFelix:
`mac`, `ip`, etc. generally do not expect `b_prev` to be set in any packet chains they interact with. In some locations where packets are walked, the illumos stack breaks only the `b_next` link, and later asserts that both these pointers are `NULL`. We should be following the same set of assumptions employed elsewhere.

See oxidecomputer/stlouis#851.


dancrossnyc (approved):
Thanks, Kyle. It took me a while to get what was going on here, but I think I understand now, and this looks good to me.

PR #882, opened 2025-11-13
`DevMap` should not expose cloneable references

FelixMcFelix:
One of the purposes of `DevMap` is to ensure that the datapath does not inappropriately clone out and extend the lifetime of any ports. As it turns out, the code using `DevMap` is doing the right thing, but its API surface is overly permissive and should be making an `Arc::clone` or similar impossible.


zeeshanlakhani (approved):
I know the review for #847 is incoming, but, as discussed, we don't want to allow for the `Arc::clone` of `DevMaps` in that work, so it would be beneficial to have this in first.

PR #909, opened 2025-12-29
Src/Dst MAC rewrites should not require `unsafe`

FelixMcFelix:
A long-standing pattern of rewriting frame MAC addresses via direct pointer access has (unfortunately) replicated into a few spots in xde. Swap that out in favour of just using normal slice modifications.

This should have the minor benefit of resolving to a `ptr::copy_nonoverlapping` instead of a `ptr::copy`.

### DIFF
--- a/xde/src/xde.rs
+++ b/xde/src/xde.rs
@@ -2398,7 +2398,7 @@ fn handle_mcast_tx<'a>(
             // so we are guaranteed enough space to have Ethernet source/destination
             // contiguous.
             fwd_pkt[..ETHER_ADDR_LEN].copy_from_slice(&*dst_mac);
-            fwd_pkt[ETHER_ADDR_LEN..2 * ETHER_ADDR_LEN]
+            fwd_pkt[ETHER_ADDR_LEN..][..ETHER_ADDR_LEN]
                 .copy_from_slice(&*src_mac);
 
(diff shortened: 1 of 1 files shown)


xde/src/xde.rs
```diff
@@ -2392,15 +2393,13 @@ fn handle_mcast_tx<'a>(
...
-                ptr::copy(mac_src.as_ptr(), rptr.add(6), 6);
-
-                MsgBlk::wrap_mblk(mblk).unwrap()
-            };
+            // Fill in outer MAC addresses.
+            // In this case we know we have pulled up the outer headers,
+            // so we are guaranteed enough space to have Ethernet source/destination
+            // contiguous.
+            fwd_pkt[..ETHER_ADDR_LEN].copy_from_slice(&*dst_mac);
+            fwd_pkt[ETHER_ADDR_LEN..2 * ETHER_ADDR_LEN]
```
rcgoodfellow: nit: the pattern you use below would be nice here, same at 2966

```rust
fwd_pkt[ETHER_ADDR_LEN..][..ETHER_ADDR_LEN].copy_from_slice(&*src_mac);
```

FelixMcFelix: Good call, thanks for the quick review.

Issue #890, opened 2025-12-01
OPTE support for external subnets


rcgoodfellow:
OPTE support for attaching external and VPC subnets directly to instances. See [RFD 599](https://rfd.shared.oxide.computer/rfd/0599#_scope_of_work) for more details.

OPTE scope of work includes.

- [x] API for managing external subnet attachments
- [x] Implementation of ingress/egress flows for traffic going to/from attached external subnets.
- [ ] Plumbing for dynamic metadata service

rcgoodfellow:
Closing this as completed in #911. We'll come back to the dynamic metadata service separately.

PR #911, opened 2025-12-31
Attached (External) Subnets

### LINKED ISSUE #703: `ActionMeta` could allow direct storage/retrieval of an `ActionMetaValue`
Just now, we have a pattern for writing into action metadata in the slowpath:

```rust
// T: ActionMetaValue

// Layer A
let t: T = /* ... */;
meta.insert(self.target.key(), self.target.as_meta());

// ...

// Layer B
let target_str = match action_meta.get(T::KEY) { /* ... */ };
let target = match T::from_meta(target_str) { /* ... */ };
```

Given that  `KEY` is static, we should be able to perform typed storage/lookup with knowledge of `T` without needing to specify the key and value explicitly.

There is also the construction of `Predicate::Meta` entries, which follows a similar manual key and value selection. A convenience method to build a predicate from an `ActionMetaValue` would be tidier in general.

FelixMcFelix:
This PR reworks certain aspects of how NAT and gateway layers are constructed to enable attached (external) subnets to function. The primary aim here is that traffic to/from these subnets (owned by the host) do not undergo NAT, and bypass spoof detection as transit IPs would. A few wider changes have been necessary to ensure that these can be attached/detached without breaking any existing transit IPs, and to ensure that traffic originated from an external subnet cannot be directed towards a private VPC recipient.

- To be compatible with add/remove of transit IPs, we have to store both transit/attached subnets consistently in `VpcCfg` rather than working directly on the port's ruleset. This also gives us the opportunity to set these on port create rather than using a followup ioctl.
- Reconfiguring the gateway layer requires that we store `DhcpCfg` in `VpcCfg` now.
- Keeping separate copies of `VpcCfg` in `XdeDev` and the port itself seemed... misleading, at best. When XDE needs that, we now reach through `port.network()`.
- `StatefulAction`s should never have had mutable access to a packet's `ActionMeta` (the string-to-string map), because they may never be rerun after an LFT is created on subsequent slowpath walks. It is now up to the generated `ActionDesc`s to update metadata -- this trait's role is similar to `StaticAction`, which *is* allowed mutable access to this metadata.
- Use of V2B/V2P mappings is now explicitly restricted to sending traffic from external/internal IPs. I put in some rework of `ActionMeta` ergonomics to make this easier to express and implement.

As far as the control plane is concerned, it should need to only:
 * Attach external/VPC subnets as they arrive.
 * Ensure that for attached VPC subnets, the system router is updated s.t. that subnet has an `Instance(<name>)` target (in turn resolved to an IP target when reifying rules for the port, as happens with instance targets today).

Answers the functional dataplane requirements of #890. Closes #703.

### DIFF
--- a/lib/oxide-vpc/src/engine/overlay.rs
+++ b/lib/oxide-vpc/src/engine/overlay.rs
@@ -289,7 +289,7 @@ impl StaticAction for EncapAction {
                 // This requires a hairpin through the customer network, but provides
                 // strong isolation which some customers require.
                 //
-                // In future we want this to be a tunable property of the VPC. In this
+                // In future we may want this to be a tunable property of the VPC. In this
                 // case we would require an extra table/poptrie per VPC, containing all
                 // external CIDR blocks visible across the VPC. We would then:
                 //  * resolve `recipient` against this table when going via an IGW,
@@ -298,6 +298,13 @@ impl StaticAction for EncapAction {
                 //    the V2B.
                 //  * Possibly add the Geneve external packet tag to the packet, esp. if
                 //    crossing VPC boundaries.
+                // This obviously works well for attached subnets, but for EIPs and FIPs
+                // we'll have quite a few /32 or /128 routing table entries which can't
+                // be aggregated unless adjacent external IPs point to the same instance
+                // (and this would probably be harmed further by SNAT allocation causing
+                // fragmentation).
+                //
+                // It's a possible optimisation, but it'd need more thought.
                 RouterTargetInternal::InternetGateway(_) => {
                     match self.v2b.get(&recipient) {
                         Some(phys) if sent_from_eip => {


lib/oxide-vpc/src/engine/nat.rs
```diff
@@ -497,6 +557,108 @@ pub fn set_nat_rules(
...
-    let (in_rules, out_rules) = create_nat_rules(cfg, req.inet_gw_map)?;
+    refresh_nat_rules(port, req.inet_gw_map.as_ref())
+}
+
+pub fn attach_subnet(
+    port: &Port<VpcNetwork>,
+    inet_gw_map: Option<&InternetGatewayMap>,
+    vpc_mappings: &Arc<VpcMappings>,
+    req: AttachSubnetReq,
+) -> Result<(), OpteError> {
```
FelixMcFelix: I had a bit of uncertainty as to whether this and `detach_subnet` should live here, `gateway`, or some other module since the feature crosses multiple layers.

lib/oxide-vpc/src/api.rs
```diff
@@ -130,6 +130,26 @@ pub struct BoundaryServices {
     pub mac: MacAddr,
 }
 
+/// Configuration for a subnet completely owned by a NIC.
+///
+/// When configured this port will allow all in/out traffic matching a CIDR to
+/// be received/sent.
+#[derive(Debug, Clone, Serialize, Deserialize, Default, Eq, PartialEq)]
+pub struct AttachedSubnetConfig {
+    /// Denotes whether this attached subnet is an external IP block,
```
rcgoodfellow: So if it's not external, then NAT will be applied? What is that being used for?

lib/oxide-vpc/src/engine/nat.rs
rcgoodfellow: Living in its own module seems to make the most sense to me. I would not think to come to the `nat` module to look for attached subnet machinery.

lib/oxide-vpc/src/engine/nat.rs
```diff
@@ -49,6 +57,7 @@ use opte::engine::snat::SNat;
 use uuid::Uuid;
 
 pub const NAT_LAYER_NAME: &str = "nat";
+const EXTERNAL_ATTACHED_SUBNET_PRIORITY: u16 = 4;
```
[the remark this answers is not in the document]
FelixMcFelix: Yes, that's the intention. Although my understanding (or hope?) from the control plane is that the usable ranges for these features shouldn't even be allowed to overlap? I.e., the addresses for external NAT/SNAT are ultimately drawn from one or more ranges in address lots, which should lock out those same ranges from being referenced/overlapped by a subnet pool in the same way that each can appear in at most one IP pool.

lib/oxide-vpc/src/engine/overlay.rs
```diff
@@ -268,28 +271,36 @@ impl StaticAction for EncapAction {
...
+            };
 
             match target {
+                // Currently, traffic directed at either attached external subnets or
+                // the external IPs of any other port always go through the V2B table.
+                // This requires a hairpin through the customer network, but provides
+                // strong isolation which some customers require.
+                //
+                // In future we want this to be a tunable property of the VPC. In this
+                // case we would require an extra table/poptrie per VPC, containing all
```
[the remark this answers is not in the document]
FelixMcFelix: Extended this a little in c647128.

lib/oxide-vpc/src/api.rs
FelixMcFelix: That tracks. I know at least today the limit on any kind of normal external IP (SNAT/EIP/FIP) is that those *must* be bound to the primary interface.

PR #924, opened 2026-01-20
Add per-member multicast source filtering

zeeshanlakhani:
Implements IGMPv3/MLDv2-based source filtering for multicast subscriptions at two levels:

- Subscription:: Each member's `SourceFilter` checked on delivery
- Forwarding: Aggregated member filters (union) checked before forwarding to a sled

This work includes:
  - SourceFilter and FilterMode types in oxide-vpc API
  - Fast-path optimization for EXCLUDE() (accept all) case
  - Per-member/port filtering on both Tx and Rx paths
  - Per-next-hop forwarding filtering: Check aggregated SourceFilter before
    forwarding to each remote sled
      - Allows skipping forwarding to sleds where no subscriber would accept the packet
      - This will involve additions to Omicron, but we have to do that anyway for OPTE changes
  - DTrace probe and stats for filtered/filtered-out packets
  - Benchmark(s) and integration tests
    - This PR also includes optimizations to the xde test suite setup for mcast
  - Note: Contains an API bump
  - Minor: clippy test cleanup

We can't really afford (stages, etc) to do this source filtering on the Tofino switch (currently), as it has to happen post-replication, which happens at the Egress portion pipeline of the sidecar. So, it makes sense to handle this in OPTE at the moment.

### DIFF
--- a/crates/opte-api/src/ip.rs
+++ b/crates/opte-api/src/ip.rs
@@ -426,20 +426,18 @@ impl FromStr for IpAddr {
 }
 
 /// An IPv4 address.
-///
-/// Manual `Ord`/`PartialOrd` implementations compare via `u32` rather than
-/// byte-by-byte, producing a single comparison instruction instead of
-/// `memcmp` or a per-byte loop. The ordering is identical (network byte
-/// order is big-endian), but the generated code is significantly faster
-/// for BTreeSet/BTreeMap lookups.
-///
-/// See `VniMac` in `xde::dev_map` for the same rationale.
 #[derive(Clone, Copy, Default, Deserialize, Eq, Hash, PartialEq, Serialize)]
 #[repr(C)]
 pub struct Ipv4Addr {
     inner: [u8; 4],
 }
 
+// Compare via `u32` rather than byte-by-byte, producing a single comparison
+// instruction instead of `memcmp` or a per-byte loop. The ordering is
+// identical (network byte order is big-endian), but the generated code is
+// significantly faster for BTreeSet/BTreeMap lookups.
+//
+// See `VniMac` in `xde::dev_map` for the same rationale.
 impl Ord for Ipv4Addr {
     #[inline]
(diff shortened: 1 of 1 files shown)


lib/oxide-vpc/src/api.rs
```diff
@@ -758,14 +758,29 @@ pub enum DelRouterEntryResp {
...
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct McastForwardingNextHop {
+    /// The unicast IPv6 address of the switch endpoint (for routing).
+    pub next_hop: NextHopV6,
+    /// Tx-only instruction for switch port group replication.
+    pub replication: Replication,
+    /// Aggregated source filter for this destination sled.
+    /// Default (Exclude with empty sources) means accept any source.
+    #[serde(default)]
+    pub source_filter: SourceFilter,
```
FelixMcFelix: Something is off here -- 'destination sled' doesn't seem right. I guess this is in fact the union of all source filters of all destination sleds reachable from the switch who owns the `next_hop`?

lib/oxide-vpc/src/api.rs
```diff
@@ -808,17 +823,100 @@ impl CmdOk for DumpMcastSubscriptionsResp {}
...
+#[repr(u8)]
+pub enum FilterMode {
+    /// Accept packets only from sources in the list.
+    /// Empty list means no sources are accepted.
+    Include = 0,
+    /// Accept packets from any source except those in the list.
+    /// Empty list means all sources are accepted (*, G).
+    #[default]
+    Exclude = 1,
+}
```
FelixMcFelix: Nit: I'm a little surprised to see this as a separate tag type rather than having, e.g., `Include(BTreeSet<IpAddr>)`.

FelixMcFelix (commented):
I think this is looking good. Some comments on the intended use of next-hops and some nits.

PR #945, opened 2026-02-26
Switch to helios3; include helios version in package FMRI

citrus-it:


FelixMcFelix (approved):
Had a quick peek at CI output, the package name looks good to me!

PR #962, opened 2026-04-08
Make `MsgBlk` allocation fallible

### LINKED ISSUE #840: `mblk_t` creation is more fallible than expected
The current implementation of `MsgBlk::new` assumes that `allocb` should never fail.

https://github.com/oxidecomputer/opte/blob/2331d7e98619b185ad2dca3df9c32eca7fa4178e/lib/opte/src/ddi/mblk.rs#L287-L292

This was hit by an omicron-deploy test [here](https://buildomat.eng.oxide.computer/wg/0/details/01K3A4TQFZ6ZSCZRGNCF30519C/JW47GjzbB9X6cPvmz0YImtlk6tTUpuCESUPuIYq9tNuqPBjU/01K3A4V8F0DEHSGCBRAABSPZWF#S682). In context this would have been asking for an extra ~70B to push in front of the packet, but there are two parts in `allocb` which can trip us up:
* This header fragment will be drawn from the `streams_dblk_128` dblk cache, using `kmem_cache_alloc(cache, KM_NOSLEEP)`.
* If this cache is exhausted, it will attempt to create new elements. The `KM_NOSLEEP` flag allows any internal allocation to exit after a few reclamation attempts (noting that the kmem cache internals are a bit more complex than just calling out to `kmem_alloc` for a 128B packet buffer).

This differs from the infallible allocation model we're using everywhere else, which is to `kmem_alloc(sz, KM_SLEEP)` (which matches Rust's expectations around allocations).

We should treat `mblk_t` creation as fallible and most likely drop packets whch fail to allocate during processing. In practice this will affect two classes of traffic: hairpin packets sent in response to packets from the guest, and outbound packets from zones (NTP, Nexus, DNS). Although this should be rare, given the length of time it's taken for this to occur once. Non-hairpin packets from guests should already have sufficient headroom and won't be vulnerable to this.

FelixMcFelix:
This PR makes the creation of new `MsgBlk`s from `allocb` correctly handle the case where the system returns `NULL`. A bunch of downstream functions relied on creating new `mblk_t`s and have similarly been modified to return an error in this case. E.g., accessing a packet's body can require a pullup of the entire frame due to memory loaning from propolis, so the impact is a little wider than expected. We have some new `KStat`s to explicitly show allocation failures, and it turns out we weren't tracking general classes of layer error before in our stats so that's remedied too since we're here.

The only places `unwrap`s have been slotted in are throughout the test harness, since mblk allocation *is* infallible in std/test contexts.

Closes #840.

### DIFF
--- a/lib/opte-test-utils/src/geneve_verify.rs
+++ b/lib/opte-test-utils/src/geneve_verify.rs
@@ -58,13 +58,15 @@ pub fn parse_geneve_packet(bytes: &[u8]) -> Result<GeneveInfo> {
 ///
 /// # Example
 /// ```no_run
-/// let snoop_output = snoop_underlay.assert_packet("on underlay");
-/// let stdout = String::from_utf8_lossy(&snoop_output.stdout);
+/// use opte_test_utils::geneve_verify;
+/// use opte::api::MulticastUnderlay;
+///
+/// let mcast_underlay = MulticastUnderlay::new("ff04::1".parse().unwrap()).unwrap();
 /// geneve_verify::assert_geneve_packet(
-///     &stdout,
-///     vni,
+///     "<...SnoopGuard output...>",
+///     77.try_into().unwrap(),
 ///     mcast_underlay,
-///     Replication::External,
+///     oxide_vpc::api::Replication::External,
 /// );
 /// ```
 pub fn assert_geneve_packet(
--- a/lib/opte-test-utils/src/dhcp.rs
+++ b/lib/opte-test-utils/src/dhcp.rs
@@ -53,7 +53,8 @@ pub fn packet_from_client_dhcpv4_message(
     let headers = (eth, ip, udp);
     let total_len = msg.buffer_len() + headers.packet_length();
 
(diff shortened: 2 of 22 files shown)


lib/opte-test-utils/src/geneve_verify.rs
```diff
@@ -58,13 +58,15 @@ pub fn parse_geneve_packet(bytes: &[u8]) -> Result<GeneveInfo> {
...
+///
+/// let mcast_underlay = MulticastUnderlay::new("ff04::1".parse().unwrap()).unwrap();
 /// geneve_verify::assert_geneve_packet(
-///     &stdout,
-///     vni,
+///     "<...SnoopGuard output...>",
+///     77.try_into().unwrap(),
 ///     mcast_underlay,
-///     Replication::External,
+///     oxide_vpc::api::Replication::External,
```
FelixMcFelix, line 69: `cargo t` was tripping up on illumos because of this doctest: `no_run` will still compile it!

PR #968, opened 2026-04-14
[multicast] Fix multicast Tx self-delivery exclusion using wrong VNI for port lookup

zeeshanlakhani:
The sender self-exclusion check in `handle_mcast_tx` constructed a VniMac key with the multicast VNI from the Geneve header instead of the port's VPC VNI, which is unique. Since DevMap keys ports by VPC VNI, the previous check never matched, causing the sender to receive its own multicast packets.

The fix -> use src_dev.vni for the VniMac lookup key.


FelixMcFelix (approved):
Simple enough fix, thanks!

PR #974, opened 2026-04-17
`SetExternalIpsReq` should treat `None` as an empty list

### LINKED ISSUE #972: `nat::set_external_ips` should handle `None` without failure
Surfaced by https://github.com/oxidecomputer/omicron/issues/10280. IPv6-capable NICs are created in omicron today without any SNAT bindings, and if the instance has no ephemeral/floating IPs then we have no external IPs for the device at all. Sled-agent will pass down `None` in this case. However, if `None` is passed for both IP sets (i.e., an IPv6-only instance) then we'll fall through to the below error case:

https://github.com/oxidecomputer/opte/blob/87ab6e17133d057f762cdcb27940faaa705737a3/lib/oxide-vpc/src/engine/nat.rs#L532-L551

It should be valid that a port can be defined (and reconfigured) to have no external IPs whatsoever. This also raises the point that when receiving `None` we should be handling this case as though we have been told to *remove* the equivalent IP bindings. Today `Some(x), None` and `None, Some(x)` will leave one set of bindings _unmodified_. My intent when writing the API might have been that `None` would avoid updating whatever rules were there and that clients would pass something like `Some(ExternalIpCfg::default())`, but I think it's better to treat this as a command to remove the IPs.

FelixMcFelix:
The control plane treats no external IPs on a port by handing us `None` for its config, where the API previously expected that we would receive `Some(Default::default())` (i.e, an explicit clear operation). However the way we use this in practice is to perform an `ensure` operation on the complete set of config available to us. This change stops us from erroring out in the `(None, None)` case and correctly removes EIP config if the control plane gives us `None` rather than `Some(Default)`.

A few of the tests were implicitly relying on the old behaviour to selectively modify v4 or v6 EIPs, which is counter to how the control plane programs these rules. These have been fixed up.

Closes #972.

### DIFF
--- a/lib/oxide-vpc/src/engine/nat.rs
+++ b/lib/oxide-vpc/src/engine/nat.rs
@@ -519,6 +519,14 @@ fn setup_ipv6_nat(
     Ok(())
 }
 
+/// Replace all external IP-related rules in the NAT layer with a new set
+/// corresponding to `req.external_ips_v4` and  `req.external_ips_v6`.
+///
+/// Passing `None` (or `Some(ExternalIpCfg::default())`) will remove all IPs
+/// of the given family.
+///
+/// Passing `Some(_)` external IPs to a port which does not support that
+/// IP address family will return an error.
 pub fn set_external_ips(
     port: &Port<VpcNetwork>,
     req: SetExternalIpsReq,
--- a/lib/oxide-vpc/src/engine/nat.rs
+++ b/lib/oxide-vpc/src/engine/nat.rs
@@ -524,31 +524,35 @@ pub fn set_external_ips(
     req: SetExternalIpsReq,
 ) -> Result<(), OpteError> {
     let cfg = &port.network().cfg;
+
+    // Check for any cases where the control plane is trying to insert external
+    // IP state for network stack this port is not configured to support.
+    if (matches!(cfg.ip_cfg, IpCfg::Ipv4(_)) && req.external_ips_v6.is_some())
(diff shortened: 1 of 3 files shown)


FelixMcFelix:
Noting here that we will likely want to backport this ontop of 28d5170f6f7d75b6b436e12207eb7ad98a2e4a15, and figure out a way to get the resulting v0.39.xyz package included in TUF repo generation for R19.2.

lib/oxide-vpc/src/engine/nat.rs
bnaecker, line 530: I think we should add a docstring here about the semantics. I definitely didn't understand what the API expected when I made the breaking changes in Omicron!

bnaecker (approved):
Changes look good to me. It's unfortunate that we've hit this semantic mismatch, and I wonder if there are ways we can prevent it in the future. A good docstring on the `set_external_ips()` function would help. Maybe there's a way we can statically prevent it too.

lib/oxide-vpc/src/engine/nat.rs
FelixMcFelix: The semantics have changed now, but yes; agreed!

FelixMcFelix: Done.

> Maybe there's a way we can statically prevent it too.

I think we have a lot of mixed ensure vs. insert semantics (piecewise vs. complete, too), it's not pleasant to keep track of. I'm hoping that we can clamp down on something consistent as part of pushing rule creation to userland for RFD580, but that's not much comfort right now...

PR #980, opened 2026-04-28
Fix dangling references in comment to renamed structures

daniel-levin:
Fixes a comment


FelixMcFelix (approved):
Thanks for fixing this up!

PR #984, opened 2026-05-05
Set OPTE maximum port MTU during creation

### LINKED ISSUE #767: Port passthrough should be removed
https://github.com/oxidecomputer/opte/blob/f5560fae02ad3fc349fabc6454c321143199ca9e/lib/oxide-vpc/src/api.rs#L481-L485

None of our testing infrastructure uses this, and it's certainly not used in the product. I think we're in a good position to strip this out, keeping in mind that there's some amount of documentation that would need amending.

FelixMcFelix:
This allows opteadm/sled-agent to define what the maximum MTU on a link will be on a per-port basis, which should be governed by some sense of control plane policy. I think that create-time is really the only time we can do this: assignment of max MTU (SDU in MAC parlance) happens as part of the call to `mac_register`, and in turn communication of MTU to a guest by viona is basically in a write-once register in the config space.

I think all that seems to need checking is that this is visible off the bat to viona via propolis, and that visibility holds whether or not the propolis is in a non-global zone.

Also closes #767.

### DIFF
--- a/bin/opteadm/src/bin/opteadm.rs
+++ b/bin/opteadm/src/bin/opteadm.rs
@@ -206,14 +206,17 @@ enum Command {
         #[arg(long)]
         src_underlay_addr: Ipv6Addr,
 
+        /// The MTU that should be assigned to the newly created OPTE port.
+        ///
+        /// If unset, this will default to the standard Ethernet MTU of 1500.
+        #[arg(long)]
+        mtu: Option<u32>,
+
         #[command(flatten)]
         external_net: ExternalNetConfig,
 
         #[command(flatten)]
         dhcp: DhcpConfig,
-
-        #[arg(long)]
-        passthrough: bool,
     },
 
     /// Delete an xde device
@@ -823,7 +826,7 @@ fn main() -> anyhow::Result<()> {
             src_underlay_addr,
             dhcp,
             external_net,
-            passthrough,
+            mtu,
         } => {
             let ip_cfg = match private_ip {
                 IpAddr::Ip4(private_ip) => {
@@ -877,7 +880,7 @@ fn main() -> anyhow::Result<()> {
                 dhcp: dhcp.into(),
             };
 
-            hdl.create_xde(&name, cfg, passthrough)?;
+            hdl.create_xde(&name, cfg, mtu)?;
         }
 
         Command::DeleteXde { name } => {
--- a/crates/opte-api/src/lib.rs
+++ b/crates/opte-api/src/lib.rs
@@ -51,7 +51,7 @@ pub use ulp::*;
 ///
 /// We rely on CI and the check-api-version.sh script to verify that
 /// this number is incremented anytime the oxide-api code changes.
-pub const API_VERSION: u64 = 40;
+pub const API_VERSION: u64 = 41;
 
 /// Major version of the OPTE package.
 pub const MAJOR_VERSION: u64 = 0;
--- a/crates/opte-api/src/ndp.rs
+++ b/crates/opte-api/src/ndp.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-// Copyright 2022 Oxide Computer Company
+// Copyright 2026 Oxide Computer Company
 
 //! Types for working with the IPv6 Neighbor Discovery Protocol
 
@@ -30,6 +30,10 @@ pub struct RouterAdvertisement {
     /// Managed address configuration, indicating that the peer can use DHCPv6
     /// to acquire an IPv6 address.
     pub managed_cfg: bool,
+
+    /// Configures whether router advertisements should advertise a given MTU to
+    /// guests.
+    pub mtu: Option<u32>,
 }
 
 impl RouterAdvertisement {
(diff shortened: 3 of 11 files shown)


FelixMcFelix:
Testing notes from falcon using `propolis-server` in the global zone. Links were setup using the below:

```bash
pfexec opteadm set-xde-underlay igb0 igb1
pfexec opteadm create-xde --guest-mac a8:40:25:ff:00:01 --private-ip 10.0.0.1 --vpc-subnet 10.0.0.0/24 --gateway-mac a8:40:25:00:00:01 --gateway-ip 10.0.0.254 --vpc-vni 1701 --src-underlay-addr fd44::01 --mtu 8500 opte0
pfexec opteadm create-xde --guest-mac a8:40:25:ff:00:02 --private-ip 10.0.0.2 --vpc-subnet 10.0.0.0/24 --gateway-mac a8:40:25:00:00:02 --gateway-ip 10.0.0.254 --vpc-vni 1701 --src-underlay-addr fd44::01 --mtu 8500 opte1
pfexec opteadm add-router-entry -p opte0 10.0.0.0/24 sub4=10.0.0.0/24 system
pfexec opteadm add-router-entry -p opte1 10.0.0.0/24 sub4=10.0.0.0/24 system
pfexec opteadm add-fw-rule -p opte0 --dir in --action allow --priority 0 --hosts any --protocol any --ports any
pfexec opteadm add-fw-rule -p opte1 --dir in --action allow --priority 0 --hosts any --protocol any --ports any
```

I modified the `duo` example in falcon to use helios and debian, using the https://github.com/oxidecomputer/falcon/pull/96 branch to create exclusive links atop `opte0` & `opte1`.

Launching the guests, we see on each:
```bash
piano login: root
Linux piano 5.10.0-9-amd64 #1 SMP Debian 5.10.70-1 (2021-09-30) x86_64

The programs included with the Debian GNU/Linux system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.

Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent
permitted by applicable law.
Last login: Tue May  5 10:40:06 UTC 2026 on ttyS0
root@piano:~# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: enp0s6: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 8500 qdisc mq state UP group default qlen 1000
    link/ether a8:40:25:ff:00:02 brd ff:ff:ff:ff:ff:ff
    inet 10.0.0.2/32 brd 10.0.0.2 scope global dynamic enp0s6
       valid_lft 86335sec preferred_lft 86335sec
    inet6 fe80::aa40:25ff:feff:2/64 scope link
       valid_lft forever preferred_lft forever
```
```bash
root@violin:~# dladm
LINK        CLASS     MTU    STATE    BRIDGE     OVER
vioif0      phys      1500   up       --         --
root@violin:~# dladm show-linkprop -pmtu
LINK         PROPERTY        PERM VALUE          DEFAULT        POSSIBLE
vioif0       mtu             rw   1500           1500           60-8500
```

Slightly different default behaviour (Linux goes straight to max MTU, illumos sticks at 1500), but otherwise it is visible to the guest. Now to check for propolis in a zone.

FelixMcFelix:
> Now to check for propolis in a zone.

From testing on standalone omicron, it would appear to work just fine!

```bash
Kernel 6.18.7-0-virt on x86_64 (/dev/ttyS0)

localhost login: root
Welcome to Alpine!

The Alpine Wiki contains a large amount of how-to guides and general
information about administrating Alpine systems.
See <https://wiki.alpinelinux.org/>.

You can setup the system with the command: setup-alpine

You may change this message by editing /etc/motd.

localhost:~# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 8500 qdisc mq state UP qlen 1000
    link/ether a8:40:25:f0:00:00 brd ff:ff:ff:ff:ff:ff
    inet 172.30.0.5/32 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::aa40:25ff:fef0:0/64 scope link 
       valid_lft forever preferred_lft forever
```

In this case I've modified `sled-agent` to create all OPTE ports with an MTU of 8500; not what we want in practice, but enough to prove out the mechanics. I was worried from my last skim over the viona-related code in propolis that we'd have some problems here, but the [underlying issue pointed at there](https://github.com/oxidecomputer/propolis/blob/9343361f922d10c5825906a69c7bb05d139a3662/lib/propolis/src/hw/virtio/viona.rs#L734-L740) is apparently fixed.

rcgoodfellow:
I've incorporated this into Omicron per RFD 689 and things are working well. See testing notes here.

https://github.com/oxidecomputer/omicron/pull/10472

rcgoodfellow (approved):

Issue #779, opened 2025-06-19
Want larger LFT/UFT tables, better resistance against churn


FelixMcFelix:
We currently have table limits of around 65k entries per direction and per layer in our UFTs and layers which rely on stateful actions. Customer deployments of front-facing guests, particularly of web services, are liable to have *far* larger flow counts in practice—potentially $O(10^6)$. We need to be able to support flowtables of larger cardinality, and to keep lookups *fast*.

There are a few things to keep in mind.
* Table sizes (capacity) may need to dynamically scale up and down to minimise memory use on the host.
* We need datastructures optimised for lookup at this scale. We use `BTreeMap`s today because a) they are provided by `alloc::collections::` and b) they provide faster lookup without needing to hash values.

However, a significant problem with the use of `BTreeMap`s is that addition/removal of entries at large table sizes is going to leave us having a [very bad time](https://github.com/oxidecomputer/opte/pull/499#discussion_r1581164767). What's giving me pause is that at, e.g. 8192 elements, we have a mean insertion time of 50.590 µs in the linked benchmark[^1]. For context, the modal packet processing times on dogfood today are around 512–768 ns when everything is in the fastpath. So you can see that's a *very* long time to hold the write lock, and a slow-path packet will pay it several times (NAT in/out, Firewall in/out, UFT) before all is said and done. Under high churn, an individual port can be locked for a very large portion of time like this.

We probably want a mix of things—some form of two-tier `Map` setup to minimise the cost of slowpath processing (and keep UFT hits at `BTreeMap` speed), and some means of making sure that a slow-path operation doesn't necessarily lockout a UFT query.

[^1]: Obviously we're using different lookup keys for our flow tables versus the `RouteKey`s linked in there. Aside from noting that `InnerFlowId`s are larger types, the main concern is the scaling characteristic in terms of $n$ entries. `HashMap`s have more expensive lookups, but insertion time looks to scale a bit less aggressively.

FelixMcFelix:
Redone the above with log-scales and using a shared random sequence of elements for add/remove rather than one known element. This should better show off crossover points for insert/lookup. I think we have good evidence for considering hashbrown at this point, but we'd need to think through what capacity reclamation would look like.

### Insert 1 element

![Image](https://github.com/user-attachments/assets/64d61c43-3b64-46fa-a870-c502e2acbd1b)

![Image](https://github.com/user-attachments/assets/850de99e-ccbe-404b-9544-c47e5aae90c4)

[...]

PR #985, opened 2026-05-07
Configure more aggressive timeouts on non-ESTABLISHED TCP flow entries

FelixMcFelix:
This PR cuts down the TCP state entry expiry times for TCP flow entries which are either still within the three-way handshake or are actively being torn down. This does not affect the validity of any LFT entries which are responsible for actually *procesing* matched packets -- these exist on their own 60s expiry cadence.

This might have some impact on flow stat tracking for one or two packets, but the correct fix there is to get https://github.com/oxidecomputer/opte/pull/744 over the line.

Should answer for
https://github.com/oxidecomputer/customer-support/issues/1125.

### DIFF
--- a/crates/opte-api/src/lib.rs
+++ b/crates/opte-api/src/lib.rs
@@ -51,7 +51,7 @@ pub use ulp::*;
 ///
 /// We rely on CI and the check-api-version.sh script to verify that
 /// this number is incremented anytime the oxide-api code changes.
-pub const API_VERSION: u64 = 39;
+pub const API_VERSION: u64 = 40;
 
 /// Major version of the OPTE package.
 pub const MAJOR_VERSION: u64 = 0;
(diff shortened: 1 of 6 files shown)


rcgoodfellow:
Added commits to bring this in line with #987. Testing in a4x2 and racklettes has been performed on this code. Note that this has a table size bump to 512k from 8k. This will result in 512k entries per-port in capacity on several tables that were previously limited to 8k.

FelixMcFelix:
Thanks for all the testing and further tweaks. I can't self-approve what is technically my own PR, but I'm happy with the changes cherry-picked from #987. The table size bumps should have no impact on workloads with low flow counts -- BTreeMaps do not / cannot preallocate capacity.

From here on I'll be prioritising #986 and #779 with this as a base. The latter is more useful for making sure that lookups/insertions/deletions past around 1000 entries remain cheap.

rcgoodfellow (approved):

rcgoodfellow:
My approval is for #985 and together with the approval from #987 i'm calling this approved.

PR #987, opened 2026-05-08
Ry/faster syn expiry

rcgoodfellow:
A fork of #985 because Github told me bad things might happen if i retarget it to an r19 derivative branch.

Update: indeed bad things happened on a retarget from #985, so this is just R19 plus the commits from #985 cherry picked.

### DIFF
--- a/crates/opte-api/src/lib.rs
+++ b/crates/opte-api/src/lib.rs
@@ -51,7 +51,7 @@ pub use ulp::*;
 ///
 /// We rely on CI and the check-api-version.sh script to verify that
 /// this number is incremented anytime the oxide-api code changes.
-pub const API_VERSION: u64 = 39;
+pub const API_VERSION: u64 = 40;
 
 /// Major version of the OPTE package.
 pub const MAJOR_VERSION: u64 = 0;
--- a/crates/opte-api/src/lib.rs
+++ b/crates/opte-api/src/lib.rs
@@ -51,7 +51,7 @@ pub use ulp::*;
 ///
 /// We rely on CI and the check-api-version.sh script to verify that
 /// this number is incremented anytime the oxide-api code changes.
-pub const API_VERSION: u64 = 40;
+pub const API_VERSION: u64 = 39;
 
 /// Major version of the OPTE package.
 pub const MAJOR_VERSION: u64 = 0;
--- a/lib/opte/src/engine/port/mod.rs
+++ b/lib/opte/src/engine/port/mod.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-// Copyright 2025 Oxide Computer Company
+// Copyright 2026 Oxide Computer Company
 
 //! A virtual switch port.
 
@@ -44,6 +44,7 @@ use super::rule::HdrTransformError;
 use super::rule::PresavedMeoi;
 use super::rule::Rule;
 use super::rule::TransformFlags;
+use super::tcp::INCIPIENT_EXPIRE_TTL;
 use super::tcp::KEEPALIVE_EXPIRE_TTL;
 use super::tcp::TIME_WAIT_EXPIRE_TTL;
 use super::tcp_state::TcpFlowState;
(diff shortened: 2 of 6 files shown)


lib/opte/src/engine/port/mod.rs
```diff
@@ -3012,8 +3015,16 @@ impl ExpiryPolicy<TcpFlowEntryState> for TcpExpiry {
         now: Moment,
     ) -> bool {
         let ttl = match entry.state().tcp_state() {
-            TcpState::TimeWait => self.time_wait_ttl,
-            _ => self.keepalive_ttl,
+            TcpState::TimeWait
+            | TcpState::LastAck
```
rcgoodfellow: Given testing observations, it seems this should move below to map to `self.incipient_ttl`.

bnaecker: Agreed

lib/opte/src/engine/tcp.rs
```diff
@@ -16,11 +16,18 @@ pub const TCP_HDR_OFFSET_SHIFT: u8 = 4;
...
-/// This value is chosen by Windows and MacOS, which is larger
-/// than Linux's default 60s. Allowances for tuned servers and/or
+/// This value is set very low to prevent SYN-flood like traffic (or many
+/// unacknowledged SYNs from the guest) from holding TCP flow entry slots for
+/// the full [`KEEPALIVE_EXPIRE_SECS`].
+pub const INCIPIENT_EXPIRE_SECS: u64 = 5;
+/// The duration after which a connection in TIME-WAIT or another closing state
+/// should be considered free for either side to reuse.
+///
+/// This value is chosen from the TIME-WAIT duratio of Windows and MacOS, which
```
adamlouis:
```suggestion
/// This value is chosen from the TIME-WAIT duration of Windows and MacOS, which
```

lib/opte/src/engine/port/mod.rs
```diff
@@ -3012,8 +3015,16 @@ impl ExpiryPolicy<TcpFlowEntryState> for TcpExpiry {
...
+            | TcpState::CloseWait
+            | TcpState::FinWait1
+            | TcpState::FinWait2 => self.quiescent_ttl,
+            TcpState::SynSent | TcpState::SynRcvd | TcpState::Listen => {
```
bnaecker: I'm slightly concerned about `LISTEN` being here. IIUC, this will expire connections that make the handshake and then don't do much for a while. I dunno how bad that is, but it seems like a "normal" connection and one we'd want to apply the existing logic to, which would map this to `keepalive_ttl`.

lib/opte/src/engine/tcp.rs
bnaecker: spelling nit "duration"

crates/opte-api/src/lib.rs
```diff
@@ -51,7 +51,7 @@ pub use ulp::*;
 ///
 /// We rely on CI and the check-api-version.sh script to verify that
 /// this number is incremented anytime the oxide-api code changes.
-pub const API_VERSION: u64 = 39;
+pub const API_VERSION: u64 = 40;
```
iliana: No, the CI should fail. Don't bump the version on an R19 branch.

lib/opte/src/engine/port/mod.rs
rcgoodfellow: Done

crates/opte-api/src/lib.rs
rcgoodfellow: ok, will revert

lib/opte/src/engine/tcp_state.rs
```diff
@@ -148,26 +148,32 @@ impl TcpFlowState {
...
                 if flags.contains(IngotTcpFlags::ACK) {
                     return Some(Established);
                 }
 
+                // We have a new inbound SYN. We assume for now the
+                // guest is listening on the given port by moving to
+                // the LISTEN state.
+                if flags.contains(IngotTcpFlags::SYN) {
+                    return Some(Listen);
+                }
```
bnaecker, line 175: If we're assuming the guest is listening, should this new state be `SynRcvd`?

lib/opte/src/engine/tcp.rs
rcgoodfellow: Fixed

rcgoodfellow: Fixed

crates/opte-api/src/lib.rs
rcgoodfellow: Done

lib/opte/src/engine/port/mod.rs
rcgoodfellow: `LISTEN` is the handshake state for passive open, I think we should be ok here. We only exit the handshake through established.

lib/opte/src/engine/tcp_state.rs
rcgoodfellow: I believe the logic is we only transition to `SynRcvd` based on getting a packet from the guest (e.g. in `flow_out`) that has the `syn` and `ack` flags set indicating that the guest has ack'd the syn so we can assert `SynRcvd` as the guest's state. And that transition is made from the `Listen` state. So the combination of (1) receiving a `syn` here gets us to `Listen` on an inbound packet and then (2) getting a  `syn+ack` on an outbound packet effectively combine to form a transition to `SynRcvd`. That's lines up with my read of the TCP state transition model and how OPTE should be reacting to it.

bnaecker (approved):
This looks good to me, thanks for double-checking and answering those questions.

PR #990, opened 2026-05-18
Flow eviction and LFT/UFT lifecycle rework

### LINKED ISSUE #788: Only outbound hit time is used in LFT expiry
https://github.com/oxidecomputer/opte/blob/3ff6a21b0eec1a76f34b7ef7cc572e79ef0fdf86/lib/opte/src/engine/layer.rs#L224-L242

Inbound flow hit times are not used in any way for expiry. As the comment here outlines, these need to be held on a shared atomic `Moment` or the expiry logic should inspect both halves.

### LINKED ISSUE #789: UFT and LFT lifetimes are not correctly coupled to one another
UFT entries (fastpath transforms) are created in response to any slowpath traversal, and have their own expiry timestamps and lifecycle. Currently this is 60s from last hit, and LFT entries (stateful allows) also expire after 60s. Suppose we have a UDP request/response pair between A (internal) and B (external):

```
Packet A->B:
 - No outbound UFT entry; slowpath.
 - Firewall; no LFT entry.
   - Rule evaluation: stateful allow.
   - In/out LFT pair are created at time t_0.
 - NAT; no LFT entry.
   - Rule evaluation: stateful allow on SNAT action.
   - In/out LFT pair are created at time t_1, choosing port p.
 - Packet is sent upstream; outbound UFT entry created at time t_2.

Packet B->A:
 - No inbound UFT entry; slowpath.
 - NAT; LFT exists (allow).
   - In LFT hit time refreshed to t_3.
 - Firewall; LFT exists (allow).
   - In LFT hit time refreshed to t_4.
 - Packet is sent to guest inbound UFT entry created at time t_5.

Packet A->B:
 - Outbound UFT entry exists; apply fastpath transform.
   - Out UFT hit time refreshed to t_6.
 - Packet is sent upstream.
```

I've omitted most of the stateless layers here, but the problem is obvious when we see that the LFT timers are never revisted: since the expiry timers are the same, our UFTs outlive all LFTs that were involved in creating them after 60s. This leads to unpleasant situations where finite resources are concerned -- e.g., another flow is able to use the port after `t_3 + 60s` (or `t_1 + 60s` accounting for #788).

The existence of a UFT should prevent any of the LFTs it created from going away. Equally, a hit on a UFT should in some way refresh the timer of every LFT involved in the process (e.g., propagate a UFT's last_hit time downwards when *it* expires).

### LINKED ISSUE #986: Create provisional table for unestablished TCP connections
Instead of TCP state tracking going directly to the TCP table where we need to keep established connections that are potentially carrying valuable traffic, we'll put connections in the TCP handshake in a provisional table.

This provisional table will have more aggressive eviction properties where we prioritize a few things.

1. All connections will get a fair chance at establishing, this means that for some number `N` every connection will get `N` amount of time in the provisional table. If the provisional table is full with all connections that have been there for less than time `N` new connections will get dropped.

2. If there are connections in the provisional table that have been there for longer than `N`, and a new connection is incoming, then the oldest resident provisional connection older than `N` will be evicted and it's corresponding UFT and LFT state (if any) will be evicted as well.

The question of how to calculate `N` is an open one. Should it be static something like a few hundred milliseconds as TCP handshaking should be prompt. Or should it be feedback controlled based on the dynamics of the table.

The provisional table and the LFT table should sum to the current limits, so you don't have a situation where you are processing provisional connections when there is not capacity to graduate them from provisional to established.

We likely also need a provisional table for connection tear down. We are seeing a large number of connections hanging out in `LAST_ACK` waiting for a TCP fin-ack. Could be the same table if that makes sense from an implementation perspective.

FelixMcFelix:
This PR reworks the LFT/UFT lifecycle so that inbound/outbound entries are correctly bound by a single lifetime, and so that a slowpath packet tracks all LFTs which were used in its handling. We then associate this list of LFTs with the flow UFT, and the TCP flow entry if created.

This allows us to correctly keep LFTs alive so long as a UFT/TCP entry refers to them. This has been a longstanding bug. The new and really useful feature is that this allows us to remove all related and dependent entries when we deliberately evict an LFT to make space for a new flow.

An expiry policy for any table can now also specifiy an "eviction priority" for a given flow. Generally speaking most flows are unevictable so long as they are well-behaved, but the main issue is that we need information from elsewhere to *know* when we can make this determination. How we tackle this is by looking at all descendent flows which rely on a given flow. Most will state no preference on the flow lifecycle, but the TCP flow entry can return a non-zero priority when inactive in a given state for too long.

Closes #986.
Closes #788.
Closes #789.

### DIFF
--- a/lib/opte/src/engine/flow_table.rs
+++ b/lib/opte/src/engine/flow_table.rs
@@ -151,37 +151,31 @@ pub trait FlowEntryInfo: fmt::Debug + Send + Sync {
 
 impl<S: FlowState> FlowEntryInfo for FlowEntry<S> {
     fn inherit_last_hit(&self, new_time: Moment) {
-        // The expectation right now is that this will be called within the
-        // port lock, so there will be no concurrent modification of the value.
-        let mut prior = self.lifetime.last_hit.load(Ordering::Relaxed);
         let new = new_time.raw_millis();
-        while new > prior {
-            if let Err(updated) = self.lifetime.last_hit.compare_exchange_weak(
-                prior,
-                new_time.raw_millis(),
-                Ordering::Relaxed,
-                Ordering::Relaxed,
-            ) {
-                prior = updated
-            } else {
-                break;
-            }
-        }
+        // An error from the below call implies a concurrent modification with
+        // a time later than `new_time`. In that case there's just nothing left
+        // to do here.
+        _ = self.lifetime.last_hit.try_update(
+            Ordering::Relaxed,
+            Ordering::Relaxed,
+            |prior| (prior < new).then_some(new),
+        );
     }
 
     fn eviction_priority(&self, now: Moment) -> Option<EvictionPriority> {
         let own_prio = self.policy.eviction_priority(self, now);
 
-        //
+        // An explicit signal that this flow is well-behaved wins outright.
         if let Some(EvictionPriority::Protected) = own_prio {
             return own_prio;
         }
-        // A priority of `None` is unevictable -- we don't have any
-        // information that tells us how poorly behaved.
+
+        // A priority of `None` tells us nothing, and will eventually resolve
+        // into `EvictionPriority::Protected` if no explicit priorities are
+        // provided.
         //
-        // If we have an explicit priority, we keep the *best*
-        // (lowest) priority which we know of. There should be at most
-        // one child in practice.
+        // If we have an explicit priority, we keep the most-protected (lowest)
+        // priority which we know of.
         let mut best_prio = own_prio;
         for maybe_child in &*self.lifetime.children.read() {
             if let Some(child) = maybe_child.0.upgrade() {
@@ -208,10 +202,10 @@ impl<S: FlowState> FlowEntryInfo for FlowEntry<S> {
 
     fn mark_evicted(&self) {
         if !self.lifetime.killed.swap(true, Ordering::Relaxed) {
-            // Any flow entry is only valid while all of its parents still exist.
-            // Timeout-driven expiry will not remove an entry while there are still
-            // live parents, but during eviction we need to go through and mark them
-            // as invalid in turn.
+            // Any flow entry is only valid while all of its parents still
+            // exist. Timeout-driven expiry will not remove an entry while there
+            // are still live parents, but during eviction we need to go through
+            // and mark them as invalid in turn.
             for maybe_child in &*self.lifetime.children.read() {
                 if let Some(child) = maybe_child.0.upgrade() {
                     child.mark_evicted();
@@ -266,7 +260,10 @@ impl<S: FlowState> FlowTable<S> {
         entry
     }
 
-    /// TODO
+    /// Add a new entry to the flow table, assigning it the same lifetime as
+    /// an existing entry in another table.
+    ///
+    /// As in [`Self::add_unchecked`], this elides the capacity check.
     pub fn add_unchecked_partner<T: FlowState>(
         &mut self,
         flow_id: InnerFlowId,
@@ -310,11 +307,11 @@ impl<S: FlowState> FlowTable<S> {
         let mut expired = vec![];
 
         self.map.retain(|flowid, entry| {
-            // A flow cannot be expired while it still has children relying upon
-            // its existence. We should also clear out any dangling entries at
-            // this time.
+            // A flow cannot be expired by the timer while it still has children
+            // relying upon its existence. Check whether any remain, and remove
+            // dangling references to child entries which have expired.
             {
-                // Since we have a write lock on the port, there shouldn't be
+                // We have a write lock on the port, so there shouldn't be
                 // contention here.
                 let mut children = entry.lifetime.children.write();
                 children.retain(|el| el.0.upgrade().is_some());
@@ -342,6 +339,10 @@ impl<S: FlowState> FlowTable<S> {
         expired
     }
 
+    /// Determine whether there is currently space for a new entry to be
+    /// inserted.
+    ///
+    /// If out of space, this method will attempt to evict an existing entry.
     pub fn check_for_space(&mut self) -> Result<()> {
         if self.map.len() < self.limit.get() as usize {
             return Ok(());
@@ -355,12 +356,18 @@ impl<S: FlowState> FlowTable<S> {
         }
     }
 
+    /// Select the flow entry most eligible for eviction (i.e., having the
+    /// numerically highest priority and the oldest timestamp).
+    ///
+    /// Entries which have been killed due to the loss of a dependency will be
+    /// used where possible.
     pub fn find_evictable_entry(&self) -> Option<(InnerFlowId, &FlowEntry<S>)> {
         let now = Moment::now();
 
-        // TODO(ky): some form of datastructure to accelerate?
+        // TODO: some form of datastructure to accelerate this?
         // Who would be responsible for keeping that up to date?
-        // If this cache is wrong, we're just hitting the O(n) scan anyhow?
+        // If that cache is wrong, we're just hitting the O(n) scan anyhow.
+
         let mut to_evict = None;
         for (key, entry) in self.map.iter() {
             if entry.is_killed() {
@@ -494,12 +501,16 @@ pub trait Dump: fmt::Debug + Send + Sync {
     fn dump(&self, hits: u64) -> Self::DumpVal;
 }
 
+/// Common functions needed from the interior state of a flow table entry.
 pub trait FlowState: Dump {
+    /// Return an iterator containing references to all flow entries from other
+    /// tables which underpin `self`.
     fn parents(&self) -> impl Iterator<Item = Arc<dyn FlowEntryInfo>> {
         [].into_iter()
     }
 }
 
+/// Lifecycle state for a flow entry or set of interlinked flow entries.
 struct FlowLifetime {
     /// This tracks the last time the flow was matched.
     ///
@@ -529,6 +540,7 @@ impl fmt::Debug for FlowLifetime {
     }
 }
 
+/// Helper newtype to deduplicate child flow entries by address.
 struct ByAddr(Weak<dyn FlowEntryInfo>);
 
 impl PartialEq for ByAddr {
@@ -554,6 +566,7 @@ impl PartialOrd for ByAddr {
 /// The FlowEntry holds any arbitrary state type `S`.
 #[derive(Debug)]
 pub struct FlowEntry<S: FlowState> {
+    /// The 5-tuple of this flow, used as the lookup key in the parent map.
     id: InnerFlowId,
 
     state: S,
@@ -561,6 +574,7 @@ pub struct FlowEntry<S: FlowState> {
     /// Number of times this flow has been matched.
     hits: AtomicU64,
 
+    /// State determining whether this flow can be expired or evicted.
     lifetime: Arc<FlowLifetime>,
 
     /// Records whether this flow predates a rule change, and
@@ -692,6 +706,9 @@ impl Dump for () {
     fn dump(&self, _hits: u64) {}
 }
 
+impl FlowState for () {}
+
+/// A score of how likely we are to evict a given flow.
 enum EvictionKey {
     Dead,
     Evictable(EvictionPriority, Moment),
@@ -707,8 +724,6 @@ mod test {
     use core::time::Duration;
 
     pub const FT_SIZE: Option<NonZeroU32> = NonZeroU32::new(16);
-
-    impl FlowState for () {}
 
     #[test]
     fn flow_expired() {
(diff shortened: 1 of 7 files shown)


FelixMcFelix:
From what I can tell this appears to have the same throughput on `glasgow` as master. I was mildly concerned that changing the `Arc<TcpFlowEntryState>` held by a UFT into a `Weak<_>` in the name of consistent leak prevention would have had a mild impact, due to the calls to `Weak::upgrade` (similar enough cost to an `Arc::clone`). Otherwise the extra work we're doing affects only the slowpath and the periodic reaper callback.

lib/opte/src/engine/port/mod.rs
```diff
@@ -521,8 +527,19 @@ pub enum DumpLayerError {
     LayerNotFound,
 }
 
+/// Operations that the main flow identifier type for a given [`NetworkImpl`]
+/// must support.
+// This should live in `opte::api`, but we don't want to introduce an
```
rcgoodfellow, line 533: Let's file a follow up issue for this.

lib/opte/src/engine/port/mod.rs
```diff
@@ -2942,6 +3012,47 @@ impl TcpFlowEntryState {
...
+    uft: &Arc<FlowEntry<UftEntry<InnerFlowId>>>,
+    dir: Direction,
+) {
+    // The goal here is to provide each LFT hit with two children where
+    // possible. These are the UFT and, when it exists, the TCP flow entry.
+    // What this means in practice is that while either is present, the LFTs
+    // should be immune to timer-driven expiry. They are *not* immune to
+    // explicit eviction.
+    //
+    // We don't arrange the UFT as a parent of the TCP flow entry because
```
rcgoodfellow, line 3029: Just food for thought, I wonder if the three way relationship between UFT/LFT/TFT is more trilateral in nature than parent/child. E.g. we just want to have the three way associations in place through pointers to be able to have policy over all three that may evolve over time and make the parent/child relationship less of a thing.

lib/opte/src/engine/port/mod.rs
```diff
@@ -2942,6 +3012,47 @@ impl TcpFlowEntryState {
...
+    if let Some(tcp) = uft.state().tcp_flow.as_ref().and_then(Weak::upgrade) {
+        let tcp_dyn: Arc<dyn FlowEntryInfo> = Arc::clone(&tcp) as _;
+        let mut parents = uft.state().parents.clone();
+        let mut inner = tcp.state().inner.lock();
+        let new_parent_slot = match dir {
+            Direction::In => &mut inner.inbound_parents,
+            Direction::Out => &mut inner.outbound_parents,
+        };
+
+        core::mem::swap(new_parent_slot, &mut parents);
```
rcgoodfellow, line 3053: What happens if the reaper runs when the parent/child lists are potentially inconsistent between the parent vec swap here and the child addition/removals below? (or other threads that may hit this inconsistent window, consecutive slowpath packets on the same flow?)

lib/opte/src/engine/port/mod.rs
```diff
@@ -3028,6 +3150,53 @@ impl ExpiryPolicy<TcpFlowEntryState> for TcpExpiry {
...
+                _ => Some(EvictionPriority::Protected),
+            },
+            TcpState::SynSent
+            | TcpState::SynRcvd
+            | TcpState::Listen
+            | TcpState::LastAck => match delta {
+                // Ordinarily we will expire flows in these states quickly. If
+                // we are under table pressure, we can allow them to be
+                // explicitly selected for removal faster.
+                a if a > self.incipient_ttl.as_milliseconds() / 2 => {
```
rcgoodfellow: Consider making the `2` a constant with some rationale on why this divisor was selected.

lib/opte/src/engine/port/mod.rs
```diff
@@ -3028,6 +3150,53 @@ impl ExpiryPolicy<TcpFlowEntryState> for TcpExpiry {
...
+            TcpState::TimeWait
+            | TcpState::CloseWait
+            | TcpState::FinWait1
+            | TcpState::FinWait2 => match delta {
+                // We can remain in a closing state for quite some time. We need
+                // to be willing to evict such entries from this cache fairly
+                // quickly, but leave them in place for the ~120s when we are
+                // not under pressure.
+                a if a > self.incipient_ttl.as_milliseconds() => {
+                    Some(EvictionPriority::Evictable(NonZeroU16::MIN))
```
rcgoodfellow: For flows that are beyond the TTL, should the priority reflect how far beyond the TTL they are, something like the delta in milliseconds so the flows furthest beyond their TTL get evicted first? Same question for line 3182 below.

lib/opte/src/engine/port/mod.rs
```diff
@@ -3028,6 +3150,53 @@ impl ExpiryPolicy<TcpFlowEntryState> for TcpExpiry {
...
+                // explicitly selected for removal faster.
+                a if a > self.incipient_ttl.as_milliseconds() / 2 => {
+                    Some(EvictionPriority::Evictable(NonZeroU16::MIN))
+                }
+                _ => Some(EvictionPriority::Protected),
+            },
+            // Allow established flows to be evicted on the same time scale as
+            // UFT/LFT expiry, but if we are not under pressure then we aim to
+            // keep the LFTs in place even while the flow is inactive.
+            TcpState::Established
```
rcgoodfellow, line 3233: For the established case, we may want to have this be tunable per OPTE port. Maybe I'm not totally understanding the model, but if we get to a place where a flow exists for a TCP session with a sparse communication pattern that has intervals of silence beyond 60 seconds, the flow will stay in place until the system comes under pressure, then the flow may be evicted in favor of new flows due to high connection churn. In that situation, is the flow dead or does it just encounter a slowpath speed bump? If dead, that seems rather unfortunate and I think we'd want to provide a knob for users to tune `FLOW_DEF_TTL` for their applications.

lib/opte/src/engine/flow_table.rs
```diff
@@ -167,33 +307,117 @@ where
...
+                None => {
+                    to_evict = Some((
+                        EvictionKey::Evictable(prio, last_hit),
+                        *key,
+                        entry,
+                    ))
+                }
+                Some((EvictionKey::Evictable(curr_prio, curr_time), ..))
+                    if prio > curr_prio
+                        || (prio == curr_prio && last_hit < curr_time) =>
```
rcgoodfellow, line 396: Is last hit what we want here, or do we want amount of time beyond TTL, as the TTL can pretty different depending on what state the flow is in? This is related to the question earlier on how we set the eviction priority when looking at the delta relative to TTL in `TcpExpiry::eviction_priority`.

lib/opte/src/engine/flow_table.rs
```diff
@@ -266,36 +496,104 @@ fn flow_expired_probe(
...
 
     fn dump(&self, hits: u64) -> Self::DumpVal;
 }
 
+/// Common functions needed from the interior state of a flow table entry.
+pub trait FlowState: Dump {
+    /// Return an iterator containing references to all flow entries from other
+    /// tables which underpin `self`.
+    fn parents(&self) -> impl Iterator<Item = Arc<dyn FlowEntryInfo>> {
+        [].into_iter()
```
rcgoodfellow: I feel like this is a bit of a bug magnet if a new impl forgets or does not realize they need to implement this. Consider removing the default and making impls fill in the method?

rcgoodfellow (commented):
Thanks for putting this together @FelixMcFelix. Questions and comments follow.

On the #986 front, while this PR does not introduce the provisional table we discussed, it looks like it does satisfy properties (1) and (2) from that ticket and therefore would satisfy the intent of the ticket. For (1) we only mark connections as eligible for eviction if the last hit is older than the TTL for the connection state they are in. For (2) the priority based eviction seems to cover that, but I have some questions on how it's implemented below.

lib/opte/src/engine/port/mod.rs
FelixMcFelix: Opened #996.

lib/opte/src/engine/port/mod.rs
FelixMcFelix: Possibly, yeah. At least for now I'm trying to keep something DAG-shaped so that we don't end up with any resource leaks or entry n-tuples ending up with an infinite lifetime because something always depends on them.

I think one thing I was maybe getting at but ended up not implementing is that perhaps UFTs are *always* evictable, but they somehow don't pass that trait onto their parent LFTs. Part of doing that would be that the TCP state isn't invalidated if the UFT disappears, which the current parent/child model allows for. Making that would have probably added yet more complexity for something we really ought to measure (i.e., there's not *as* much need if all the table sizes are the same).

lib/opte/src/engine/port/mod.rs
FelixMcFelix: The port itself is wrapped in a `KRwLock`. Any slowpath hit, or access by the cleanup callback / reaper can only operate on the port and its tables within an exclusive `write()`.

I'll add a comment advertising that we'll need to be more aware of this when we inevitably rethink the locking strategy here at some point.

lib/opte/src/engine/port/mod.rs
FelixMcFelix: I've named a local const `MIDPOINT_TO_EXPIRY` which should suffice, I think.

lib/opte/src/engine/port/mod.rs
FelixMcFelix: The flow would be dead in that situation, but having protection under low load is strictly better than where we are before this change -- any flow dependent on a firewall exception or SNAT entry will be dead after 60s of inactivity, no matter what. I agree that it should be tunable, but I don't really know what the best way to plumb it would be right now.

We could raise this threshold here ("timeouts, timeouts, always wrong"), but I think the separate priority spaces might do a little better here by taking out misbehaved flows first.

lib/opte/src/engine/flow_table.rs
FelixMcFelix: Yes, but I don't think we can always enforce that here in a way that's not confusing (i.e., if we wanted to make UFT entries always evictable then there is no TTL). Similarly you could have policies which are purely protocol state-driven, rather than including timing information. I've tried to capture this for TCP with usng the priority scale.

lib/opte/src/engine/port/mod.rs
FelixMcFelix: Good point. I've made the logic here for misbehaving flows take a linear interpolation across between the evictable time and expiry time to capture this and the below comment:
* `NonZeroU16::MIN` is reserved for established flows which are getting long in the tooth. In this case `last_hit - current_time` will always reflect distance from the TTL.
* `[NonZeroU16::MIN + 1, NonZeroU16::MAX - 1]` for the other states.
* `NonZeroU16::MAX` for the closed state.

lib/opte/src/engine/flow_table.rs
FelixMcFelix: Done in [dba5a03](https://github.com/oxidecomputer/opte/pull/990/commits/dba5a0313b0c6ec8aacde73b30440e4fafb02cb5).

lib/opte/src/engine/port/mod.rs
rcgoodfellow: Thanks!

lib/opte/src/engine/flow_table.rs
rcgoodfellow: My concerns here are addressed by the changes to how flow priority is calculated in `eviction_priority`.

lib/opte/src/engine/port/mod.rs
```diff
@@ -3028,6 +3157,140 @@ impl ExpiryPolicy<TcpFlowEntryState> for TcpExpiry {
...
+        // states before well-behaved but inactive flows.
+        const MISBEHAVED_MIN: NonZeroU16 = NonZeroU16::MIN.saturating_add(1);
+        const MISBEHAVED_MAX: NonZeroU16 =
+            NonZeroU16::new(NonZeroU16::MAX.get() - 1).unwrap();
+
+        /// Choose a priority value between `MISBEHAVED_MIN..=MISBEHAVED_MAX`
+        /// based on `delta_ms`'s position between two timestamps.
+        ///
+        /// Returns `None` if start > end.
+        fn priority_lerp(
```
rcgoodfellow: It would be good to have some unit tests for this function.

rcgoodfellow (approved):
Thanks @FelixMcFelix. Things look good to me. Just one request for some unit tests for the `priority_lerp` function.

lib/opte/src/engine/port/mod.rs
FelixMcFelix: Thanks for the re-review Ry: I'll need to reorganise a tad to pull it out since you can't have tests for an inner item, but that shouldn't prove too bad.

PR #998, opened 2026-05-29
Adjust `RouteCache` to mitigate contention

### LINKED ISSUE #997: Need to better handle contention around `RouteCache`
A guest which chooses to spin up many ill-fated TCP connection attempts (#985, #990) runs into another bottleneck, that being the `RouteCache` which sits between OPTE and the underlay network. This table is small (sized 512 entries), and while it does have an eviction policy what we've seen in practice is that this traffic pattern causes contention and thrashing when all of these packets invariably miss on the table, find it full, and evict a possibly-good entry.

Copying out from https://github.com/oxidecomputer/customer-support/issues/1188#issuecomment-4497160655, we need to:

> * Bump the table size,
> * Disable eviction and just go straight to a raw lookup when table capacity is exceeded,
> * Shrink the write lock scope to cover only insertion, such that readers aren't blocked while one thread is querying illumos itself.

Combined, these should ensure that actually-established flows eventually end up represented in the cache, and that we can sustain a higher active connections-per-second. 

#539 will properly solve this issue and give us a far smaller routing table (because it will *not* scale with the number of active connections), but will require additional interfaces in the kernel to realise. The path described here is a shorter-term fix while we determine what those interfaces should look like.

FelixMcFelix:
This PR follows up on the steps outlined in #997. The main changes are to increase the table size, and to minimise the amount of time holding the write lock on a miss by fetching a new route _before_ taking the lock. This includes refusing to attempt an insert on a failed lookup in a full table.

One of the steps outlined in the issue, not evicting present entries, was already implemented.

Closes #997.

### DIFF
--- a/xde/src/route.rs
+++ b/xde/src/route.rs
@@ -34,7 +34,7 @@ const EXPIRE_ROUTE_LIFETIME: Duration = Duration::from_millis(100);
 const REMOVE_ROUTE_LIFETIME: Duration = Duration::from_millis(1000);
 
 /// Maximum cache size, set to prevent excessive map modification latency.
-const MAX_CACHE_ENTRIES: usize = 32;
+const MAX_CACHE_ENTRIES: usize = 8192;
 
 unsafe extern "C" {
     pub safe fn __dtrace_probe_next__hop(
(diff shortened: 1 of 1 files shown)


FelixMcFelix:
Some quick testing notes on throughput on `glasgow` during the implementation of the above:

| | `-P 1` | `-P 64` | `-P 128` |
|--------|--------|--------|--------|
| master | 14.0 | 31.3 | 31.3 |
| `MAX_CACHE_ENTRIES = 32` | 14.2 | 27.4 | 25.9 |
| + reduced lock scope | 14.0 | 29.1 | 28.7 |
| +`MAX_CACHE_ENTRIES = 8192` | 14.2 | 31.7 | 31.7 | 

My read is that the lock scoping changes (and refusal to take a write at suspected max capacity) do help in the cases that we have more flows than slots. I'll see about getting some lockstat runs.

xde/src/route.rs
```diff
@@ -517,38 +517,47 @@ impl RouteCache {
...
-        let entry_exists = match &maybe_route {
-            Entry::Occupied(e) if e.get().is_valid(t) => {
-                route_hit_probe(map_ptr_int, &key);
-                return (*e.get()).into();
+        //
+        // This shouldn't duplicate work between threads, given that guests
+        // will place traffic from any flow onto the same tx queue.
+        //
+        // If someone *did* write in before us, we still have a valid route.
+        let route = match next_hop(&key, xde) {
```
rcgoodfellow: Small moment of terror when I initially parsed this as this function becoming recursive.

xde/src/route.rs
```diff
@@ -517,38 +517,47 @@ impl RouteCache {
...
             }
-            Entry::Occupied(_) => true,
-            _ => false,
         };
 
+        if !probably_space_remaining {
+            return route;
+        }
+
+        let mut route_cache = self.0.write();
```
rcgoodfellow, line 571: Perhaps putting the route cache update code in it's own method could help prevent accidental extension of the lifetime of this write guard in the case where stuff gets added to this function after the match that follows.

rcgoodfellow (approved):
Thanks @FelixMcFelix. Just one non-blocking suggestion here.

xde/src/route.rs
FelixMcFelix: It's been like this a while, but I'll change it to `get_os_next_hop` for avoidance of any doubt. 😅

xde/src/route.rs
FelixMcFelix: Yep, good call. Done.

PR #1000, opened 2026-06-02
[fix, multicast] wire up missing mcast-tx-no-inner-ip SDT probe

zeeshanlakhani:
The DTrace script `opte/dtrace/opte-mcast-delivery.d` had carried a clause for `mcast-tx-no-inner-ip` since #924, but the corresponding SDT probe was never added on the Rust side, which I (@zeeshanlakhani) introduced.

This fix wires the probe up at the existing fallback path in `handle_mcast_tx` that drops a packet when the inner L3 header is unavailable for source-filter evaluation. The probe takes the `mblk_addr` so the DTrace consumer can inspect the failing packet, matching the pattern established by similar probes: `mcast-tx-pullup-fail`, `mcast-no-fwd-entry`, etc.

Note: no behavior change.


FelixMcFelix (approved):
Thanks!

PR #1012, opened 2026-06-29
[ecmp] Deduplicate redundant multicast next hops via per-target ECMP selection

zeeshanlakhani:
The multicast forwarding loop emitted a copy to every programmed next hop. A next hop is a switch endpoint that fans out the packet to every sled (the underlay) or out the front panel (the external) behind its port groups. Redundant next hops sharing a replication target therefore reach the same destinations and deliver the flow twice, and a receiver cannot tell the duplicates apart, so we handle deduplication at Tx.

This PR adds a `select_nexthops` function, which resolves one next hop per replication target in a single scan over the forwarding table, partitioning eligible hops by replication mode and indexing each partition with the flow's L4 hash. Selection prefers a `Both` replication hop that satisfies the pair with one packet, splitting the flow into per-target copies only when no single hop covers both. The emit loop builds its emissions (at most two) directly from that selection, so the emitted replication reflects which targets chose the hop. All next hops stay programmed for (warm) failover.

Each group's next-hop set is bounded by `MAX_MULTICAST_NEXT_HOPS`, enforced at the set ioctl, so selection partitions eligible hops on the stack without allocating. The ioctl also rejects `Reserved` replication entries with EINVAL, as an accepted entry serves no replication target. Previously, the emit loop cloned and silently discarded them.

Also included:

*Tests*: 
- I've added an Rx-path isolation test that drives `handle_mcast_rx` directly by injecting a raw Geneve-over-IPv6 multicast frame over DLPI, with no guest Tx, so a delivered packet can only have arrived via the underlay receive path. 
- The harness also drops the `add_multicast_router_entry` helper, which programmed the gateway's transit-IPs table and was a no-op for multicast (the gateway and router already admit multicast in both directions).
- It also builds injected frames via ingot's `Emit` rather than mocked-up `mblks`.
- We now block `SnoopGuard` startup until snoop reports its DLPI stream bound, closing a capture race against the test's (possible) first send.

(Local) builds: we thread in TGT_BASE through the buildomat jobs so artifacts can be written to a user-owned directory, and guard the cleanup chown against re-rooting the tree when run elevated locally.

On dtrace, we added per-key mcast aggregations for E2E delivery verification. The delivery script tracked multicast events only in coarse rollups (by event, vni, port, group), which confirm that traffic moved but cannot attribute a specific group's fanout or loss to a port, next hop, or underlay address. End-to-end emulation testing provides this resolution to assert which sleds and switch ports a flow reached.

### DIFF
--- a/xde-tests/src/lib.rs
+++ b/xde-tests/src/lib.rs
@@ -8,6 +8,16 @@ use anyhow::Result;
 use anyhow::anyhow;
 use anyhow::bail;
 use opte_ioctl::OpteHdl;
+use opte_test_utils::Ethernet;
+use opte_test_utils::Ethertype;
+use opte_test_utils::GENEVE_PORT;
+use opte_test_utils::Geneve;
+use opte_test_utils::HeaderLen;
+use opte_test_utils::IngotIpProto;
+use opte_test_utils::Ipv4;
+use opte_test_utils::Ipv6;
+use opte_test_utils::MsgBlk;
+use opte_test_utils::Udp;
 use oxide_vpc::api::AddFwRuleReq;
 use oxide_vpc::api::AddRouterEntryReq;
 use oxide_vpc::api::Address;
@@ -149,6 +159,20 @@ pub const GENEVE_UNDERLAY_FILTER: &str = "ip6 and udp port 6081";
 /// Underlay device name used in single-sled test topology.
 /// The simnet pair creates a loopback underlay for multicast tests.
 pub const UNDERLAY_TEST_DEVICE: &str = "xde_test_sim1";
+
+/// Underlay device used to inject raw frames into the receive path.
+///
+/// A frame written here (the simnet `end_a`) is received on its peer
+/// [`UNDERLAY_TEST_DEVICE`] (`end_b`), rises through `xde_test_vnic1`'s MAC
+/// client, and reaches XDE's `xde_rx` callback.
+pub const UNDERLAY_INJECT_DEVICE: &str = "xde_test_sim0";
+
+/// Service access point is bound on the raw injection stream purely to reach
+/// DLPI's `DL_IDLE` state, a precondition of `dlpi_send`. For ethernet the
+/// service access point is the ethertype. In `DLPI_RAW` it plays no role in
+/// building the frame, so this is an unused experimental ethertype chosen to
+/// avoid demuxing real inbound traffic back into the stream.
+const INJECT_SAP: u32 = 0x4000;
 
 /// This is a wrapper around the ztest::Zone object that encapsulates common
 /// logic needed for running the OPTE tests zones used in this test suite.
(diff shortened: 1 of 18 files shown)


xde/src/xde.rs
```diff
@@ -2302,6 +2302,113 @@ struct MulticastRxContext<'a> {
...
+///
+/// One candidate is chosen per target per flow while all remain programmed in
+/// the forwarding table, so any peer can carry the flow on failover. The caller
+/// suppresses the redundant copy on the others.
+///
+/// Only next hops whose source filter admits `inner_src` are candidates. For an
+/// any-source group (the default `Exclude(empty)`) every hop for the target
+/// qualifies.
+///
+/// For a source-filtered group, only the hops that permit this source do, so a
```
FelixMcFelix:
```suggestion
/// For a source-filtered group, only the hops that permit this source qualify, so a
```

xde/src/xde.rs
```diff
@@ -2302,6 +2302,113 @@ struct MulticastRxContext<'a> {
...
+fn select_nexthops(
+    next_hops: &BTreeMap<NextHopV6, (Replication, SourceFilter)>,
+    inner_src: oxide_vpc::api::IpAddr,
+    l4_hash: u32,
+) -> ReplicationSelection {
+    // A candidate is eligible when its source filter admits this flow and its
+    // next hop serves the target. The count pass precedes the indexing pass.
+    let select = |target: ReplicationTarget| {
+        let count = next_hops
+            .iter()
```
FelixMcFelix:
```suggestion
            .values()
```

xde/src/xde.rs
```diff
@@ -2302,6 +2302,113 @@ struct MulticastRxContext<'a> {
...
+            .count();
+        (count > 0).then(|| l4_hash as usize % count).and_then(|idx| {
+            next_hops
+                .iter()
+                .filter(|(_, (replication, source_filter))| {
+                    source_filter.allows(inner_src)
+                        && target.includes(replication)
+                })
+                .map(|(next_hop, _)| *next_hop)
+                .nth(idx)
```
FelixMcFelix: I don't really love the double linear scan here (or, I guess, $2 \times O(m \log(n))$ for $m$ next hops and $n$ filters in the set), but I think it's probably always going to be cheaper than allocating...

I don't think we'd ever expect to see more than two entries here, right? I.e., one list of source filters per switch. Are we able to maybe hold a small list of keys on the stack that we acquire during the first scan?

xde/src/xde.rs
```diff
@@ -2422,12 +2529,32 @@ fn handle_mcast_tx<'a>(
...
+        // reach the same external multicast network, and underlay candidates
+        // reach the same sled subscribers. Emitting to every next hop for a
+        // target would duplicate the stream.
+        //
+        // We therefore run a two-pass `%` ECMP select-one per target. The first
+        // pass counts source-eligible candidates for the target, and the second
+        // selects `l4_hash % count` in the same stable order. Because the
+        // candidate switches are redundant, picking one avoids duplication. The
+        // two targets are selected independently, so a `Both` replication next
+        // hop can be the choice for one target and not the other.
```
FelixMcFelix: I'm a little confused with the semantics here. The way this reads to me is that if we have a scenario like the below where we have subscribers to groups A and B (putting aside source-based filtering for now...):

```
              + - - - - - +
              +  upstream +
              + [Group B] +
              + - - - - - +
                     ^
                     |
+---------+      +---------+
| switch0 |      | switch1 |
+---------+      +---------+
   |/     |/     \|     \|
Sled A  Sled B  Sled C  Sled D
+-------[Groups A, B]--------+
```

Both `switch0` and `switch1` are available as nexthops in some form, and from how you're structuring this I'd think that `switch0` is an `Underlay` nexthop for both A and B, and `switch1` is `Underlay` for A and `Both` for B.

I think route resolution for group A works as intended, as one of the two switches is always picked for the sole `underlay` entry in `ReplicationSelection`. For group B, it seems as though `underlay` will be either `switch0`/`switch1` based on flowhash, and `external` will be `switch1`. Is that right, or will dendrite and friends only plumb `switch1` as a valid nexthop for this group? If so then we have a 50% chance to perform a host-side duplication rather than letting the switch do everything, which I don't think we want. I.e., it's *correct*, but maybe not *optimal* overhead-wise -- with this tied to the flow ID, this means that some flows will consistently have 2x the Tx cost.

xde/src/xde.rs
```diff
@@ -2422,12 +2529,32 @@ fn handle_mcast_tx<'a>(
...
-            // This filter is the union of all subscriber filters for
-            // this next hop. If no subscriber would accept this source,
-            // skip forwarding.
+            // Check aggregated source filter before forwarding. This filter is
+            // the union of all subscriber filters for destinations reachable
+            // through this next hop. If no subscriber would accept this source,
+            // we skip forwarding. Selection has already excluded this hop. This
+            // second check preserves per-hop drop telemetry for filtered
+            // entries.
             if !source_filter.allows(ctx.inner_src) {
```
FelixMcFelix: Rechecking this is another $O(\log(n))$ for all $n$ filters, and at this point we already know what our nexthop(s) are. Would we be better off incrementing the stat in `select_nexthops`?

xde/src/xde.rs
```diff
@@ -2444,6 +2571,41 @@ fn handle_mcast_tx<'a>(
...
 
+            // Compose the per-flow selections into this hop's effective
+            // replication. A hop keeps a target only if it is that target's
+            // choice.
+            //
+            // A `Both` replication hop is narrowed when it is the choice for
+            // one target but not the other, and skipped entirely when it is the
+            // choice for neither. This emits exactly one external copy and one
+            // underlay copy per flow while each target can land on a different
+            // switch.
```
FelixMcFelix: Related to above: when/why do we want to do that when one packet sent to the `Both` nexthop satisfies all recipients?

FelixMcFelix (commented):
Thanks Zeeshan, I think this looks correct for `Underlay` replication. I've not looked at the added tests yet, I wanted to make sure I'm on the right page with how the external/underlay split is handled in the datapath first.

xde/src/xde.rs
zeeshanlakhani: Yeah, I went through a few iterations here, but I guess we can be more specific/restrictive here. I was going to bring in the heapless dep, which I hope is ok.

xde/src/xde.rs
zeeshanlakhani: Yeah, you're definitely right about the 2x cost (looking at it). When the split occurs, the copies carry disjoint instructions (i.e., Underlay to switch0, External to switch1), so delivery always stays correct, but at the cost of 2x sled-side Tx for those flows. With symmetric modes across the switches, the two candidate sets are identical, so `hash % count` already coalesces to one `Both` replication packet. The split only arises in the heterogeneous shape you considered, which I screwed up and didn't think through. Good catch. 
                                                                                                       
> Is that right, or will dendrite and friends only plumb switch1 as a valid nexthop for this group?

Today, the latter. I've updated the Omicron side (incoming) to elect one switch per group to own the external entry and program the sled-sourced next hop to that same slot owner (hash-seeded on the group UUID, then sticky to the "incumbent" selection), and treat a duplicate external entry on the other switch as drift to clean up. The underlay group does exist on both switches, but only as a failover candidate with its member list remaining populated. It's never handed to OPTE as a second next hop. That code also notes that true ECMP over both switches is the more correct longer-term solution. 

For the fix here, we can coalesce-first, making OPTE optimal either way. When at least one `Both` hop admits the source, we can then hash-select among the `Both` hops and emit one packet. We fall back to independent per-target selection only when no single hop covers both targets. That keeps 1 Tx per flow regardless of whether or not dendrite plumbs switch0 for group B at all. This relates to your question in https://github.com/oxidecomputer/opte/pull/1012#discussion_r3528701101.

xde/src/xde.rs
zeeshanlakhani: Answered in the thread above: with coalesce-first, a covering `Both1 rep hop always "wins" outright, where narrowing only remains for the disjoint case.

xde-tests/src/lib.rs
```diff
@@ -729,6 +753,135 @@ pub fn ensure_underlay_admin_scoped_route_v6(interface: &str) -> Result<()> {
...
+            outer_group_bytes[15],
+        ]),
+        source: MacAddr::from([0x00, 0x11, 0x22, 0x33, 0x44, 0x55]),
+        ethertype: Ethertype::IPV6,
+    };
+
+    let mut frame =
+        MsgBlk::new_ethernet_pkt((outer_eth, outer_ip, outer_udp, geneve));
+    frame.append(inner_pkt);
+    let bytes: Vec<u8> = frame.iter().flat_map(|n| n.iter().copied()).collect();
```
FelixMcFelix: Why go through a mocked up `mblk_t` when we're in userland? 

```suggestion
    let bytes = (outer_eth, outer_ip, outer_udp, geneve, inner_pkt).emit_vec();
```

xde-tests/src/lib.rs
```diff
@@ -729,6 +753,135 @@ pub fn ensure_underlay_admin_scoped_route_v6(interface: &str) -> Result<()> {
...
+        length: (Udp::MINIMUM_LENGTH + payload.len()) as u16,
+        ..Default::default()
+    };
+
+    let mut inner_pkt =
+        MsgBlk::new_ethernet_pkt((inner_eth, inner_ip, inner_udp));
+    if !payload.is_empty() {
+        inner_pkt.append(MsgBlk::copy(payload));
+    }
+    let inner_len = inner_pkt.byte_len();
```
FelixMcFelix: You don't need to serialise this yet.

```suggestion
    let inner_pkt = (inner_eth, inner_ip, inner_udp, payload);
    let inner_len = inner_pkt.packet_length();
```

xde-tests/tests/multicast_rx.rs
```diff
@@ -267,6 +270,113 @@ fn test_xde_multicast_rx_dual_family() -> Result<()> {
...
+    let mcast_group = Ipv4Addr::from([224, 0, 0, 251]);
+    let vni = Vni::new(DEFAULT_MULTICAST_VNI)?;
+    let mcast_underlay =
+        MulticastUnderlay::new("ff04::e000:fb".parse().unwrap()).unwrap();
+
+    // Establish the M2P mapping (cleaned up on drop). No forwarding entry is
+    // configured because forwarding drives Tx replication only.
+    let _mcast = MulticastGroup::new(mcast_group.into(), mcast_underlay)?;
+
+    // Allow IPv4 multicast through the receiver's firewall and subscribe it.
```
FelixMcFelix: Firewall? From the ioctl in use, it's the gateway layer. Also, isn't the port automatically constructed to [allow all multicast dest-addr traffic through the gateway layer](https://github.com/oxidecomputer/opte/blob/7c71b8b6fb6a04a0340edaaecc99cb7c546ab9ab/lib/oxide-vpc/src/engine/gateway/mod.rs#L279-L293)? I'm not sure that `add_multicast_router_entry` needs to exist on those grounds.

If that's insufficient then we should be asking why, because the control plane shouldn't be `allow_cidr`ing for anything other than transit IPs.

FelixMcFelix (commented):
Sorry for the delay. I think the dataplane parts look sensible now, just some comments on the test harness.

xde-tests/tests/multicast_rx.rs
zeeshanlakhani: Yeppers on both counts. The ioctl is the gateway transit-IPs mechanism, not the firewall, and it turns out it was doing nothing here: the gateway already admits multicast in both directions (outbound via the no-spoof rules, inbound via the dedicated `224.0.0.0/4` and `ff00::/8` rules) and the router layer installs a priority-0 outbound multicast allow. 

I've removed the calls from this PR's tests, and enacted some cleanup.

zeeshanlakhani:
@FelixMcFelix updated.

FelixMcFelix (approved):
Neat, thanks, this looks good. Yeah the packet generation is really clean using `((nested), tuples)` and `Emit` that way in userland, that's one of the things I was going for with ingot 😅

PR #1013, opened 2026-06-30
[opteadm] Add show-xde-underlay command.

daniel-levin:
This command only adds value for debugging and development purposes. It is unlikely to be of much value in the field. It is acutely useful for experimenting with different combinations of underlying macs, such as this Chelsio and Mellanox contraption:

```
$ opteadm show-xde-underlay
LINK    MTU   MAC
mlxcx0  9000  6C:B3:11:88:AC:84
cxgbe1  9000  00:07:43:2F:E6:C8
```

### DIFF
--- a/crates/opte-api/src/cmd.rs
+++ b/crates/opte-api/src/cmd.rs
@@ -80,6 +80,11 @@ pub enum OpteCmd {
     /// Requires that no XDE ports exist.
     ClearXdeUnderlay = 73,
 
+    /// Describe the XDE underlay devices. Probably only useful for development
+    /// and debugging purposes, as the underlay devices will always be cxgbe0/1 on actual
+    /// sleds, at least until Metro arrives.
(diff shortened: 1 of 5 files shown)


crates/opte-api/src/cmd.rs
```diff
@@ -488,3 +494,21 @@ pub struct RuleDump {
     pub data_predicates: Vec<String>,
     pub action: String,
 }
+
+/// Tack on 'Resp' to the name to avoid polluting
+/// the space of symbols in XDE-proper.
```
FelixMcFelix, line 499: This probably shouldn't be a doc comment.

FelixMcFelix (approved):
Thanks, I think this looks good! Even if the answer is what we expect 99.9% of the time, it's useful to have in the toolbox to actively confirm that underlay config is set in dev/production debugging I'd think.

PR #1014, opened 2026-06-30
Construct UnderlayDev in its own impl block.

daniel-levin:
At first glance, the definition of UnderlayDev suggests it is just a normal, data-owning struct. It actually owns an opaque handle returned by dls and a per-CPU data structure. In other words, it's hard to construct and the code to do so could be clearly cordoned off, which is exactly what this change does.


FelixMcFelix (approved):
Looks cleaner to me: thanks for working on this!

PR #1021, opened 2026-07-02
Clippy must check for let-else chains in XDE

daniel-levin:
The older syntax can be automatically re-written by Clippy. This lint is not turned on by default in the toolchain used to compile xde. I also removed an unnecessary string clone.

### DIFF
--- a/xde/src/lib.rs
+++ b/xde/src/lib.rs
@@ -20,6 +20,7 @@
 #![feature(alloc_error_handler)]
 #![feature(rustc_private)]
 #![deny(unused_must_use)]
+#![deny(clippy::manual_let_else)]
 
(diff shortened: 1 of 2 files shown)


xde/src/lib.rs
```diff
@@ -20,6 +20,7 @@
 #![feature(alloc_error_handler)]
 #![feature(rustc_private)]
 #![deny(unused_must_use)]
+#![deny(clippy::manual_let_else)]
```
FelixMcFelix, line 23: Ah, it's classified under `pedantic` apparently? Either way it's a good one to enforce.

FelixMcFelix (approved):
Thanks for tidying up!

PR #1022, opened 2026-07-03
Add non-default feature flag 'profiling' to xde

daniel-levin:
This feature flag is off by default, and can only be enabled by manually editing xde/Cargo.toml. It makes it possible to conditionally disable inlining and name mangling on functions of interest. In particular, it has been applied to xde_rx_one and xde_mc_tx_one. While there are probes that serve as proxies for these functions, it is still valuable to be able to probe them directly while working locally.
```
$ dtrace -l | grep "xde_.*x_one entry"
fbt               xde                        xde_rx_one entry
fbt               xde                     xde_mc_tx_one entry
```
Should this workflow become frequently used, it can be made available in the build xtask, but this is a larger project for later, if at all.

### DIFF
--- a/xde/Cargo.toml
+++ b/xde/Cargo.toml
@@ -23,3 +23,6 @@ zerocopy.workspace = true
 crate-type = ["staticlib"]
 name = "xde"
 
+[features]
+default = []
+profiling = []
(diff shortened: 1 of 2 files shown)


FelixMcFelix (approved):
Thanks!

PR #1027, opened 2026-07-14
Generate ICMP packet-too-big messages 

### LINKED ISSUE #999: OPTE should produce 'fragmentation needed' ICMP(v6) hairpins
With the MTU being a known quantity in any port after #984, we need to inform an instance sending oversize frames that it has erred so that PMTUD can function. In this case, OPTE needs to reply with an ICMP 'destination unreachable(fragmentation needed)' or ICMPv6 'packet too big' message on any inbound/outbound traffic which violates the MTU.

There are a few things to be careful of:

* Packets which are GRO/TSO-eligible need to pass by OPTE unharmed so long as the MSS is within bounds.
* Inbound hairpin packets should work (i.e., which are then reflected back on the underlay), but I don't believe they've been tested in any load-bearing capacity til now.
* When replying to an inbound packet we need to re-encapsulate the ICMP message, and we may need to NAT it. What external IP should OPTE choose in this case? For VPC traffic, will the recipient OPTE allow through a packet whose source address is the same as itself?

FelixMcFelix:
This PR introduces a new error handler to `NetworkImpl` to respond to oversized packets without explicit GRO flags present. There are a couple of reasons that this is a new method on the trait:

* We can't have this handled in the standard rule/layer form because we need to catch packets *irrespective of whether they take the fast or slow path*.
* Source-address selection semantics, choice of encapsulation etc. are all defined by the VPC in question, so this cannot be a global OPTE behaviour.
* `NetworkImpl::handle_pkt()` is the wrong place to coerce into handling this, since it's a *terminal* state for any packet, whereas we want, as a possibility, the `NetworkImpl` to just allow through a packet if desired. We'd also need to provide the handler some context on what layer of OPTE called into it.

I think in future we could extend this callback to a `handle_error` function or similar, but since this is the only case I'd rather treat it as YAGNI until a use comes up.

From my testing so far this correctly hairpins packets which have originated from the NIC, and successfully re-encapsulates when doing so. Checksums *look* correct according to `overwatch`. Some packet manipulation before `Port::process` was needed to make sure that packet LSO flags are in place ahead-of-time, but the existing work from the initial ingot rewrite keeps that state in place after the packet is rewritten.

Closes #999.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,8 @@ ctor = "0.10"
 darling = "0.23"
 dyn-clone = "1.0"
 heapless = "0.8"
-ingot = "0.1.1"
+# ingot = "0.1.1"
+ingot = { git = "https://github.com/oxidecomputer/ingot", branch = "extra-impls" }
 ipnetwork = { version = "0.21", default-features = false }
 itertools = { version = "0.15", default-features = false }
 libc = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,8 +50,7 @@ ctor = "0.10"
 darling = "0.23"
 dyn-clone = "1.0"
 heapless = "0.8"
-# ingot = "0.1.1"
-ingot = { git = "https://github.com/oxidecomputer/ingot", branch = "extra-impls" }
+ingot = "0.2.0"
 ipnetwork = { version = "0.21", default-features = false }
 itertools = { version = "0.15", default-features = false }
 libc = "0.2"
--- a/lib/opte/src/engine/icmp/v4.rs
+++ b/lib/opte/src/engine/icmp/v4.rs
@@ -221,8 +221,6 @@ impl<B: ByteSlice> QueryEcho for ValidIcmpV4<B> {
     }
 }
 
-pub const ICMP_DU_FRAGMENTATION_NEEDED: u8 = 4;
-
 /// Internal structure of an ICMP _Destination Unreachable_'s `rest_of_hdr`.
 #[derive(Clone, Debug, Eq, Hash, PartialEq, Ingot)]
 #[ingot(impl_default)]
@@ -237,8 +235,12 @@ pub struct DestinationUnreachable {
     /// over.
     ///
     /// This must be zero for all message types other than
-    /// [`ICMP_DU_FRAGMENTATION_NEEDED`].
+    /// [`Self::FRAGMENTATION_NEEDED`].
     ///
     /// RFC 1191, §4.
     pub mtu: u16be,
+}
+
+impl DestinationUnreachable {
+    pub const FRAGMENTATION_NEEDED: u8 = 4;
 }
--- a/lib/opte/src/engine/icmp/v4.rs
+++ b/lib/opte/src/engine/icmp/v4.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-// Copyright 2025 Oxide Computer Company
+// Copyright 2026 Oxide Computer Company
 
 //! ICMPv4 headers and processing.
 
@@ -219,4 +219,26 @@ impl<B: ByteSlice> QueryEcho for ValidIcmpV4<B> {
             _ => None,
         }
     }
+}
+
+pub const ICMP_DU_FRAGMENTATION_NEEDED: u8 = 4;
+
+/// Internal structure of an ICMP _Destination Unreachable_'s `rest_of_hdr`.
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Ingot)]
+#[ingot(impl_default)]
+pub struct DestinationUnreachable {
+    _rsvd: u8,
+    /// The number of 32-bit words of the original packet included in this
+    /// message's body, after padding to a 4B boundary.
+    ///
+    /// RFC 4884, §4.1.
+    pub length: u8,
+    /// The MTU of the link which this packet was too large to be forwarded
+    /// over.
+    ///
+    /// This must be zero for all message types other than
+    /// [`ICMP_DU_FRAGMENTATION_NEEDED`].
+    ///
+    /// RFC 1191, §4.
+    pub mtu: u16be,
 }
--- a/lib/opte/src/engine/mod.rs
+++ b/lib/opte/src/engine/mod.rs
@@ -253,15 +253,13 @@ pub trait NetworkImpl {
     /// Return the parser for this network implementation.
     fn parser(&self) -> Self::Parser;
 
-    /// Handle a packet which is larget than the underlying [`Port`]'s
+    /// Handle a packet which is larger than the underlying [`port::Port`]'s
     /// MTU. Packets will be caught before applying any existing
     /// UFT transform or applying rule-based processing.
     ///
     /// OPTE will raise this error for any packets larger than the MTU
-    /// without appropriate large send/receive offload flags set by the
-    /// caller.
-    ///
-    /// [`Port`]: port::Port
+    /// which do not have appropriate large send/receive offload flags
+    /// set by the driver.
     // NOTE: we can exchange this in future for a generic `handle_error`,
     // if more use cases become apparent.
     fn handle_oversize<'a, T: Read + Pullup + 'a>(
--- a/lib/opte/src/engine/mod.rs
+++ b/lib/opte/src/engine/mod.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-// Copyright 2025 Oxide Computer Company
+// Copyright 2026 Oxide Computer Company
 
 //! The engine in OPTE.
 //!
@@ -162,6 +162,27 @@ pub enum HdlPktAction {
 /// processing.
 pub struct HdlPktError(pub &'static str);
 
+/// The action to take on a packet which has raised one or more error
+/// conditions during processing.
+///
+/// OPTE can raise certain errors to the [`NetworkImpl`], so that it
+/// can decide how to respond to certain events such as traffic larger
+/// than the specified MTU.
+///
+/// This differs from [`HdlPktAction`] in that an allow action *does not
+/// short-circuit the remainder of normal packet processing*.
+pub enum HdlErrAction {
+    /// OPTE should continue processing this packet.
+    ContinueProcessing,
+
+    /// The packet raising this error must be dropped.
+    Deny,
+
+    /// The packet should not be processed, and a given reply packet
+    /// should be sent as a reply.
+    Hairpin(MsgBlk),
+}
+
 /// An implementation of a particular type of network.
 ///
 /// OPTE is a generalized engine for processing and transforming
@@ -231,6 +252,25 @@ pub trait NetworkImpl {
 
     /// Return the parser for this network implementation.
     fn parser(&self) -> Self::Parser;
+
+    /// Handle a packet which is larget than the underlying [`Port`]'s
+    /// MTU. Packets will be caught before applying any existing
+    /// UFT transform or applying rule-based processing.
+    ///
+    /// OPTE will raise this error for any packets larger than the MTU
+    /// without appropriate large send/receive offload flags set by the
+    /// caller.
+    ///
+    /// [`Port`]: port::Port
+    // NOTE: we can exchange this in future for a generic `handle_error`,
+    // if more use cases become apparent.
+    fn handle_oversize<'a, T: Read + Pullup + 'a>(
+        &self,
+        dir: Direction,
+        pkt: &mut Packet<FullParsed<T>>,
+    ) -> Result<HdlErrAction, HdlPktError>
+    where
+        T::Chunk: ByteSliceMut + IntoBufPointer<'a>;
 }
 
 /// A packet parser for the network implementation.
--- a/lib/opte/src/engine/packet.rs
+++ b/lib/opte/src/engine/packet.rs
@@ -45,6 +45,7 @@ use crate::d_error::DError;
 use crate::ddi::mblk::MsgBlk;
 use crate::ddi::mblk::MsgBlkIterMut;
 use crate::ddi::mblk::MsgBlkNode;
+use crate::ddi::mblk::request_mblk_offload;
 use crate::engine::flow_table::FlowEntryInfo;
 use crate::engine::geneve::GeneveMeta;
 use alloc::boxed::Box;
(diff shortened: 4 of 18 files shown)


lib/opte/src/engine/parse.rs
```diff
@@ -104,6 +105,73 @@ impl<B: ByteSlice + SplitByteSlice> Ulp<B> {
...
 }
 
+impl<B: ByteSliceMut> Ulp<B> {
+    /// Use the L3 pseudoheader checksum and body checksum to derive
+    /// and update the checksum in any upper protocol.
+    pub fn compute_checksum(
+        &mut self,
+        pseudo_csum: Checksum,
+        mut body_csum: Checksum,
+    ) {
```
FelixMcFelix, line 115: This function is moved/deduplicated out of packet.rs.

Cargo.toml
```diff
@@ -50,7 +50,8 @@ ctor = "0.10"
 darling = "0.23"
 dyn-clone = "1.0"
 heapless = "0.8"
-ingot = "0.1.1"
+# ingot = "0.1.1"
+ingot = { git = "https://github.com/oxidecomputer/ingot", branch = "extra-impls" }
```
FelixMcFelix: Calling this out 'til https://github.com/oxidecomputer/ingot/pull/34 is merged.

FelixMcFelix:
https://github.com/oxidecomputer/dendrite/issues/323 and guest loopback added more complexity to both packet generation and handling in the driver than I would have originally liked, but I feel fairly confident in the former pieces. I'm hoping to get some local omicron testing done in the next few days to validate it end-to-end.

FelixMcFelix:
I've done some extra testing on falcon by manually spinning up some OPTE ports on the same host, one at 1500B MTU and one at 8500B MTU (matching the product). I had wanted to test via standalone omicron for completeness, but I'm missing some aspect in my setup to actually assign VMs to the single virtual sled that I couldn't fully work out.

Running a `thundermuffin` UDP server on the 1500MTU VM and a client + snoop on the 8500B MTU VM:

```sh
root@violin:~# dladm
LINK        CLASS     MTU    STATE    BRIDGE     OVER
vioif0      phys      1500   up       --         --
root@violin:~# /opt/cargo-bay/thundermuffin -t udp server 10.0.0.1
```

With sub-MTU packets:
```sh
root@piano:~# dladm
LINK        CLASS     MTU    STATE    BRIDGE     OVER
vioif0      phys      8500   up       --         --
root@piano:~# /opt/cargo-bay/thundermuffin -t udp -b 1000 client 10.0.0.1
[0] 863.516 mbps
[1] 926.414 mbps
[2] 1.010 gbps
[3] 1.000 gbps
[4] 1.026 gbps
[5] 972.039 mbps
[6] 950.336 mbps
[7] 1.048 gbps
[8] 996.555 mbps
[9] 1.022 gbps
------
981.534 mbps
root@piano:~#
```
```sh
root@piano:~# snoop -rv -d vioif0 icmp
Using device vioif0 (promiscuous mode)
root@piano:~#
```

With larger frames:

```sh
root@piano:~# /opt/cargo-bay/thundermuffin -t udp -b 5000 client 10.0.0.1
[0] 2.408 gbps
[1] 2.630 gbps
[2] 2.704 gbps
[3] 2.625 gbps
[4] 2.692 gbps
[5] 2.543 gbps
[6] 2.569 gbps
[7] 2.367 gbps
[8] 2.559 gbps
[9] 2.628 gbps
------
2.572 gbps
root@piano:~#
```
```sh
root@piano:~# snoop -rv -d vioif0 icmp
Using device vioif0 (promiscuous mode)

<SNIP>

ETHER:  ----- Ether Header -----
ETHER:
ETHER:  Packet 43104 arrived at 12:49:16.34477
ETHER:  Packet size = 590 bytes
ETHER:  Destination = a8:40:25:ff:0:2,
ETHER:  Source      = a8:40:25:0:0:2,
ETHER:  Ethertype = 0800 (IP)
ETHER:
IP:   ----- IP Header -----
IP:
IP:   Version = 4
IP:   Header length = 20 bytes
IP:   Type of service = 0x00
IP:         xxx. .... = 0 (precedence)
IP:         ...0 .... = normal delay
IP:         .... 0... = normal throughput
IP:         .... .0.. = normal reliability
IP:         .... ..0. = not ECN capable transport
IP:         .... ...0 = no ECN congestion experienced
IP:   Total length = 576 bytes
IP:   Identification = 0
IP:   Flags = 0x0
IP:         .0.. .... = may fragment
IP:         ..0. .... = last fragment
IP:   Fragment offset = 0 bytes
IP:   Time to live = 128 seconds/hops
IP:   Protocol = 1 (ICMP)
IP:   Header checksum = 23be
IP:   Source address = 10.0.0.254, 10.0.0.254
IP:   Destination address = 10.0.0.2, 10.0.0.2
IP:   No options
IP:
ICMP:  ----- ICMP Header -----
ICMP:
ICMP:  Type = 3 (Destination unreachable)
ICMP:  Code = 4 (Needed to fragment: next hop MTU = 1500)
ICMP:  Checksum = 7ead
ICMP:
ICMP:  [ subject header follows ]
ICMP:
ICMP:IP:   ----- IP Header -----
ICMP:IP:
ICMP:IP:   Version = 4
ICMP:IP:   Header length = 20 bytes
ICMP:IP:   Type of service = 0x00
ICMP:IP:         xxx. .... = 0 (precedence)
ICMP:IP:         ...0 .... = normal delay
ICMP:IP:         .... 0... = normal throughput
ICMP:IP:         .... .0.. = normal reliability
ICMP:IP:         .... ..0. = not ECN capable transport
ICMP:IP:         .... ...0 = no ECN congestion experienced
ICMP:IP:   Total length = 5028 bytes -- truncated
ICMP:IP:   Identification = 27838
ICMP:IP:   Flags = 0x0
ICMP:IP:         .0.. .... = may fragment
ICMP:IP:         ..0. .... = last fragment
ICMP:IP:   Fragment offset = 0 bytes
ICMP:IP:   Time to live = 255 seconds/hops
ICMP:IP:   Protocol = 17 (UDP)
ICMP:IP:   Header checksum = 2788
ICMP:IP:   Source address = 10.0.0.2, 10.0.0.2
ICMP:IP:   Destination address = 10.0.0.1, 10.0.0.1
ICMP:IP:   No options
ICMP:IP:
ICMP:UDP:  ----- UDP Header -----
ICMP:UDP:
ICMP:UDP:  Source port = 43293
ICMP:UDP:  Destination port = 4747
ICMP:UDP:  Length = 5008 (Not all data contained in this fragment)
ICMP:UDP:  Checksum = 0000 (no checksum)
ICMP:UDP:
UDP:

^Croot@piano:~
```

This gives me confidence that the hairpin-in-loopback handling is done correctly.

xde/src/stats.rs
```diff
@@ -175,6 +175,7 @@ impl XdeStats {
             (In, ParseError::UnrecognisedTunnelOpt { .. }) => {
                 &self.in_drop_bad_tun_opt
             }
+            (In, _) => &self.in_drop_misc,
```
rcgoodfellow, line 178: Did this come into play in your testing? Should we add a specific stat for drop for fragmentation needed?

lib/opte/src/engine/icmp/v4.rs
```diff
@@ -220,3 +220,25 @@ impl<B: ByteSlice> QueryEcho for ValidIcmpV4<B> {
         }
     }
 }
+
+pub const ICMP_DU_FRAGMENTATION_NEEDED: u8 = 4;
```
rcgoodfellow: Does it make sense to scope this to the DU struct?

```rust
impl DestinationUnreachable {
    pub const FRAGMENTATION_NEEDED: u8 = 4;
}
```

lib/opte/src/engine/mod.rs
```diff
@@ -231,6 +252,25 @@ pub trait NetworkImpl {
 
     /// Return the parser for this network implementation.
     fn parser(&self) -> Self::Parser;
+
+    /// Handle a packet which is larget than the underlying [`Port`]'s
```
rcgoodfellow: s/larget/larger

lib/opte/src/engine/mod.rs
```diff
@@ -231,6 +252,25 @@ pub trait NetworkImpl {
...
+    /// MTU. Packets will be caught before applying any existing
+    /// UFT transform or applying rule-based processing.
+    ///
+    /// OPTE will raise this error for any packets larger than the MTU
+    /// without appropriate large send/receive offload flags set by the
+    /// caller.
+    ///
+    /// [`Port`]: port::Port
```
rcgoodfellow: I'm not sure what this line in the comment is telling me.

lib/opte/src/engine/packet.rs
```diff
@@ -813,13 +813,29 @@ where
 
     #[inline]
     pub fn mblk_addr(&self) -> uintptr_t {
-        self.state.base_ptr
+        self.state.base_ptr.as_ptr() as uintptr_t
```
rcgoodfellow: Consider avoiding `as` with the following?

```rust
self.state.base_ptr.addr().into()
```

lib/opte/src/engine/packet.rs
```diff
@@ -1162,14 +1178,14 @@ impl<T: Read + Pullup> Packet<FullParsed<T>> {
...
```
rcgoodfellow: Consider avoiding `as` with the following?

```rust
self.state.base_ptr.addr().into()
```

lib/opte/src/engine/packet.rs
```diff
@@ -1317,71 +1274,13 @@ impl<T: Read + Pullup> Packet<FullParsed<T>> {
...
 
         // If a ULP exists and provided a checksum, then update it.
-        if let Some(mut body_csum) = self.body_csum()
+        if let Some(body_csum) = self.body_csum()
             && let Some(ulp) = self.state.meta.headers.inner_ulp.as_mut()
         {
-            let mut csum = body_csum;
             // Unwrap: Can't have a ULP without an IP.
             let ip = self.state.meta.headers.inner_l3.as_ref().unwrap();
-            // Add pseudo header checksum.
```
rcgoodfellow, line 1326: Nice dedup

xde/src/xde.rs
```diff
@@ -3535,34 +3603,59 @@ fn xde_rx_one(
...
+            // switch for generated ICMP traffic. In this case our source and
+            // destination MAC will be zeroed out. If this is the case, then
+            // we need to redetermine which underlay port to use.
+            if hppkt.len() < Ethernet::MINIMUM_LENGTH {
+                // We failed to return a packet with enough bytes to hold
+                // Ethernet in the first layer.
+                return None;
+            }
+
+            let stream = if hppkt[0..ETHER_ADDR_LEN] == [0u8; ETHER_ADDR_LEN]
```
rcgoodfellow: I'm reading this code as 

```rust
if hppkt.eth.src == MacAddr::ZERO || hppkt.eth.dst == MacAddr::ZERO
```

It would be nice to have a comment to that effect or come up with a more apparent way of expressing this in the code itself.

lib/oxide-vpc/src/engine/mod.rs
```diff
@@ -124,6 +169,425 @@ impl NetworkImpl for VpcNetwork {
...
+                    (
+                        L3::<&mut [u8]>::from(new_v4),
+                        Ulp::<&mut [u8]>::from(new_icmp),
+                    ),
+                    // RFC 1812, §4.3.2.3
+                    // This is a larger quantity than that recommended in RFC 1191
+                    // (L3 + 64 bits), and can be expected by certain nested protocols
+                    // like QUIC. RFC 8899 (PLPMTUD for QUIC) directly indicates this
+                    // larger value.
+                    576,
```
rcgoodfellow: Perhaps a named constant for this could be helpful, something like RFC8899_MAX_ICMP_PACKET_SIZE.

lib/oxide-vpc/src/engine/mod.rs
```diff
@@ -124,6 +169,425 @@ impl NetworkImpl for VpcNetwork {
...
+            + truncated_original.byte_len();
+        let encapped_len = u16::try_from(encapped_len)
+            .expect("maximmum inner packet size is below 1300B");
+
+        // Now we can specialise the encap layers with the inner packet size,
+        // and identify a switch address to reach any external sender if required.
+        let encap = encap
+            .as_mut()
+            .map(|(eth, l3, gv)| {
+                // ...Sidecar will give us a zeroed source address on NAT'd
```
rcgoodfellow, line 554: We can probably have the P4 code write the source address as the appropriate vtep loopback address. Seems like that would save a v2b lookup below?

rcgoodfellow (commented):
Thanks @FelixMcFelix. Overall looks great. Comments follow.

xde/src/stats.rs
FelixMcFelix: No, this one covers the `ParseError::NoBuffer` which needed to be added in. Hitting that one is impossible for us because it requires that we have done something silly like advance all the way to the end of a `MsgBlkIterMut` before starting the parse, but `rustc` still demands complete `match` coverage.

lib/opte/src/engine/icmp/v4.rs
FelixMcFelix: Good call. Done.

lib/opte/src/engine/mod.rs
FelixMcFelix: Done.

lib/opte/src/engine/packet.rs
FelixMcFelix: Done.

lib/opte/src/engine/packet.rs
FelixMcFelix: Done.

lib/oxide-vpc/src/engine/mod.rs
FelixMcFelix: The way I've tried to write this is a form of future proofing, but that will work with sidecar as it operates today. I'm keen to make sure that this feature works fully out the gate for customers going entirely via external-attached subnets, where all VM-to-VM traffic is hairpinned via the customer network today.

At the very head of the function I'm reflecting the L2 & L3 encap addresses, and here we only perform the V2B lookup *if* the destination is zero (i.e., the original source was zero). This wipes the L2 addresses when filling in the new L3 destination. That's the marker for `xde` to in turn do a full `RouteCache` access. So, once https://github.com/oxidecomputer/dendrite/issues/323 is implemented and we get to this step with a pair of valid L3 addresses we don't touch the V2B, the L2 headers will be left intact, and `xde` can directly reflect the packet and we avoid any extra work. I hope that this best supports both states of the world?

lib/oxide-vpc/src/engine/mod.rs
FelixMcFelix: I've pulled this and the IPv6 minimum MTU out to constants now.

xde/src/xde.rs
FelixMcFelix: Oh yeah, we have a whole parsing library to that effect, don't we. 😅 `ValidEthernet::parse` is equivalent in cost to just the length comparison, and it cleans this right up.

lib/opte/src/engine/mod.rs
FelixMcFelix: That's just a named link in the doc-comment, I've moved it inline.

FelixMcFelix:
Thanks Ry, I hope this covers your questions/concerns.

rcgoodfellow (approved):

PR #1033, opened 2026-07-23
Fill out cache occupancy and eviction stats

### LINKED ISSUE #1032: Several table quantities are not exposed as kstats
A rough summary of kstats which would be nice to have, since these are handled a little inconsistently and I think we want parity here for something we can surface to the control plane:

`opte$n`:
* `in_uft_flows`, `in_uft_capacity`, `out_uft_flows`, `out_uft_capacity`. LFTs have equivalent stats already.
* `in_uft_evictions`, `out_uft_evictions`. We will not directly hit these today, but may do so in future if we make, e.g., all UFTs subject to LRU at capacity.

`opte$n_$layer`
* `evictions`. We will actively hit this based on TCP state.

`opte$n_routecache` (separate kstat? Currently `Port` owns the top-level kstat name, rather than the driver).
* `capacity`
* `occupancy` (not flows, because with #539 we will track destinations)
* `hit`
* `miss`
* `miss_table_full`

FelixMcFelix:
This PR adds in aditional kstats on each OPTE port corresponding to UFT, LFT, TCP table, and route cache state. I've tried to keep them broadly consistent and make sure that we're surfacing the same information consistently, with the exception of the Tx route cache which has different eviction/replacement semantics.

One thing this has made clear to me is that there are some deficiencies around kstat management. Ideally I'd want to make better use of `ks_update`, and allow `xde` to define the base `KStatNamed` and then the `Port` to push its own keys via `kstat_named_init`. That would also let us wrap `FlowTable` in a kstat-handling layer and cut out a bunch of the internal duplication that `Port` has to take on. I'll write up some thoughts here.

Closes #1032.

### DIFF
--- a/lib/opte/src/engine/port/mod.rs
+++ b/lib/opte/src/engine/port/mod.rs
@@ -346,13 +346,18 @@ impl PortBuilder {
             ),
         };
 
+        let stats = PortStats::new();
+        stats.in_uft_capacity.set(u64::from(data.uft_in.get_limit().get()));
+        stats.out_uft_capacity.set(u64::from(data.uft_out.get_limit().get()));
+        stats.tcp_capacity.set(u64::from(data.tcp_flows.get_limit().get()));
+
         Ok(Port {
             name: self.name.clone(),
             name_cstr: self.name_cstr,
             mac: self.mac,
             ectx: self.ectx,
             epoch: AtomicU64::new(1),
-            stats: KStatNamed::new("xde", &self.name, PortStats::new())?,
+            stats: KStatNamed::new("xde", &self.name, stats)?,
             net,
             data: KRwLock::new(data),
         })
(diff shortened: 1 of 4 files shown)


lib/opte/src/engine/port/mod.rs
```diff
@@ -2680,6 +2742,23 @@ impl<N: NetworkImpl> Port<N> {
...
+            Direction::In => {
+                (&data.uft_in, &vals.in_uft_flows, &vals.in_uft_evictions)
+            }
+            Direction::Out => {
+                (&data.uft_out, &vals.out_uft_flows, &vals.out_uft_evictions)
+            }
+        };
+        let n_flows = uft.num_flows();
+        flow_stat.set(u64::from(n_flows));
+        if n_flows == uft.get_limit().get() {
```
rcgoodfellow: Can you add a comment describing what's going on here? Why do we increment the evict stat by one if the flows are maxed out at the time of uft kstat creation?

rcgoodfellow (commented):
Thanks @FelixMcFelix just one question below

lib/opte/src/engine/port/mod.rs
FelixMcFelix: Urgh, slight mistake on my part. This should be a check that the flow count *prior* to a successful addition was at capacity.

FelixMcFelix: Okay, this is fixed and better explained now. I'm hoping the rebase via this new-fangled stacked PR system doesn't break the ongoing review state.

rcgoodfellow (approved):

PR #1036, opened 2026-07-23
Limit max depth when parsing DHCPv6 options

bnaecker:
Adds a max recursive depth when parsing DHCPv6 options. This can cause a stack overflow with a malicious packet that recursively nests things like IA_TAs. Adds a unit test too.

### DIFF
--- a/lib/opte/src/engine/dhcpv6/options.rs
+++ b/lib/opte/src/engine/dhcpv6/options.rs
@@ -233,6 +233,25 @@ impl<'a> Option<'a> {
     const LEN: Range<usize> = 2..4;
     const DATA_START: usize = 4;
 
+    // Maximum allowed depth for recursively parsing options from bytes.
+    //
+    // DHCPv6 options are recursive, meaning they can contain other options. RFC
+    // 8415 is not explicit about the maximum allowed depth, but we can sort of
+    // infer it. All Options are contained in a Message. Options like the IA_TA
+    // or IA_NA can contain options, so we have at least 2 levels. Each of those
+    // can contain a Status or an IaAddr option. The latter of those can contain
+    // other options, but the spec generally only mentions those as being Status
+    // options themselves. So:
+    //
+    // Message
+    //   - IA_TA / IA_NA option
+    //     - IA Address option
+    //       - Status option
+    //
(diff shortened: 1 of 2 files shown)


lib/opte/src/engine/dhcpv6/options.rs
```diff
@@ -1110,4 +1137,32 @@ mod test {
...
+    #[test]
+    fn fail_on_deeply_nested_options() {
+        let id = IaId(0x1de);
+        let iata = IaTa { id, options: Vec::new() };
+        let base = Option::IaTa(iata);
+        let item_len = base.buffer_len();
+        let mut wrapped = base;
+        for id in 0..Option::MAX_DEPTH {
+            wrapped = Option::IaTa(IaTa {
+                id: IaId(id.try_into().unwrap()),
```
FelixMcFelix:
```suggestion
                id: IaId(id.into()),
```

lib/opte/src/engine/dhcpv6/options.rs
```diff
@@ -506,10 +532,7 @@ impl<'a> IaNa<'a> {
         size_of::<IaId>() + 2 * size_of::<Lifetime>() + option_len
     }
 
-    fn from_bytes(buf: &'a [u8]) -> Result<Self, Error> {
-        if buf.len() < Self::OPTIONS {
```
FelixMcFelix, line 510: We still need this buffer length check, right? All of the successive field reads (3 x `u32`) will immediately unwrap.

FelixMcFelix (commented):
Good catch Ben, thanks for fixing this.

lib/opte/src/engine/dhcpv6/options.rs
bnaecker: We do, thanks for catching. Overeager :)

bnaecker: Done in 34e5f0f

lib/opte/src/engine/dhcpv6/options.rs
bnaecker: Fixed in 34e5f0f

FelixMcFelix (approved):
Thank you!

PR #1039, opened 2026-08-18
Limit the number of LFT->UFT references allowed

FelixMcFelix:
XXX Expand

* UFT table entries have the wrong expiry policy, and TCP flows can
  thrash heavily rather than blocking insertion.
* Eviction should not delegate cleanup of child references to the
  periodic task.
* The number of child references on any FlowEntry should be limited to
  prevent unbounded growth -- we expect this to be 2 active children per
  hit LFT!


FelixMcFelix:
Closing in favour of #1042 -- #1047.

PR #1040, opened 2026-08-19
[xde] update xde_mc_multicst comment: guest multicast filter signals are legit in viona v7

zeeshanlakhani:
Minor change here.


FelixMcFelix (commented):
I think the extra context is useful, but the original contents should remain. The main reason this commentary is here is because OPTE *is* maximally permissive from a device perspective, but the text is suggesting that we could allow XDE to plumb/remove `addrp` as new rules in the gateway layer when asked (or pass this information to our multicast reconciler, or similar).

FelixMcFelix (approved):

PR #1042, opened 2026-08-28
Add microbenchmarks for table expiry & eviction

FelixMcFelix:
This PR adds benchmarks to `cargo ubench` for the above functions
at different levels of table occupancy to get a bead on what
the actual costs involved are.

For expiry/cleanup tests, we set the maximum table size and then mark
some P∊[0,1] flows as having timestamps such that the flow should be
removed entirely. For eviction benchmarks, we ensure that we have at
least one full LFT in addition to a full UFT before processing a packet
on a new 5-tuple.

### DIFF
--- a/bench/src/packet.rs
+++ b/bench/src/packet.rs
@@ -27,6 +27,8 @@ use opte_test_utils::icmp::gen_icmpv6_echo;
 use opte_test_utils::icmp::generate_ndisc;
 use opte_test_utils::*;
 use std::collections::BTreeMap;
+use std::sync::atomic::AtomicU64;
+use std::sync::atomic::Ordering;
 
 pub type TestCase = (MsgBlk, Direction);
 
@@ -42,6 +44,16 @@ pub trait BenchPacket {
 
     /// Return a list of discrete scenarios
     fn test_cases(&self) -> Vec<Box<dyn BenchPacketInstance>>;
+
+    /// Are `generate`d packets worth benchmarking for parser performance?
+    fn do_parse_benchmark(&self) -> bool {
+        true
+    }
+
+    /// Is packet processing allowed to fail due to table size constraints?
+    fn allow_failure(&self) -> bool {
+        false
+    }
 }
 
 /// An individual packet to time the parse/process timing of.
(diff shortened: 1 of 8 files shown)


bench/src/packet.rs
```diff
@@ -334,6 +354,145 @@ impl BenchPacketInstance for UlpProcessInstance {
...
+        let my_index = self.index.fetch_add(1, Ordering::Relaxed);
+
+        // SYN packets (or small UDP) are the easiest way to prod at
+        // UFT expiry behaviour.
+        let src_port = (my_index / u64::from(u16::MAX)) as u16;
+        let dst_port = (my_index % u64::from(u16::MAX)) as u16;
+
+        let body = &[][..];
+
+        let eth = Ethernet {
```
rcgoodfellow, line 450: Cool seeing ingot used in test cases like this!

lib/opte/src/ddi/time.rs
```diff
@@ -58,6 +59,23 @@ impl Add<Duration> for Moment {
     }
 }
 
+impl Sub<Duration> for Moment {
+    type Output = Self;
+
+    fn sub(self, rhs: Duration) -> Self::Output {
+        cfg_if! {
```
rcgoodfellow: In the no-std case we can have subtraction that results in a negative value. The std case will panic with overflow if `rhs` is larger than `self.inner`. Should both cases be a saturating sub?

lib/opte/src/ddi/time.rs
```diff
@@ -81,7 +99,9 @@ impl Moment {
             } else {
                 static FIRST_TS: OnceLock<Instant> = OnceLock::new();
 
-                let first_ts = *FIRST_TS.get_or_init(Instant::now);
+                let first_ts = *FIRST_TS.get_or_init(||
+                    Instant::now() - Duration::from_mins(5)
```
rcgoodfellow, line 105: I'm not following what's happening here. We define an instant that is 5 minutes ago and then the moment called `now` is the time since five minutes ago and whatever time elapsed between the recording of `first_ts` and return value construction? I think some comments are needed here.

bench/src/packet.rs
FelixMcFelix: Thanks for the reviews, Ry. Yeah, it generally makes things pretty easy for me to work with.

lib/opte/src/ddi/time.rs
FelixMcFelix: We only need this `impl` for setting up test cases, so I've taken out the `no_std` half and gated it accordingly. I think saturating there is a good idea for now.

I think for `hrtime_t` in the kernel context becoming negative is fine: that just means we have a timestamp from before the CPU's TSC was last reset. I've opened https://github.com/oxidecomputer/opte/issues/1052 as a related issue, I'd evidently missed the signedness of the type. Fixing that would probably be the best way to unify the behaviour here I think.

lib/opte/src/ddi/time.rs
FelixMcFelix: I think this one was explained in the review for when we were adding in atomic timestamps on table entries but we never transcribed it to the code, sorry.

We have this slightly awkward construction because there's no API on `std::time::Instant` to convert it into any kind of raw numeric type, whereas `gethrtime` gives us that quite explicitly. So to enable atomic timestamps for `std` we need to set some timestamp as our zero value to get a meaningful `u64` (or `i64`). The problem here being that if we want to set flow entries to some time before we run `cargo test`/`cargo ubench`, then we do hit a panicking subtract. It might be possible to come up with something using `SystemTime`, but this is all a work around for the test harness. I've tried to clarify what's going on here better.

rcgoodfellow (approved):

PR #1043, opened 2026-08-28
FlowEntry child list should have a bounded length

FelixMcFelix:
LFT `FlowEntry` elements track a list of all their children. These are our UFT and TCP flow entries -- if we evict an LFT entry, then the child entries can no longer be used. We also use these lists to determine whether an LFT entry can be evicted using the active TCP flow state information, if present.

We know that this set should be bounded, so this PR makes this into a hard requirement to cap the amount of time we can spend computing any one flow's eviction score. This was not the cause of our performance woes relating to flow expiry, but I think that it's a strong belt-and-braces property worth enforcing all of the time.

### DIFF
--- a/lib/opte/src/engine/flow_table.rs
+++ b/lib/opte/src/engine/flow_table.rs
@@ -144,7 +144,7 @@ pub trait FlowEntryInfo: fmt::Debug + Send + Sync {
     fn eviction_priority(&self, now: Moment) -> Option<EvictionPriority>;
 
     /// Set `self` as a parent node to `child`.
-    fn push_child(&self, child: &Arc<dyn FlowEntryInfo>);
+    fn push_child(&self, child: &Arc<dyn FlowEntryInfo>) -> Result<()>;
 
     /// Remove `child` from this entry's list of children.
     fn remove_child(&self, child: &Arc<dyn FlowEntryInfo>);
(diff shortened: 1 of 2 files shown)


rcgoodfellow (approved):
LGTM. Just one non blocking note on observability.

lib/opte/src/engine/flow_table.rs
```diff
@@ -200,9 +200,21 @@ impl<S: FlowState> FlowEntryInfo for FlowEntry<S> {
...
+
+        // Sadly, BTreeSet::entry remains a nightly API.
+        let to_place = ByAddr(Arc::downgrade(child));
+        let can_insert =
+            children.len() < Self::MAX_CHILDREN || children.contains(&to_place);
+        if can_insert {
+            children.insert(to_place);
+            Ok(())
+        } else {
+            Err(OpteError::MaxCapacity(
```
rcgoodfellow, line 214: I think getting here is very unexpected? Should we have a counter here so this can be monitored? It looks like we can got most of the information we'd want from the upstack `__dtrace_probe_port__processs_return` for active debugging. But having a more passive monitoring capability might be nice.

FelixMcFelix: Thanks, I've added the kstats `in_child_capacity_err` and `out_child_capacity_err` in 3133a16f.

rcgoodfellow (approved):

PR #1046, opened 2026-08-28
Limit eviction search to at most 2048 entries

### LINKED ISSUE #1041: Eviction candidates take a long time to identify in large `FlowTable`s
At high levels of table occupancy, we're spending more time than expected trying to identify evictable table entries in `eviction_priority` and related. To some extent this is expected as we're performing a linear scan over all the resident entries -- we haven't identified a valid strategy for dealing with freshness/correctness of e.g. a small secondary eviction likelihood pool per `FlowTable`. However, the cost (and impact on pps rate) appears to suddenly increase when the occupancy goes past a table-dependent threshold.

A hypothesis I had on this is covered by #1039 -- I had originally thought that we could be tracking (many) more children than anticipated per LFT entry, which would greatly inflate the work done for each. But from racklette testing that does not appear to be the case. Other than that I'm not convinced yet that there's anything accidentally quadratic going on under the hood that could explain a greater-than-linear cost in $N$.

We've been seeing this degradation in XDE. I'll see whether similar spikes in performance present themselves in userland as a next step.

FelixMcFelix:
This commit introduces a hard limit to the number of elements that a
slowpath traversal of an LFT is willing to check for eviction scores. In
practice, walking the entire datastructrure has proven unacceptably
expensive, and greatly harms our overall throughput at high load.

Each `FlowTable` includes an eviction counter that we use as the
starting point for a scan. This is filled in whenever an eviction lookup
runs past its budget, if that budget is lower than the total size. This
prevents us from rechecking the same entries over and over.

There is more work to be done around making sure that we can have better
estimates ahead-of-time of likely-evictable flows, but we're not there
yet! Flow cleanup is already an expensive operation, such that we need
to narrow lock granularity before we can push any pre-prep work into the
periodic task.

Closes #1041.

### DIFF
--- a/lib/opte/src/engine/flow_table.rs
+++ b/lib/opte/src/engine/flow_table.rs
@@ -1395,7 +1395,7 @@ mod test {
         let c3 = evict_ft.eviction_cursor.unwrap();
         assert!(c2 < c3);
 
-        evict_ft.expire(&c3);
+        evict_ft.expire(&c3, true);
 
         // A scan should wrap around at the end of the map, if we hit the end
         // with some remaining budget. Accordingly exepct that the flow ID
--- a/lib/opte/src/engine/flow_table.rs
+++ b/lib/opte/src/engine/flow_table.rs
@@ -279,6 +279,10 @@ pub struct FlowTable<S: FlowState> {
 }
 
 impl<S: FlowState> FlowTable<S> {
+    /// The maximum number of distinct entries that a single call to
+    /// [`Self::find_evictable_entry`] will consider.
+    const SCAN_BUDGET: usize = 2048;
+
     /// Add a new entry to the flow table, returning a shared refrence to
     /// the entry.
     ///
@@ -493,15 +497,16 @@ impl<S: FlowState> FlowTable<S> {
         // TODO: some form of datastructure to accelerate this?
         // Who would be responsible for keeping that up to date?
         // If that cache is wrong, we're just hitting the limited scan anyhow.
-        const SCAN_BUDGET: usize = 2048;
         let len = self.map.len();
-        let to_scan = len.min(SCAN_BUDGET);
+        let to_scan = len.min(Self::SCAN_BUDGET);
 
         let mut to_evict = None;
 
         let mut visited = 0;
         while visited < to_scan {
-            let map = if let Some(from) = self.eviction_cursor.take() {
+            let map = if let Some(from) = self.eviction_cursor.take()
+                && to_scan < len
+            {
                 Either::Left(self.map.range(from..))
             } else {
                 Either::Right(self.map.iter())
@@ -554,7 +559,10 @@ impl<S: FlowState> FlowTable<S> {
             }
         }
 
-        to_evict.map(|(_, k, v)| (k, v.as_ref()))
+        to_evict.map(|(_, k, v)| {
+            self.eviction_cursor = Some(k);
+            (k, v.as_ref())
+        })
     }
 
     /// Get the maximum number of entries this flow table may hold.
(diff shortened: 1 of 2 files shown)


lib/opte/src/engine/flow_table.rs
```diff
@@ -448,51 +462,80 @@ impl<S: FlowState> FlowTable<S> {
...
-            let prio = entry.eviction_priority(now).unwrap_or_default();
-            if let EvictionPriority::Protected = prio {
-                continue;
-            }
+        let mut visited = 0;
+        while visited < to_scan {
+            let map = if let Some(from) = self.eviction_cursor.take()
+                && to_scan < len
+            {
+                Either::Left(self.map.range(from..))
```
rcgoodfellow, line 510: So we're going to look at `SCAN_BUDGET` entries starting at the cursor, marking flows as evictable as we find them and then set the cursor to the last evictable entry we found. This raises a few questions

1. The ordering of the underlying btree map is unrelated to evictability, so entries will shuffle around relative to a fixed point cursor as the table churns. Maybe i'm misunderstanding but I could see pathological cases where evictable flows are always shifting away from the scan budget region?

2. Is the ordering of the btree as it exists today meaningful? Should the order be based on evictability in some way such that we go from max scan which will contain a mix of evictables and non-evictables to a max scan of purely evictables (in which case should there even be a max or should we pop until we have no more evictables?). But differently, should the flow table be a priority queue?

FelixMcFelix: > but I could see pathological cases where evictable flows are always shifting away from the scan budget region?

I think so, yeah. I'm not sure what to do on that front. My feeling is that traffic that would do so is updating the `last_hit` of flows, such that they would never be evicted under a full scan anyway.

> Is the ordering of the btree as it exists today meaningful?

It's not meaningful when walking the iterator directly, but the key needs to be just the flow ID (since their `Ord` property is how `BTreeMap` provides $\mathcal{O}(\log{n})$ lookups). 

I did do some testing as to how valuable a secondary `BTreeSet<(Moment, InnerFlowId)>` would be for the benefit of the cleanup task, and it can save us a lot of time there. With eviction though that's insufficient -- a SYN-only flow is at max priority after just a few seconds of inactivity, so looking at all flows ordered by timestamp doesn't help us out. Priority itself changes passively based on e.g. flow state -- maybe part of the issue is that the eviction priority traits are needlessly expressive, and we could subdivide them by class. Not sure how we'd keep that up to date, but there might be something in using fixed classes like that.

FelixMcFelix: So the position we're in now is that (recapping some of Tuesday's discussion):

* The cleanup task will still deterministically expire _all_ timed-out entries in a table.
* Eviction from the dataplane is even more best-effort now, and while it was never guaranteed we're making the trade that we likely won't evict the single worst flow we can find.
* The cost of doing full table scans in the cleanup task is expensive enough that we can't offload (much) prep work to it that might make the above estimate substantially better.

I think a decent tradeoff I can make for now is that any evictable elements that aren't picked for packet A might go into a smaller per-table cache, seen by subsequent packets B, C, etc. that count against the 2048 limit. I think this should alleviate the risk of going past a good set of candidates for one packet and then finding *none* for the next $n$ packets or so.

FelixMcFelix: @rcgoodfellow I think this should be a bit better on this front, as of [30334a5](https://github.com/oxidecomputer/opte/pull/1046/commits/30334a55741479f113f26c22ded6a75686228078).

rcgoodfellow (approved):

PR #1047, opened 2026-08-28
Don't lock eviction parent set if it's likely empty

FelixMcFelix:
By keeping a counter outside of the locked `children` list, we can avoid the cost of acquiring a readlock when an entry has nothing which depends on it. This is most relevant for UFT and TCP flow table entries, which should never have any children, but calculating the eviction score for an LFT entry benefits indirectly since they call `eviction_priority` for each child.

We also use this now to identify flows which have no children, and thus can be evicted because they are ready for expiry but are not cleaned up by the periodic task yet. When this occurs we can exit the eviction candidate scan early.


lib/opte/src/engine/flow_table.rs
```diff
@@ -651,6 +672,10 @@ struct FlowLifetime {
     /// Whether this flow entry has been explicitly removed.
     killed: AtomicBool,
 
+    /// An estimate of the number of elements in [`Self::children`], used
+    /// to avoid locking and querying consistently empty elements.
+    n_children: AtomicUsize,
```
rcgoodfellow, line 704: May want to consider wrapping `n_children` and `children` up in a struct with an interface that makes it impossible for them to get out of sync. Seems like it'd be pretty easy to update `children` in some future PR and forget to update `n_children`.

rcgoodfellow (approved):

PR #1054, opened 2026-09-16
Emit unsolicited Router Advertisements periodically

### LINKED ISSUE #1053: OPTE needs to send unsolicited router adverts
OPTE implements the basics of NDP, and advertises itself as a default router. However, it only does this in response to router _solicitations_, and never sends unsolicited RAs. RFC 4861 is [pretty clear](https://www.rfc-editor.org/info/rfc4861/#section-6.2.1) in the sections on `{Min,Max}RtrAdvLifetime` that conforming implementations MUST send periodic unsolicited RAs (somewhere between every few seconds and every few minutes). We need to implement a periodic in this case that schedules those unsolicited RAs.

bnaecker:
- Add a `NetworkImpl` trait method for returning "autonomously-generated" packets, those the network can spontaneously decide to emit. The `VpcNetwork` emits unsolicited RAs in its implementation.
- Add some tests of the unsolicited RA handling
- Wire up a new `Periodic` in XDE that periodically checks for autonomous packets and delivers them to each port as needed.
- Fixes #1053


bnaecker:
Ok there are still some kinks to suss out around real tests in CI, but I think this is ready for a first round of reviews @FelixMcFelix. I believe I've structured it like we discussed, where the network is periodically polled for "autonomous" packets it wants XDE to deliver. Thanks for your input!

FelixMcFelix:
> Ok there are still some kinks to suss out around real tests in CI

```bash
kyle@farme:~/gits/opte$ cargo xtask install
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.18s
     Running `target/debug/xtask install`
Building opteadm (release).
    Finished `release` profile [optimized + debuginfo] target(s) in 0.12s
Building xde (release-lto).
    Finished `release-lto` profile [optimized + debuginfo] target(s) in 0.27s
Linking xde kmod...
Building xde dev link helper (release).
    Finished `release` profile [optimized + debuginfo] target(s) in 0.17s
Command requires admin privileges to install xde kernel module. Continue? [yY] y
devfsadm: driver failed to attach: xde                                                                                                                                                      Warning: Driver (xde) successfully added to system but failed to attach
kyle@farme:~/gits/opte$ dmesg | tail -n 8
Sep 17 12:12:50 farme genunix: [ID 826211 kern.notice]  '___errno'
Sep 17 12:12:50 farme genunix: [ID 472681 kern.notice] WARNING: mod_load: cannot load module 'xde'
Sep 17 12:13:05 farme genunix: [ID 819705 kern.notice] /kernel/drv/amd64/xde: undefined symbol
Sep 17 12:13:05 farme genunix: [ID 826211 kern.notice]  '___errno'
Sep 17 12:13:05 farme genunix: [ID 472681 kern.notice] WARNING: mod_load: cannot load module 'xde'
Sep 17 12:13:05 farme genunix: [ID 819705 kern.notice] /kernel/drv/amd64/xde: undefined symbol
Sep 17 12:13:05 farme genunix: [ID 826211 kern.notice]  '___errno'
Sep 17 12:13:05 farme genunix: [ID 472681 kern.notice] WARNING: mod_load: cannot load module 'xde'
```

Thinking on what changed, you've added the `sys_rng` flag to rand. I think results in us using the [`getrandom` backend](https://github.com/rust-random/getrandom/blob/master/src/backends/getrandom.rs), i.e., libc's `getrandom` function. AFAICT we're blowing up on `___errno` because there *is* a symbol in the kernel named `getrandom`, corresponding to the implementation of its syscall. But the library's expecting that it's in userland.

Can we seed an `impl Rng` with something from `random_get_pseudo_bytes` or the `genrandom` defined in `<sys/random.h>`?

bnaecker:
Yes! I was talking with Robert yesterday about how to get randomness here, this was a stopgap. I’ll switch to a proper kernel interface today.

bnaecker:
Ok, I've fixed up the randomness stuff and run a sanity check locally. I shorted the RA intervals to be in [3, 6] seconds and created a local IPv6 port. Here's what snoop shows:

```
bnaecker@obsidian : ~/opte $ pfexec snoop -r -d opte0 -t d
Using device opte0 (promiscuous mode)
  0.00000 fe80::aa40:25ff:fe00:1 -> ff02::1      ICMPv6 Router advertisement
  4.99999 fe80::aa40:25ff:fe00:1 -> ff02::1      ICMPv6 Router advertisement
  4.00001 fe80::aa40:25ff:fe00:1 -> ff02::1      ICMPv6 Router advertisement
  4.99999 fe80::aa40:25ff:fe00:1 -> ff02::1      ICMPv6 Router advertisement
  4.00000 fe80::aa40:25ff:fe00:1 -> ff02::1      ICMPv6 Router advertisement
  5.99998 fe80::aa40:25ff:fe00:1 -> ff02::1      ICMPv6 Router advertisement
```

Snapping to the 1s resolution is expected, because the new periodic check only looks for autonomous packets once per second, even though the RA intervals are finer grained. The packet is "ready", but `xde` hasn't asked for it yet.

crates/opte-api/src/lib.rs
```diff
@@ -51,7 +51,7 @@ pub use ulp::*;
 ///
 /// We rely on CI and the check-api-version.sh script to verify that
 /// this number is incremented anytime the oxide-api code changes.
-pub const API_VERSION: u64 = 41;
+pub const API_VERSION: u64 = 42;
```
FelixMcFelix, line 54: *sigh* Every time we need to bump this without the ioctl API actually changing, I shed a tear. One of these days I'll get to serialising the schema for that interface, à la RFD 580.

lib/oxide-vpc/src/engine/gateway/icmpv6.rs
```diff
@@ -108,3 +99,19 @@ pub(super) fn setup(
...
+pub fn router_advert(cfg: &VpcCfg) -> RouterAdvertisement {
+    RouterAdvertisement::new(
+        // From the guest's VPC MAC.
+        cfg.guest_mac,
+        // The MAC from which we respond, i.e., OPTE's MAC.
+        cfg.gateway_mac,
+        // "Managed Configuration", indicating the guest needs to use DHCPv6 to
+        // acquire an IPv6 address.
+        true,
+        // Out current disposition is to always include an MTU announcement in
```
FelixMcFelix:
```suggestion
        // Our current disposition is to always include an MTU announcement in
```

Sorry, I left that typo in there last time...

lib/oxide-vpc/src/engine/mod.rs
```diff
@@ -111,21 +119,160 @@ impl VpcParser {
...
+    //
+    // This is stored in nanos, derived from a `Moment`, so that we can use it
+    // in an atomic.
+    next_time_to_send: AtomicU64,
+    // The raw packet to send.
+    //
+    // We know this upfront, because the unsolicited RA is always from the same
+    // address and sent to a multicast address, so no logic needed.
+    packet: Vec<u8>,
+}
```
FelixMcFelix: Nit: I think these should all be doc-comments?

lib/oxide-vpc/src/engine/mod.rs
```diff
@@ -111,21 +119,160 @@ impl VpcParser {
...
+            Ipv6Addr::ALL_NODES,
+        );
+        UnsolicitedRa {
+            next_time_to_send: AtomicU64::new(
+                Moment::now().raw() + Self::random_interval(),
+            ),
+            packet: MsgBlk::new_ethernet_pkt((eth, ip6, ulp_body))
+                .as_ref()
+                .to_vec(),
+        }
```
FelixMcFelix: You can simplify the packet creation a little more:

```suggestion
        let pkt_layers = RouterAdvertisement::new(
            cfg.guest_mac,
            cfg.gateway_mac,
            true,
            Some(cfg.mtu),
        )
        .build_router_advert_for(
            // Multicast MAC derived from the All-Nodes MC IPv6 address.
            MacAddr::from([0x33, 0x33, 0x00, 0x00, 0x00, 0x01]),
            Ipv6Addr::ALL_NODES,
        );
        UnsolicitedRa {
            next_time_to_send: AtomicU64::new(
                Moment::now().raw() + Self::random_interval(),
            ),
            packet: pkt_layers.emit_vec(),
        }
```

That way we avoid round-tripping through an `mblk_t`.

lib/oxide-vpc/src/engine/mod.rs
```diff
@@ -111,21 +119,160 @@ impl VpcParser {
...
+            cfg.guest_mac,
+            cfg.gateway_mac,
+            true,
+            Some(cfg.mtu),
+        )
+        .build_router_advert_for(
+            // Multicast MAC derived from the All-Nodes MC IPv6 address.
+            MacAddr::from([0x33, 0x33, 0x00, 0x00, 0x00, 0x01]),
+            Ipv6Addr::ALL_NODES,
+        );
```
FelixMcFelix, line 215:
```suggestion
        .build_router_advert_for(
            Ipv6Addr::ALL_NODES.multicast_mac().expect("..."),
            Ipv6Addr::ALL_NODES,
        );
```

I think that should resolve at compile time, those are `const fn`s.

lib/oxide-vpc/src/engine/mod.rs
```diff
@@ -111,21 +119,160 @@ impl VpcParser {
...
+            use rand::TryRng as _;
+            use rand::rngs::SysRng;
+            SysRng.try_next_u64().unwrap_or(0)
+        } else {
+            unsafe extern "C" {
+                pub fn random_get_pseudo_bytes(
+                    ptr: *mut u8,
+                    size: usize,
+                ) -> illumos_sys_hdrs::c_int;
+            }
```
FelixMcFelix: Can we move this up to `crates/illumos-sys-hdrs/src/kernel.rs`?

lib/oxide-vpc/src/engine/mod.rs
```diff
@@ -111,21 +119,160 @@ impl VpcParser {
...
+            let mut x: u64 = 0;
+            unsafe {
+                random_get_pseudo_bytes(
+                    (&mut x as *mut u64).cast(),
```
FelixMcFelix:
```suggestion
                    (&raw mut x).cast(),
```

lib/oxide-vpc/src/engine/mod.rs
```diff
@@ -111,21 +119,160 @@ impl VpcParser {
...
+                .to_vec(),
+        }
+    }
+
+    /// Return a random RA interval within our min / max.
+    fn random_interval() -> u64 {
+        // NOTE: This isn't a very high-quality RNG, but it's also not that
+        // important. We don't care about modulo bias or the fact that this is
+        // fallible and defaults to 0, and we're using a 1s resolution in the
+        // Periodic that XDE uses to actually check for packets anyway.
```
FelixMcFelix, line 248: I think this is acceptable for now. When we need to we can actually `impl Rng` on something, backed by `random_get_pseudo_bytes`, so that we can use rand's distributions and so on.

lib/oxide-vpc/src/engine/mod.rs
```diff
@@ -111,21 +119,160 @@ impl VpcParser {
...
+    ///
+    /// If it is not yet time, the internal timestamp is not changed, and false
+    /// is returned instead.
+    fn is_time_to_send(&self) -> bool {
+        let now = Moment::now().raw();
+        self.next_time_to_send
+            .try_update(Ordering::Relaxed, Ordering::Relaxed, |timeout_at| {
+                // Returns Some(now) if we've reached our timeout, or None
+                // otherwise. The value is only updated in that former case.
+                (now > timeout_at).then_some(now + Self::random_interval())
```
FelixMcFelix: I might move the `Self::random_interval()` outside of this closure, apparently `try_update` can call it several times. I would have thought that would only apply to `update`, but I think it covers the case where someone else updates the stored integer.

lib/oxide-vpc/src/engine/mod.rs
```diff
@@ -602,6 +758,16 @@ impl NetworkImpl for VpcNetwork {
...
     }
+
+    fn autonomous_packets(&self) -> impl Iterator<Item = (Direction, MsgBlk)> {
+        if let Some(last_ra) = &self.last_unsolicited_ra
+            && last_ra.is_time_to_send()
+        {
+            return vec![(Direction::In, MsgBlk::copy(&last_ra.packet))]
+                .into_iter();
+        }
+        vec![].into_iter()
```
FelixMcFelix: You can save a heap allocation with `Option::into_iter`. Up to you on whether the `filter`/`map` style is uglier than if-elsing.

```suggestion
        self.last_unsolicited_ra
            .as_ref()
            .filter(|ra| ra.is_time_to_send())
            .map(|ra| (Direction::In, MsgBlk::new_ethernet_pkt(&ra.packet)))
            .into_iter()
```

The other change, `new_ethernet_pkt`, is just setting up 2 bytes of padding for illumos's benefit.

lib/oxide-vpc/Cargo.toml
```diff
@@ -36,6 +36,7 @@ opte.workspace = true
 cfg-if.workspace = true
 ingot = { workspace = true, optional = true }
 poptrie = { workspace = true, optional = true }
+rand = { workspace = true, optional = true, default-features = false, features = [ "sys_rng" ]}
```
FelixMcFelix: I think we should drop the feature here, and move `"rand/sys_rng"` to the std feature's deps. Leaving this on feels like it might break debug kmod builds or come back and catch us if LTO fails to strip out `SysRng`.

Cargo.toml
```diff
@@ -61,7 +61,7 @@ pcap-parser = "0.17"
 postcard = { version = "1", features = ["alloc"], default-features = false }
 proc-macro2 = { version = "1" }
 quote = "1.0"
-rand = "0.10.0"
+rand = { version = "0.10.0", default-features = false, features = [ "sys_rng" ] }
```
FelixMcFelix: Just `default-features = false`, I think.

xde/src/xde.rs
```diff
@@ -1156,6 +1164,40 @@ fn shared_periodic_expire(_: &mut ()) {
...
+    let devmap = {
+        let devs = state.devs.read();
+        for dev in devs.iter() {
+            for (dir, pkt) in dev.port.network().autonomous_packets() {
+                match dir {
+                    Direction::In => {
+                        // TODO-correctness: We're delivering these packets
+                        // directly to the guest. It might be more prudent to
+                        // pass it through the normal `port.process()` pipeline.
+                        postbox.post(dev.postbox_key, pkt);
```
FelixMcFelix, line 1185: Unsure about this comment. I guess how I look at this is that it's no different from having called `port::process(guest_arp_request)` and getting a hairpin. I think it's fine to assume that packets generated by OPTE itself are directly suitable for the guest

FelixMcFelix (commented):
Thanks Ben, and thanks for the discussion ahead of time. I think what's here generally looks good. Hope these comments make sense!

crates/opte-api/src/lib.rs
bnaecker: Ha yeah that would be nice :)

lib/oxide-vpc/src/engine/mod.rs
bnaecker: Oh! It is there already! I guess I just need to make that a dep of `oxide-vpc`?

lib/oxide-vpc/src/engine/mod.rs
bnaecker: Coool, I've never used the raw pointer syntax. Thanks!

lib/oxide-vpc/src/engine/mod.rs
bnaecker: Ah good catch, thanks.

lib/oxide-vpc/src/engine/mod.rs
bnaecker: Yeah, that's better thanks. I was trying to use `std::iter::{once,empty}` but you get type-mismatches and I thought we'd have to `Box` either way. This is great.

lib/oxide-vpc/Cargo.toml
bnaecker: Oh, right thanks. I meant to reset this an the package's Cargo.toml. I'll do that now.

lib/oxide-vpc/src/engine/mod.rs
bnaecker: Ah I see. This _is_ included, but only in `xde/src/ip.rs`. I'll move it to the header crate, thanks.

lib/oxide-vpc/src/engine/gateway/icmpv6.rs
bnaecker: Fixed in cfedf2f

lib/oxide-vpc/src/engine/mod.rs
bnaecker: Made so in cfedf2f

lib/oxide-vpc/src/engine/mod.rs
bnaecker: Done in cfedf2f

lib/oxide-vpc/src/engine/mod.rs
bnaecker: Done in cfedf2f

lib/oxide-vpc/src/engine/mod.rs
bnaecker: Moved to sys-hdrs in cfedf2f

lib/oxide-vpc/src/engine/mod.rs
bnaecker: Hoisted out in cfedf2f

lib/oxide-vpc/src/engine/mod.rs
bnaecker: Done in https://github.com/oxidecomputer/opte/commit/cfedf2fefb7af5035e7793dac6a3eb9ca2879cc9

Cargo.toml
bnaecker: Done in https://github.com/oxidecomputer/opte/commit/cfedf2fefb7af5035e7793dac6a3eb9ca2879cc9

lib/oxide-vpc/Cargo.toml
bnaecker: Cleaned up in https://github.com/oxidecomputer/opte/commit/cfedf2fefb7af5035e7793dac6a3eb9ca2879cc9

lib/oxide-vpc/src/engine/mod.rs
bnaecker: Done in https://github.com/oxidecomputer/opte/commit/cfedf2fefb7af5035e7793dac6a3eb9ca2879cc9

bnaecker:
Thanks for the feedback @FelixMcFelix!

FelixMcFelix (approved):

PR #1056, opened 2026-09-18
Explicitly reject fragmented IP traffic during parsing

FelixMcFelix:
OPTE's engine has no way to map IP fragment identities for a given src/dst pair back to full 5-tuples, and so cannot provide any sane transit to such packets. However we haven't been filtering this traffic, which can lead to any non-leading fragments being mishandled.

This PR drops all such packets early on. Realistic support down the line will probably require that we make `L4<_>` optional in `Inbound` packets, but a lot of thought will need to go into working out how we'd begin to handle SNAT etc. and other transforms.

On microbenchmarks there is a _small_ hit to per-packet cost (~5%?), but testing on glasgow it seems there is no aggregate performance cost.


lib/opte/src/engine/parse.rs
```diff
@@ -333,6 +337,34 @@ impl fmt::Display for MeoiError {
...
+                    matches!(v, Ok(ValidLowRentV6Eh::IpV6ExtFragment(_)))
+                }),
+                BoxedHeader::Repr(_) => {
+                    unreachable!("ingot must output `Raw` when parsing")
+                }
+            }
+        }
+    };
+
+    if !fragmented { ParseControl::Continue } else { ParseControl::Reject }
```
bnaecker:
```suggestion
    if fragmented { ParseControl::Reject } else { ParseControl::Continue }
```

bnaecker (commented):
Looks good! One nit about the parse control logic, but not blocking. Test looks like it covers it.

bnaecker (approved):

